PR #6, opened 2021-11-17
[Session demo app] Display album artwork in media notifications

dturner:


tonihei:
We can make the equivalent change internally.

dturner:
Thanks for doing this 👍

PR #34, opened 2022-01-30
Fix build error when androidxMediaModulePrefix is specified

h6ah4i:
When the media3 modules are referred from an external project and `gradle.ext.androidxMediaModulePrefix` is specified, build error occurs like the following.

### Error message

```
A problem occurred evaluating project ':media-lib-common'.
> Project with path ':media-media-lib-cast' could not be found in project ':media-lib-common'.
```
(`gradle.ext.androidxMediaModulePrefix = "media-"` in this case)

As you can see, the build script of the common module is trying to use an incorrect named project which has duplicated prefixes.


tonihei:
Thanks for reporting this issue and providing a fix!

PR #64, opened 2022-03-31
Add support for RTSP VP9

rakeshnitb:
Change-Id: Id658564495af13c35fa78ecde9ab587557aabb47


libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -56,6 +56,10 @@
 
   private static final String GENERIC_CONTROL_ATTR = "*";
 
+  /** Default width and height for VP9. */
+  private static final int DEFAULT_VP9_WIDTH = 320;
```
claincly, line 60: For VP8, we ended up with a javadoc like this, please do the same here.

```
  /**
   * Default width for VP8.
   *
   * <p>RFC7741 never uses codec specific data (like width and height) in the fmtp attribute. These
   * values are taken from <a
   * href=https://cs.android.com/android/platform/superproject/+/master:frameworks/av/media/codec2/components/vpx/C2SoftVpxDec.cpp;drc=749a74cc3e081c16ea0e8c530953d0a247177867;l=70>Android's
   * software VP8 decoder</a>.
   */
```

https://github.com/androidx/media/blob/f75710be933aca23f24c5b6d8f2f1925ba1da543/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java#L93

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -129,6 +133,10 @@ public int hashCode() {
         checkArgument(!fmtpParameters.isEmpty());
         processH265FmtpAttribute(formatBuilder, fmtpParameters);
         break;
+      case MimeTypes.VIDEO_VP9:
+        // VP9 does not require a FMTP attribute. So Setting default width and height.
```
claincly, line 137: Please re-word to match the VP8 implementation: 

```
// VP8 never uses fmtp width and height attributes (RFC7741 Section 6.2), setting default
// width and height.
```

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+import androidx.media3.common.util.Util;
+import androidx.media3.exoplayer.rtsp.RtpPacket;
+import androidx.media3.exoplayer.rtsp.RtpPayloadFormat;
+import androidx.media3.extractor.ExtractorOutput;
+import androidx.media3.extractor.TrackOutput;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+
+/**
+ * Parses an VP9 byte stream carried on RTP packets, and extracts VP9 Access Units. Refer to
+ * @link https://datatracker.ietf.org/doc/html/draft-ietf-payload-vp9 for more details.
```
claincly, line 35: Maybe use?

`<a href="...">this draft RFC</a>`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+ */
+/* package */ final class RtpVP9Reader implements RtpPayloadReader {
```
claincly, line 37: Please rename `RtpVP9Reader` to `RtpVp9Reader` (lower case p in VP9), same in the TAG.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+  private static final String TAG = "RtpVP9Reader";
+
+  private static final long MEDIA_CLOCK_FREQUENCY = 90_000;
+
+  private final RtpPayloadFormat payloadFormat;
+
+  private @MonotonicNonNull TrackOutput trackOutput;
+  @C.BufferFlags private int bufferFlags;
+
+  private long firstReceivedTimestamp;
```
claincly, line 48: Add this javadoc to match other readers:

```
  /**
   * First received RTP timestamp. All RTP timestamps are dimension-less, the time base is defined
   * by {@link #MEDIA_CLOCK_FREQUENCY}.
   */
```

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+  private long startTimeOffsetUs;
+  private static int previousSequenceNumber;
+  /** The combined size of a sample that is fragmented into multiple RTP packets. */
+  private int fragmentedSampleSizeBytes;
+  private static int width;
+  private static int height;
+  private static boolean gotFirstPacketOfVP9Frame;
```
claincly, line 55: Add this javadoc (if accurate)
```
  /**
   * Whether the first packet of one VP9 frame is received. A VP9 frame can be split into two RTP
   * packets.
   */
```

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+
+  private @MonotonicNonNull TrackOutput trackOutput;
+  @C.BufferFlags private int bufferFlags;
+
+  private long firstReceivedTimestamp;
+  private long startTimeOffsetUs;
+  private static int previousSequenceNumber;
+  /** The combined size of a sample that is fragmented into multiple RTP packets. */
+  private int fragmentedSampleSizeBytes;
+  private static int width;
```
claincly, line 53: general question, why making these fields static? Can we remove `static`?

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+  private static int height;
+  private static boolean gotFirstPacketOfVP9Frame;
+  private boolean isKeyFrame;
+  private boolean isOutputFormatSet;
+
+  /** Creates an instance. */
+  public RtpVP9Reader(RtpPayloadFormat payloadFormat) {
+    this.payloadFormat = payloadFormat;
+    firstReceivedTimestamp = C.TIME_UNSET;
+    startTimeOffsetUs = 0;
```
claincly, line 63: add comment `// The start time offset must be 0 until the first seek.
`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+    previousSequenceNumber = C.INDEX_UNSET;
+    fragmentedSampleSizeBytes = 0;
+    width = C.INDEX_UNSET;
```
claincly, line 66: Can we use `C.LENGTH_UNSET` here? ditto for height

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+    height = C.INDEX_UNSET;
+    gotFirstPacketOfVP9Frame = false;
+    isKeyFrame = false;
+    isOutputFormatSet = false;
+  }
+
+  @Override
+  public void createTracks(ExtractorOutput extractorOutput, int trackId) {
+    trackOutput = extractorOutput.track(trackId, C.TRACK_TYPE_VIDEO);
+    castNonNull(trackOutput).format(payloadFormat.format);
```
claincly, line 76: I think you can drop the `castNonNull` here.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+    //        +-+-+-+-+-+-+-+-+
+
+    int header = payload.readUnsignedByte();
+    if (!gotFirstPacketOfVP9Frame) {
+      // For start of VP9 partition B=1 as per VP9 RFC Section 4.2.
+      if ((header & 0x08) == 0) {
+        Log.w(
+            TAG,
+            Util.formatInvariant(
+                "First payload octet of the RTP packet is not the beginning of a new VP9 partition,"
```
claincly, line 167: Maybe "Start of a frame bit (B) in the payload header is unset in the first VP9 Frame."

and I think the error message is clear enough so that you can remove the comment on L162.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+                    + " Dropping current packet." + header));
```
claincly, line 168: I don't find it necessary to log the header here.

- If you think it's necessary, please log with format string, like `Util.formatInvariant("xxxx %d", header);` instead of using `+`.
- else, just log the string and remove `Util.formatInvariant`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+      // Check M for 15 bits PictureID.
+      if ((optionalHeader & 0x80) != 0) {
+        if (payload.bytesLeft() < 1) {
+          return false;
+        }
+      }
+    }
+
+    // Flexible-mode not implemented.
+    checkArgument((header & 0x10) == 0, "VP9 flexible mode unsupported");
```
claincly, line 198: - Please specify the reason not to support it
- Move this check above to right after getting the header.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+
+    // Check L optional header present.
```
claincly, line 200:
```
Check if optional L header is present.
      ^^                   ^^
```

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+    if ((header & 0x20) != 0) {
```
claincly, line 201: As a general comment, using `ParsableBitArray` may be easier here.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+      if (payload.bytesLeft() < 1) {
+        return false;
+      }
+      // Check TL0PICIDX header present (non-flexible mode).
+      if ((header & 0x10) == 0) {
+        payload.skipBytes(1);
+      }
+    }
+
+    // Check V optional header present, Refer Section 4.2.1.
```
claincly, line 212: Check if optional V header is present. Refer to Section 4.2.1.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+    if ((header & 0x02) != 0) {
+      int scalabilityStr = payload.readUnsignedByte();
```
claincly, line 214: What is `Str` in `scalabilityStr`? Is it `stream`? If so please spell it out

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+      int numSpatialLayers = (scalabilityStr & 0xe0) >> 5;
```
claincly, line 215: `(scalabilityStr >> 5) & 0x7`

If you do `& 0xe0` and then shift, there could be sign extension. Basically `0xF0000000 >> 1` is `0xF8000000` instead of `0x78000000`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+      int scalabilityStrLength = ((scalabilityStr & 0x10) != 0) ? numSpatialLayers + 1 : 0;
+
+      if ((scalabilityStr & 0x10) != 0) {
+        if (payload.bytesLeft() < scalabilityStrLength * 4) {
```
claincly, line 219: Can you explain the `4` here? Maybe link to the RFC and make a constant.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+          return false;
+        }
+        for (int index = 0; index < scalabilityStrLength; index++) {
+          width = payload.readUnsignedShort();
```
claincly, line 223: is making width and height static to accommodate the assignment here? If so I would suggest making this method non-static.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+  private static long toSampleUs(
+      long startTimeOffsetUs, long rtpTimestamp, long firstReceivedRtpTimestamp) {
+    return startTimeOffsetUs
+        + Util.scaleLargeTimestamp(
+            (rtpTimestamp - firstReceivedRtpTimestamp),
+            /* multiplier= */ C.MICROS_PER_SECOND,
+            /* divisor= */ MEDIA_CLOCK_FREQUENCY);
+  }
+
+  private static boolean parseVP9Descriptor(ParsableByteArray payload, int packetSequenceNumber)
```
claincly, line 141: I would recommend renaming the method to `validateVp9Descriptor`. And document that the position in the `ParsableByteArray` is moved upon returnning.

We used this javadoc for VP8:

```
  /**
   * Returns {@code true} and sets the {@link ParsableByteArray#getPosition() payload.position} to
   * the end of the descriptor, if a valid VP8 descriptor is present.
   */

```

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+  public void consume(ParsableByteArray data, long timestamp, int sequenceNumber, boolean rtpMarker)
+      throws ParserException {
+    checkStateNotNull(trackOutput);
+
+    if (parseVP9Descriptor(data, sequenceNumber)) {
+      if (fragmentedSampleSizeBytes == 0 && gotFirstPacketOfVP9Frame) {
+        isKeyFrame = (data.peekUnsignedByte() & 0x04) == 0;
+      }
+
+      if (!isOutputFormatSet && width > 0 && height > 0) {
```
claincly, line 92: You initialized width/height to `UNSET`, you can check it using `UNSET` here like `width != UNSET`, rather than `width > 0`.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+      int fragmentSize = data.bytesLeft();
+      // Write the video sample.
+      trackOutput.sampleData(data, fragmentSize);
+      fragmentedSampleSizeBytes += fragmentSize;
+
+      if (rtpMarker) {
+        if (firstReceivedTimestamp == C.TIME_UNSET) {
+          firstReceivedTimestamp = timestamp;
+        }
+        bufferFlags = isKeyFrame ? C.BUFFER_FLAG_KEY_FRAME : 0;
```
claincly, line 109: You can inline this and remove `bufferFlags`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,230 @@
...
+        long timeUs = toSampleUs(startTimeOffsetUs, timestamp, firstReceivedTimestamp);
+        trackOutput.sampleMetadata(
+            timeUs,
+            bufferFlags,
+            fragmentedSampleSizeBytes,
+            /* offset= */ 0,
+            /* encryptionData= */ null);
```
claincly, line 116: `/* cryptoData= */ null);`

claincly (changes requested):
Please also rebase to the newest head in `main`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
rakeshnitb: I'm following RFC dictionary for logging.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
rakeshnitb: Live555 doesn't support flexible mode, there are only reading and passing required Payload Descriptor (|I|P|L|F|B|E|V|Z|).

claincly:
looks good, will start merging process internally

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,240 @@
...
+
+      if ((scalabilityStructure & 0x10) != 0) {
+        if (payload.bytesLeft() < scalabilityStructureLength * SCALABILITY_STRUCTURE_SIZE) {
+          return false;
+        }
+        for (int index = 0; index < scalabilityStructureLength; index++) {
+          width = payload.readUnsignedShort();
+          height = payload.readUnsignedShort();
+        }
+      }
```
claincly, line 236: I think it's also worth adding the code that checks `G` bit, or `N_G`. Can you test if this works?

```
      // Check G bit, skips all additional temporal layers.
      if ((scalabilityStructure & 0x08) != 0) {
        // Reads N_G.
        int picturesCount = payload.readUnsignedByte();
        payload.skipBytes(picturesCount);
      }
```

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVP9Reader.java
```diff
@@ -0,0 +1,240 @@
...
+        trackOutput.sampleMetadata(
+            timeUs,
+            isKeyFrame ? C.BUFFER_FLAG_KEY_FRAME : 0,
+            fragmentedSampleSizeBytes,
+            /* offset= */ 0,
+            /* cryptoData= */ null);
+        fragmentedSampleSizeBytes = 0;
+        gotFirstPacketOfVP9Frame = false;
+      }
+      previousSequenceNumber = sequenceNumber;
```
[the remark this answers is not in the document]
rakeshnitb: By moving "previousSequenceNumber = sequenceNumber;" down, packet(s) with Invalid Payload Descriptor will also get logged in previousSequenceNumber which can lead to directly get follow-on packet without start Code packet.

PR #71, opened 2022-04-24
Workaround OkHttp lack of clean interrupt support

yschimke:
Workaround for https://github.com/square/okhttp/issues/3146

There is a draft PR https://github.com/square/okhttp/pull/7185/files which documents OkHttp's ideal handling of cancellation including interrupts.

But a few key points

1) This is a target state, and OkHttp does not currently handle interrupts correctly.  In the past this has been identified, and the advice is to avoid interrupts on Http threads, see discussion on https://github.com/square/okhttp/issues/1902. Also an attempt at a fix here https://github.com/square/okhttp/pull/7023 which wasn't in a form to land.

2) Even with this fixed, it is likely to never be optimal, because of OkHttp sharing a socket connection for multiple inflight requests.

From https://github.com/square/okhttp/pull/7185

```
Thread.interrupt() is Clumsy
----------------------------

`Thread.interrupt()` is Java's built-in mechanism to cancel an in-flight `Thread`, regardless of
what work it's currently performing.

We recommend against using `Thread.interrupt()` with OkHttp because it may disrupt shared resources
including HTTP/2 connections and cache files. In particular, calling `Thread.interrupt()` may cause
unrelated threads' call to fail with an `IOException`.
```

This PR leaves the Loader/DataSource thread parked on a countdown latch, while this may seem wasteful and an additional context switch. However in practice the response isn't returned until the Http2Connection and Http2Reader have a response from the server and these means effectively parking in a `wait()` statement here https://github.com/square/okhttp/blob/9e039e94123defbfd5f11dc64ae146c46b7230eb/okhttp/src/jvmMain/kotlin/okhttp3/internal/http2/Http2Stream.kt#L140

TODO

- [ ] test in CI
- [ ] Verify (again) that it fixes the demo-session example with frequent scrubbing.

### DIFF
--- a/libraries/datasource_okhttp/src/androidTest/java/androidx/media3/datasource/okhttp/OkHttpDataSourceContractTest.kt
+++ b/libraries/datasource_okhttp/src/androidTest/java/androidx/media3/datasource/okhttp/OkHttpDataSourceContractTest.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2020 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package androidx.media3.datasource.okhttp
+
+import android.net.Uri
+import androidx.media3.datasource.DataSource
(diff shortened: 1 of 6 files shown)


libraries/datasource_okhttp/src/androidTest/java/androidx/media3/datasource/okhttp/OkHttpDataSourceContractTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.google.common.collect.ImmutableList
+import okhttp3.Call
+import okhttp3.OkHttpClient
+import org.junit.Rule
+import org.junit.runner.RunWith
+
+/** [DataSource] contract tests for [OkHttpDataSource].  */
+@RunWith(AndroidJUnit4::class)
+class OkHttpDataSourceContractTest : DataSourceContractTest() {
```
yschimke, line 31: I can delete this, since it's covered in normal tests.

ojw28:
Closing in favor of submitting internally. Thanks!

PR #96, opened 2022-06-28
Annotated result of StubPlayer.getPlayerError with @Nullable

fengdai:


marcbaechinger:
Thanks for the fix!

When doing a pull request, you need to request from and against the `main` branch. A pull request results in a commit on the `main` branch and will then be merged into the `release` branch with the next release process.

For the case of this CL you can chose to send it again compared and requested against the `main` branch. I can apply this change internally, as it is easy to do. It won't be credited/attributed to your GitHub account in the latter case. So please let me know what you prefer. Thanks again for using Media3!

fengdai:
I've changed the target branch to `main`.

PR #110, opened 2022-07-06
Add RTP VP8 test

basantwanishraddha:
Exoplayer RTP VP8 Reader had few issues.
Created a [issue](https://partnerissuetracker.corp.google.com/u/0/issues/238153477) describing the same.

Fixed the reported issue in the PR and added RTP VP8 tests as well.


claincly (approved):
Looks good overall, merging internally.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVp8Reader.java
```diff
@@ -195,6 +200,23 @@ private boolean validateVp8Descriptor(ParsableByteArray payload, int packetSeque
...
+   * <p>Call this method only when receiving a end of VP8 partition
+   */
+  private void outputSampleMetadataForFragmentedPackets() {
+    trackOutput.sampleMetadata(
+        sampleTimeUsOfFragmentedSample,
+        isKeyFrame ? C.BUFFER_FLAG_KEY_FRAME : 0,
+        fragmentedSampleSizeBytes,
+        /* offset= */ 0,
+        /* cryptoData= */ null);
+    fragmentedSampleSizeBytes = 0;
```
claincly, line 215: It seems if `fragmentedSampleSizeBytes` is reset to zero here, the logic on L100 will run on 

- The first received RTP packet, or
- The first received RTP packet after a seek.

Is that correct? If so, is it intentional?

claincly: Like I think what you did in #115 looks "more correct"?

basantwanishraddha:
The merge commit 9d9bbe3d33721b1301ae16df7f87f9aa51d06b89, addresses the two review comments. We can skip these 2 commits.

claincly:
Thanks for making the change after. Could you create a new PR for the two recent commits, as the original PR is already merged into `main`?

basantwanishraddha:
The review comments addressed in the recent two commits are already resolved in commit 9d9bbe3d33721b1301ae16df7f87f9aa51d06b89 and merged on main. We can discard the recent two commits.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVp8Reader.java
claincly: I was asking a yes-or-no question here - the logic 

```java
      if (fragmentedSampleSizeBytes == C.LENGTH_UNSET && gotFirstPacketOfVp8Frame) {
        isKeyFrame = (data.peekUnsignedByte() & 0x01) == 0;
      }
```

runs on the first received RTP packet, or the first packet arrives after a seek ONLY, because you reset it to 0 in `outputSampleMetadataForFragmentedPackets` rather than `LENGTH_UNSET`.

**Is that correct? If so, is it intentional?**

basantwanishraddha: It was not intentional, fragmentedSampleSizeBytes should have been reset with LENGTH_UNSET. It should run on every first received RTP packet of a fragment.

claincly: Thanks for the clarification.

claincly:
Already merged: internal 460513413

PR #115, opened 2022-07-08
Update VP9 Reader to handle missing frames/fragments.

ManishaJajoo:
Change-Id: I3cdc738b25fda1292ae07769656a3b18721b2402

### DIFF
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVp9Reader.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVp9Reader.java
@@ -16,6 +16,7 @@
 package androidx.media3.exoplayer.rtsp.reader;
 
 import static androidx.media3.common.util.Assertions.checkArgument;
+import static androidx.media3.common.util.Assertions.checkState;
 import static androidx.media3.common.util.Assertions.checkStateNotNull;
 
 import androidx.media3.common.C;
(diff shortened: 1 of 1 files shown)


libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpVp9Reader.java
```diff
@@ -85,19 +91,19 @@ public void createTracks(ExtractorOutput extractorOutput, int trackId) {
...
     checkStateNotNull(trackOutput);
 
     if (validateVp9Descriptor(data, sequenceNumber)) {
-      @C.BufferFlags int bufferFlags = 0;
-      if (fragmentedSampleSizeBytes == 0
-          && gotFirstPacketOfVP9Frame
-          && (data.peekUnsignedByte() & 0x04) == 0) {
-        bufferFlags = C.BUFFER_FLAG_KEY_FRAME;
+      if (fragmentedSampleSizeBytes == 0 && gotFirstPacketOfVP9Frame) {
+        isKeyFrame = (data.peekUnsignedByte() & 0x04) == 0;
```
claincly, line 106: I'm assuming you are parsing the VP9 payload descriptor's E bit, which is 

> E:  End of a frame.  MUST be set to 1 for the final RTP packet of a
      VP9 frame, and 0 otherwise.  This enables a decoder to finish
      decoding the frame, where it otherwise may need to wait for the
      next packet to explicitly know that the frame is complete.  Note
      that, if spatial scalability is in use, more frames from the same
      picture may follow; see the description of the B bit above.

I think this bit equals to zero is not the sufficient condition that the current frame is a key frame, no?

claincly: SG, thanks

claincly:
I noticed the handling of `fragmentedSampleSizeBytes` is a bit different from #110, where `fragmentedSampleSizeBytes` is initialized to `LENGTH_UNSET`. Can you do the similar here?

claincly:
No need to change - merged already

PR #119, opened 2022-07-14
Rtp h263 test and fix

ManishaJajoo:


libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH263ReaderTest.java
```diff
@@ -0,0 +1,198 @@
...
+      new RtpPacket.Builder()
+          .setTimestamp((int) 2599168056L)
+          .setSequenceNumber(40290)
+          .setMarker(true)
+          .setPayloadData(
+              Bytes.concat(
+                  /*payload header */ getBytesFromHexString("0000"), FRAME_1_FRAGMENT_2_DATA))
+          .build();
+  private static final byte[] FRAME_1_DATA =
+      Bytes.concat(getBytesFromHexString("0000"), FRAME_1_FRAGMENT_1_DATA, FRAME_1_FRAGMENT_2_DATA);
```
claincly, line 66: curious, why the data needs `0000` in the beginning?

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH263ReaderTest.java
```diff
@@ -0,0 +1,198 @@
...
+import org.mockito.junit.MockitoRule;
+
+/**
+ * Unit test for {@link RtpH263Reader}.
+ */
+@RunWith(AndroidJUnit4.class)
+public final class RtpH263ReaderTest {
+  private static final byte[] FRAME_1_FRAGMENT_1_DATA =
+      getBytesFromHexString("80020c0419b7b7d9591f03023e0c37b");
+  private final RtpPacket FRAME_1_FRAGMENT_1 =
```
claincly, line 45: can you make the field `static final`?

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH263ReaderTest.java
ManishaJajoo: According to RFC4629 Section 6.1.1.  " A packet that begins at the location of a Picture, GOB, slice, EOS,
or EOSBS start code shall omit the first two (all zero) bytes from the H.263+ bitstream", H263 reader changes the payload header data (i.e 0400) to 0000. So the final output contains 0000 + frame data.

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH263ReaderTest.java
ManishaJajoo: Since this data gets changed by the reader, making it static will affect other tests.

claincly: Do you mean that the tests will see different payload data even when they reference `FRAME_1_FRAGMENT_1`?

Like `FRAME_1_FRAGMENT_1` is different in `consume_validPackets` and `consume_outOfOrderPackets`?

ManishaJajoo: Yes. After the first test, FRAME_1_FRAGMENT_1 will change. It will be 0000 + frame_1_fragment_1_data + frame_1_fragment_2_data. 
The payload header part will no longer be 0400. It will be 0000.

claincly: I don't think it's good practice to have that in test then. A reader would expect the test to run independently, like if I see `consume_outOfOrderPackets` references `FRAME_1_FRAGMENT_1`, I expect the payload header to be 0400.

On a more general note, `FRAME_1_FRAGMENT_1` looks like a constant and let's keep it constant. The tests shouldn't have inner dependencies. Imagine one day I move the tests around, and they are not run in the current order.

ManishaJajoo: That makes sense. I have changed the reader to keep the input constant. I have updated the test as well. Please have a look

claincly: Thanks for making the change, I think we can do it another way though. Your current approach involves in copying the all the received RTP packets again, which increases the pressure on the memory system. The newly introduced memory copy is only needed in tests, but is not actually needed during normal operation.

What about implement a method that copies the packets in tests every time it's referenced? Like 
```java
private static RtpPacket copyPacket(RtpPacket packet) {
  RtpPacket.Builder builder =
      new RtpPacket.Builder()
          .setPadding(packet.padding)
          .setMarker(packet.marker)
          .setPayloadType(packet.payloadType)
          .setSequenceNumber(packet.sequenceNumber)
          .setTimestamp(packet.timestamp)
          .setSsrc(packet.ssrc);

  if (packet.csrc.length > 0) {
    builder.setCsrc(Arrays.copyOf(packet.csrc, packet.csrc.length));
  }
  if (packet.payloadData.length > 0) {
    builder.setPayloadData(Arrays.copyOf(packet.payloadData, packet.payloadData.length));
  }
  return builder.build();
}
```

claincly:
Merged internally.

PR #132, opened 2022-07-30
Disable Channel badges for API 26 and lower

### LINKED ISSUE #131: API <=26: MediaNotification will add a Badge to the App Icon
New APIs will never show a badge for a MediaStyle notification. It seems that for older API versions it needs to be deactivated manually.

### Media3 Version
1.0.0-beta02

tzugen:
Fixes #131

### DIFF
--- a/libraries/session/src/main/java/androidx/media3/session/DefaultMediaNotificationProvider.java
+++ b/libraries/session/src/main/java/androidx/media3/session/DefaultMediaNotificationProvider.java
@@ -374,6 +374,7 @@ public class DefaultMediaNotificationProvider implements MediaNotification.Provi
     NotificationChannel channel =
         new NotificationChannel(
             NOTIFICATION_CHANNEL_ID, NOTIFICATION_CHANNEL_NAME, NotificationManager.IMPORTANCE_LOW);
+    channel.setShowBadge(false);
     notificationManager.createNotificationChannel(channel);
   }
 


libraries/session/src/main/java/androidx/media3/session/DefaultMediaNotificationProvider.java
```diff
@@ -374,6 +374,7 @@ private void ensureNotificationChannel() {
     NotificationChannel channel =
         new NotificationChannel(
             NOTIFICATION_CHANNEL_ID, NOTIFICATION_CHANNEL_NAME, NotificationManager.IMPORTANCE_LOW);
+    channel.setShowBadge(false);
```
icbaker, line 377: The linked issue (https://github.com/androidx/media/issues/131) and the title of this PR indicate you expect this to only affect API 26 and below, but this line is only executed on API 26 and above (since API < 26 will return early above on L372).

So is it more correct to describe this as a problem specific to API 26? If so, it might be clearer to gate this with `Util.SDK_INT == 26` with a comment.

Actually: I did some quick testing on a Pixel 2 emulator on different API levels, I see a badge on API 26 and 27, but not 28 - do you agree with that? In that case, let's gate it on `Util.SDK_INT == 26 || Util.SDK_INT == 27`

tzugen: Yes I share your observation.
I am not sure I understand why the code to setup the channel is not executed on ApI 26, because Channels were introduced with API 26...

icbaker: > I am not sure I understand why the code to setup the channel is not executed on ApI 26, because Channels were introduced with API 26...

Why do you believe that to be the case? This code is executed when `Util.SDK_INT < 26` is `false`, i.e. when `SDK_INT >= 26` is `true`, i.e. it **is** executed on API 26.

icbaker:
I will merge this with the additional gating and comment - no need to add any more commits here.

icbaker:
Ah, just spotted this PR is made against the `release` branch, could you update it to be against the `main` branch as described in https://github.com/androidx/media/blob/release/CONTRIBUTING.md#pull-requests? I tried just changing it in the GitHub UI but it created a lot of conflicts - it may be easier to just open a new one and close this one?

tzugen:
Closing, see #141

Issue #137, opened 2022-08-01
Make customization of OkHttpDataSource easier


tzugen:
I want to extend OkHttpDataSource, because we are creating our calls with Retrofit.

Suggestions:
- Remove "private" visibility from executeCall() & closeConnectionQuietly(), makeRequest(), so that our overridden functions can use them or override them.
- Remove "private" visibility from dataSpec, bytesRead, bytesToRead, callFactory, response, responseByteStream, opened

icbaker:
Please can you add some more detail about what you're trying to achieve, rather than focussing on how to do it (extending `OkHttpDataSource`).

We generally prefer not to make our types extensible, and where we do make them extensible we try and make the contract between layers of the inheritance very clear (see my comment on the linked PR for more detail) - so we'd prefer to understand what you're trying to implement at a higher level, so that we can suggest possible alternatives that don't require opening up significant implementation details of `OkHttpDatasSource`.

tzugen:
We are using Retrofit because it has a very nice API. It is based on Ok, but the way you invoke it is different. You dont construct requests manually, but define API endpoints and their parameters, and then call these like you woukd call any other function.

So right now we have a nearly identical copy of the OkHttpDataFactory, with only a few lines of difference inside the open() function. The open() function is actually extendable, but we cant access any of the properties and internal functions so its of little use.

tzugen:
But obviously this is annoying to maintain..

icbaker:
> We are using Retrofit because it has a very nice API.

It's not obvious to me when an app using media3 needs to directly interact with the API of any HTTP stack, since everything is abstracted via [`DataSource`](https://github.com/androidx/media/blob/release/libraries/datasource/src/main/java/androidx/media3/datasource/DataSource.java). I would expect the choice between HTTP stacks would be based on their behavior and features, rather than their API - so if Retrofit uses OkHttp under the covers, then what's the difference from a media3-perspective between using it and just using media3's `OkHttpDataSource`?

tzugen:
Ok, maybe we are using it wrong. We just put the id of the requested file and the bitrate into the dataSpec and resolve the uri in the DataSource with Retrofit.

icbaker:
Would [`ResolvingDataSource`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/upstream/ResolvingDataSource.html) work for that?

tzugen:
Perfect, I managed to get this to work quiet quickly. Thank you!!

PR #138, opened 2022-08-01
Make a bunch of properties and functions accessible for extending cla…

tzugen:
See the bug report  #137


icbaker:
Please see my replies on https://github.com/google/ExoPlayer/pull/10423 for why we are unlikely to accept this PR in its current form.

PR #149, opened 2022-08-15
DefaultMediaNotificationProvider: Fix missing comma in documentation.

vanniktech:
Originally I wanted to extend `DefaultMediaNotificationProvider` such that I can pass custom channel id / name etc. Then I saw that this is already being done in `main` branch. 

Any chance when there's a new release with the new bits from `DefaultMediaNotificationProvider`?


rohitjoins:
Thank you for the PR. We don't have any planned releases as of now. However, our next release will contain the new bits from `DefaultMediaNotificationProvider `.

PR #150, opened 2022-08-15
DefaultMediaNotificationProvider: Use metadata.subtitle as content text of notification.

vanniktech:
Why is the artist currently being used? It took me quite a while to figure out and my first instinct would have been to show the subtitle.


vanniktech:
Android Auto also uses subtitle for their list item, e.g.:

Screen Shot 2022-08-15 at 20 32 17

rohitjoins:
@vanniktech Thank you for this PR. I'm going to close this in favour of providing methods to override content title and content text. Will update here once I merge those changes.

PR #162, opened 2022-09-04
Add support for RTSP Mp4a-Latm

rakeshnitb:
Added Mp4a-latm RTP Packet reader and added support for Mp4a-latm
playback through RTSP.

Change-Id: Ia590393f53ca880af926907843f6bea9ff0f4b35

### DIFF
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
@@ -219,12 +219,12 @@ import com.google.common.collect.ImmutableMap;
           }
           checkArgument(!isConfigPresent, "cpresent == 0 means we need to parse config");
           @Nullable String configInput = fmtpParameters.get(PARAMETER_MP4V_CONFIG);
-          if (configInput != null && configInput.length() % 2 == 0) {
-            Pair<Integer, Integer> configParameters = getSampleRateAndChannelCount(configInput);
-            channelCount = configParameters.first;
-            clockRate = configParameters.second;
-            formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);
-          }
+          checkArgument(configInput != null && configInput.length() % 2 == 0,
+              "only support out-band mode");
+          Pair<Integer, Integer> configParameters = getSampleRateAndChannelCount(configInput);
+          channelCount = configParameters.first;
+          clockRate = configParameters.second;
+          formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);
         }
         processAacFmtpAttribute(formatBuilder, fmtpParameters, channelCount, clockRate);
         break;
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
@@ -55,7 +55,7 @@ import com.google.common.collect.ImmutableMap;
   private static final String PARAMETER_H265_SPROP_VPS = "sprop-vps";
   private static final String PARAMETER_H265_SPROP_MAX_DON_DIFF = "sprop-max-don-diff";
   private static final String PARAMETER_MP4V_CONFIG = "config";
-  private static final String PARAMETER_MP4A_CPRESENT = "cpresent";
+  private static final String PARAMETER_MP4A_C_PRESENT = "cpresent";
 
   /** Prefix for the RFC6381 codecs string for AAC formats. */
   private static final String AAC_CODECS_PREFIX = "mp4a.40.";
@@ -212,13 +212,13 @@ import com.google.common.collect.ImmutableMap;
         checkArgument(channelCount != C.INDEX_UNSET);
         checkArgument(!fmtpParameters.isEmpty());
         if(mediaEncoding.equals(RtpPayloadFormat.RTP_MEDIA_MPEG4_AUDIO)) {
-          boolean isCPresent = true;
-          if (fmtpParameters.get(PARAMETER_MP4A_CPRESENT) != null && fmtpParameters.get(
-              PARAMETER_MP4A_CPRESENT).equals("0")) {
-            isCPresent = false;
+          boolean isConfigPresent  = true;
+          if (fmtpParameters.get(PARAMETER_MP4A_C_PRESENT) != null && fmtpParameters.get(
+              PARAMETER_MP4A_C_PRESENT).equals("0")) {
+            isConfigPresent  = false;
           }
           @Nullable String configInput = fmtpParameters.get(PARAMETER_MP4V_CONFIG);
-          if (!isCPresent && configInput != null && configInput.length() % 2 == 0) {
+          if (!isConfigPresent  && configInput != null && configInput.length() % 2 == 0) {
             Pair<Integer, Integer> configParameters = getSampleRateAndChannelCount(configInput);
             channelCount = configParameters.first;
             clockRate = configParameters.second;
@@ -326,25 +326,22 @@ import com.google.common.collect.ImmutableMap;
    * configuration (ISO/IEC14496-3, Chapter 1.7.3).
    */
   private static Pair<Integer, Integer> getSampleRateAndChannelCount(String configInput) {
-    int channelCount = 0, sampleRate = 0;
-    byte[] configBuffer = Util.getBytesFromHexString(configInput);
-    ParsableBitArray scratchBits = new ParsableBitArray(configBuffer);
-    int audioMuxVersion = scratchBits.readBits(1);
+    ParsableBitArray config  = new ParsableBitArray(Util.getBytesFromHexString(configInput));
+    int audioMuxVersion = config .readBits(1);
     if (audioMuxVersion == 0) {
-      checkArgument(scratchBits.readBits(1) == 1, "Invalid allStreamsSameTimeFraming.");
-      scratchBits.readBits(6);
-      checkArgument(scratchBits.readBits(4) == 0, "Invalid numProgram.");
-      checkArgument(scratchBits.readBits(3) == 0, "Invalid numLayer.");
+      checkArgument(config .readBits(1) == 1, "Only supports one allStreamsSameTimeFraming.");
+      config .readBits(6);
+      checkArgument(config .readBits(4) == 0, "Only supports one program.");
+      checkArgument(config .readBits(3) == 0, "Only supports one numLayer.");
       @Nullable AacUtil.Config aacConfig = null;
       try {
-        aacConfig = AacUtil.parseAudioSpecificConfig(scratchBits, false);
+        aacConfig = AacUtil.parseAudioSpecificConfig(config , false);
       } catch (ParserException e) {
         throw new IllegalArgumentException(e);
       }
-      sampleRate = aacConfig.sampleRateHz;
-      channelCount = aacConfig.channelCount;
+      return Pair.create(aacConfig.channelCount, aacConfig.sampleRateHz);
     }
-    return Pair.create(channelCount, sampleRate);
+    throw new IllegalArgumentException ("Only support audio mux version 0");
   }
 
   private static void processMPEG4FmtpAttribute(
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
@@ -267,7 +267,8 @@ import com.google.common.collect.ImmutableMap;
     }
 
     checkArgument(clockRate > 0);
-    return new RtpPayloadFormat(formatBuilder.build(), rtpPayloadType, clockRate, fmtpParameters);
+    return new RtpPayloadFormat(
+        formatBuilder.build(), rtpPayloadType, clockRate, fmtpParameters, mediaEncoding);
   }
 
   private static int inferChannelCount(int encodingParameter, String mimeType) {
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
@@ -31,7 +31,9 @@ import androidx.annotation.VisibleForTesting;
 import androidx.media3.common.C;
 import androidx.media3.common.Format;
 import androidx.media3.common.MimeTypes;
+import androidx.media3.common.ParserException;
 import androidx.media3.common.util.CodecSpecificDataUtil;
+import androidx.media3.common.util.ParsableBitArray;
 import androidx.media3.common.util.UnstableApi;
 import androidx.media3.common.util.Util;
 import androidx.media3.extractor.AacUtil;
@@ -208,6 +210,13 @@ import com.google.common.collect.ImmutableMap;
       case MimeTypes.AUDIO_AAC:
         checkArgument(channelCount != C.INDEX_UNSET);
         checkArgument(!fmtpParameters.isEmpty());
+        if(mediaEncoding.equals(RtpPayloadFormat.RTP_MEDIA_MPEG4_AUDIO)) {
+          Pair<Integer, Integer> cfgOpts = getSampleRateAndChannelCountFromAudioConfig(
+              fmtpParameters, channelCount, clockRate);
+          channelCount = cfgOpts.first;
+          clockRate = cfgOpts.second;
+          formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);
+        }
         processAacFmtpAttribute(formatBuilder, fmtpParameters, channelCount, clockRate);
         break;
       case MimeTypes.AUDIO_AMR_NB:
@@ -299,6 +308,38 @@ import com.google.common.collect.ImmutableMap;
         ImmutableList.of(
             // Clock rate equals to sample rate in RTP.
             AacUtil.buildAacLcAudioSpecificConfig(sampleRate, channelCount)));
+  }
+
+  /**
+   * Parses an MPEG-4 Audio Stream Mux configuration, as defined in ISO/IEC14496-3. FMTP attribute
+   * contains config which is a byte array containing the MPEG-4 Audio Stream Mux configuration to
+   * parse.
+   */
+  private static Pair<Integer, Integer> getSampleRateAndChannelCountFromAudioConfig(
+      ImmutableMap<String, String> fmtpAttributes,
+      int channelCount,
+      int sampleRate) {
+    @Nullable String configInput = fmtpAttributes.get(PARAMETER_MP4V_CONFIG);
+    if (configInput != null && configInput.length() % 2 == 0) {
+      byte[] configBuffer = Util.getBytesFromHexString(configInput);
+      ParsableBitArray scratchBits = new ParsableBitArray(configBuffer);
+      int audioMuxVersion = scratchBits.readBits(1);
+      if (audioMuxVersion == 0) {
+        checkArgument(scratchBits.readBits(1) == 1, "Invalid allStreamsSameTimeFraming.");
+        scratchBits.readBits(6);
+        checkArgument(scratchBits.readBits(4) == 0, "Invalid numProgram.");
+        checkArgument(scratchBits.readBits(3) == 0, "Invalid numLayer.");
+        AacUtil.Config aacConfig = null;
+        try {
+          aacConfig = AacUtil.parseAudioSpecificConfig(scratchBits, false);
+        } catch (ParserException e) {
+          throw new IllegalArgumentException(e);
+        }
+        sampleRate = aacConfig.sampleRateHz;
+        channelCount = aacConfig.channelCount;
+      }
+    }
+    return Pair.create(channelCount, sampleRate);
   }
 
   private static void processMPEG4FmtpAttribute(
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
@@ -217,8 +217,9 @@ import com.google.common.collect.ImmutableMap;
               PARAMETER_MP4A_C_PRESENT).equals("0")) {
             isConfigPresent  = false;
           }
+          checkArgument(!isConfigPresent, "cpresent == 0 means we need to parse config");
           @Nullable String configInput = fmtpParameters.get(PARAMETER_MP4V_CONFIG);
-          if (!isConfigPresent  && configInput != null && configInput.length() % 2 == 0) {
+          if (configInput != null && configInput.length() % 2 == 0) {
             Pair<Integer, Integer> configParameters = getSampleRateAndChannelCount(configInput);
             channelCount = configParameters.first;
             clockRate = configParameters.second;
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
@@ -55,6 +55,7 @@ import com.google.common.collect.ImmutableMap;
   private static final String PARAMETER_H265_SPROP_VPS = "sprop-vps";
   private static final String PARAMETER_H265_SPROP_MAX_DON_DIFF = "sprop-max-don-diff";
   private static final String PARAMETER_MP4V_CONFIG = "config";
+  private static final String PARAMETER_MP4A_CPRESENT = "cpresent";
 
   /** Prefix for the RFC6381 codecs string for AAC formats. */
   private static final String AAC_CODECS_PREFIX = "mp4a.40.";
@@ -211,11 +212,18 @@ import com.google.common.collect.ImmutableMap;
         checkArgument(channelCount != C.INDEX_UNSET);
         checkArgument(!fmtpParameters.isEmpty());
         if(mediaEncoding.equals(RtpPayloadFormat.RTP_MEDIA_MPEG4_AUDIO)) {
-          Pair<Integer, Integer> cfgOpts = getSampleRateAndChannelCountFromAudioConfig(
-              fmtpParameters, channelCount, clockRate);
-          channelCount = cfgOpts.first;
-          clockRate = cfgOpts.second;
-          formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);
+          boolean isCPresent = true;
+          if (fmtpParameters.get(PARAMETER_MP4A_CPRESENT) != null && fmtpParameters.get(
+              PARAMETER_MP4A_CPRESENT).equals("0")) {
+            isCPresent = false;
+          }
+          @Nullable String configInput = fmtpParameters.get(PARAMETER_MP4V_CONFIG);
+          if (!isCPresent && configInput != null && configInput.length() % 2 == 0) {
+            Pair<Integer, Integer> configParameters = getSampleRateAndChannelCount(configInput);
+            channelCount = configParameters.first;
+            clockRate = configParameters.second;
+            formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);
+          }
         }
         processAacFmtpAttribute(formatBuilder, fmtpParameters, channelCount, clockRate);
         break;
@@ -311,33 +319,30 @@ import com.google.common.collect.ImmutableMap;
   }
 
   /**
-   * Parses an MPEG-4 Audio Stream Mux configuration, as defined in ISO/IEC14496-3. FMTP attribute
-   * contains config which is a byte array containing the MPEG-4 Audio Stream Mux configuration to
-   * parse.
+   * Returns a {@link Pair} of sample rate and channel count, by parsing the
+   *  MPEG4 Audio Stream Mux configuration.
+   *
+   * <p>fmtp attribute {@code config} includes the MPEG4 Audio Stream Mux
+   * configuration (ISO/IEC14496-3, Chapter 1.7.3).
    */
-  private static Pair<Integer, Integer> getSampleRateAndChannelCountFromAudioConfig(
-      ImmutableMap<String, String> fmtpAttributes,
-      int channelCount,
-      int sampleRate) {
-    @Nullable String configInput = fmtpAttributes.get(PARAMETER_MP4V_CONFIG);
-    if (configInput != null && configInput.length() % 2 == 0) {
-      byte[] configBuffer = Util.getBytesFromHexString(configInput);
-      ParsableBitArray scratchBits = new ParsableBitArray(configBuffer);
-      int audioMuxVersion = scratchBits.readBits(1);
-      if (audioMuxVersion == 0) {
-        checkArgument(scratchBits.readBits(1) == 1, "Invalid allStreamsSameTimeFraming.");
-        scratchBits.readBits(6);
-        checkArgument(scratchBits.readBits(4) == 0, "Invalid numProgram.");
-        checkArgument(scratchBits.readBits(3) == 0, "Invalid numLayer.");
-        AacUtil.Config aacConfig = null;
-        try {
-          aacConfig = AacUtil.parseAudioSpecificConfig(scratchBits, false);
-        } catch (ParserException e) {
-          throw new IllegalArgumentException(e);
-        }
-        sampleRate = aacConfig.sampleRateHz;
-        channelCount = aacConfig.channelCount;
+  private static Pair<Integer, Integer> getSampleRateAndChannelCount(String configInput) {
+    int channelCount = 0, sampleRate = 0;
+    byte[] configBuffer = Util.getBytesFromHexString(configInput);
+    ParsableBitArray scratchBits = new ParsableBitArray(configBuffer);
+    int audioMuxVersion = scratchBits.readBits(1);
+    if (audioMuxVersion == 0) {
+      checkArgument(scratchBits.readBits(1) == 1, "Invalid allStreamsSameTimeFraming.");
+      scratchBits.readBits(6);
+      checkArgument(scratchBits.readBits(4) == 0, "Invalid numProgram.");
+      checkArgument(scratchBits.readBits(3) == 0, "Invalid numLayer.");
+      @Nullable AacUtil.Config aacConfig = null;
+      try {
+        aacConfig = AacUtil.parseAudioSpecificConfig(scratchBits, false);
+      } catch (ParserException e) {
+        throw new IllegalArgumentException(e);
       }
+      sampleRate = aacConfig.sampleRateHz;
+      channelCount = aacConfig.channelCount;
     }
     return Pair.create(channelCount, sampleRate);
   }
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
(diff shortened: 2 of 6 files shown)


libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
```diff
@@ -0,0 +1,157 @@
...
+
+import com.google.common.collect.ImmutableMap;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+
+/**
+ * Parses an MP4A-LATM byte stream carried on RTP packets, and extracts MP4A-LATM Access Units.
+ * Refer to RFC3016 for more details.
+ */
+@UnstableApi
+/* package */ final class RtpMp4aPayloadReader  implements RtpPayloadReader {
```
claincly, line 42: rename to `RtpMp4aReader` to keep consistent. Also, remove the extra space before `implements`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
```diff
@@ -0,0 +1,157 @@
...
+
+  @Override
+  public void seek(long nextRtpTimestamp, long timeUs) {
+    firstReceivedTimestamp = nextRtpTimestamp;
+    fragmentedSampleSizeBytes = 0;
+    startTimeOffsetUs = timeUs;
+  }
+
+  // Internal methods.
+  private static long toSampleUs(
```
claincly, line 118: remove and use `import static androidx.media3.exoplayer.rtsp.reader.RtpReaderUtils.toSampleTimeUs;`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -267,7 +267,8 @@ public int hashCode() {
     }
 
     checkArgument(clockRate > 0);
-    return new RtpPayloadFormat(formatBuilder.build(), rtpPayloadType, clockRate, fmtpParameters);
+    return new RtpPayloadFormat(
+        formatBuilder.build(), rtpPayloadType, clockRate, fmtpParameters, mediaEncoding);
```
claincly, line 271: Is `processAacFmtpAttribute` applied to both ADTS AAC (the one we already support) and LATM AAC (this one)?

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
```diff
@@ -0,0 +1,157 @@
...
+   * Parses an MPEG-4 Audio Stream Mux configuration, as defined in ISO/IEC14496-3. FMTP attribute
+   * contains config which is a byte array containing the MPEG-4 Audio Stream Mux configuration to
+   * parse.
+   *
+   * @param fmtpAttributes The format parameters, mapped from the SDP FMTP attribute.
+   * @return The number of subframes.
+   * @throws ParserException If the MPEG-4 Audio Stream Mux configuration cannot be parsed due to
+   *     unsupported audioMuxVersion.
+   */
+  public static Integer getNumOfSubframesFromMpeg4AudioConfig(
```
claincly, line 137: It seems the number of subframes is constant (it's from `fmtp`). If so please move this logic into `RtspMediaTrack` where we process `fmtp` for other formats. You can pass this number of subframes in from the constructor instead of calculating it every time.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
```diff
@@ -0,0 +1,157 @@
...
+      sampleOffset++;
+      data.setPosition(sampleOffset);
+
+      // Write the audio sample
+      trackOutput.sampleData(data, sampleLength);
+      sampleOffset += sampleLength;
+      fragmentedSampleSizeBytes += sampleLength;
+    }
+    if (rtpMarker) {
+      long timeUs =
```
claincly, line 102: inline?

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
```diff
@@ -0,0 +1,157 @@
...
+          toSampleUs(startTimeOffsetUs, timestamp, firstReceivedTimestamp, payloadFormat.clockRate);
+      trackOutput.sampleMetadata(
+          timeUs, C.BUFFER_FLAG_KEY_FRAME, fragmentedSampleSizeBytes, 0, null);
```
claincly, line 105: add `/* paramName= */` tags

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
```diff
@@ -0,0 +1,157 @@
...
+
+  @Override
+  public void consume(
+      ParsableByteArray data, long timestamp, int sequenceNumber, boolean rtpMarker)
+      throws ParserException {
+    checkStateNotNull(trackOutput);
+
+    int sampleOffset = 0;
+    int numSubFrames = getNumOfSubframesFromMpeg4AudioConfig(payloadFormat.fmtpParameters);
+    for (int subFrame = 0; subFrame <= numSubFrames; subFrame++) {
```
claincly, line 83: do you need to loop `numSubFrames + 1` times?

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
```diff
@@ -0,0 +1,157 @@
...
+      int sampleLength = 0;
+
+      /* Each subframe starts with a variable length encoding */
+      for (; sampleOffset < data.bytesLeft(); sampleOffset++) {
```
claincly, line 87: I find this piece of loop hard to follow - can you give an example data?

Say if `data.getData()[sampleOffset] & 0xff)` is `0xFF`, you still have to add it to the sampleLength?

claincly (changes requested):

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
harishdm: Luyuan, would it be okay if this is called the constructor in this file, so that the function remains private to this class and constructor arguments remain consistent with other readers?

claincly: That sounds good also!

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
rakeshnitb: Yes.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
rakeshnitb: I have called this function in onReceivingFirstPacket() function.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
rakeshnitb: sampleLength are save into byte address. If the size of frame is 261 then it is saved in array as [255, 6] So, I accumulating total size of the subFrame.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
rakeshnitb: With my understanding when the numSubFrames is zero then each RTP packet has one frame of data, similarly if the numSubFrames is 1 then we will have two frames.  Let me know if you think otherwise. 
We are unable to find any clip which has numSubFrames more than zero so this part of code is not tested. Can you share some clips which have non-zero numSubFrames.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
```diff
@@ -134,7 +140,7 @@ private static long toSampleUs(
    * @throws ParserException If the MPEG-4 Audio Stream Mux configuration cannot be parsed due to
    *     unsupported audioMuxVersion.
    */
-  public static Integer getNumOfSubframesFromMpeg4AudioConfig(
+  private static Integer getNumOfSubframesFromMpeg4AudioConfig(
```
claincly, line 143: return an int instead of Integer to save auto-boxing cost

claincly, line 143: You mentioned the streams you tested have got only 0 sub frames, does it mean that the parsing code in this method is never executed?

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
```diff
@@ -70,16 +77,24 @@ public void createTracks(ExtractorOutput extractorOutput, int trackId) {
   public void onReceivingFirstPacket(long timestamp, int sequenceNumber) {
     checkState(firstReceivedTimestamp == C.TIME_UNSET);
     firstReceivedTimestamp = timestamp;
-  }
+    try {
+      numSubFrames = getNumOfSubframesFromMpeg4AudioConfig(payloadFormat.fmtpParameters);
```
claincly, line 81: why not putting this in the constructor? it doesn't seem that f`fmtpParameters` can change in between packets. this way `numSubFrames` can be final

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
```diff
@@ -39,25 +41,30 @@
...
 
   private final RtpPayloadFormat payloadFormat;
   private @MonotonicNonNull TrackOutput trackOutput;
   private long firstReceivedTimestamp;
+  private int previousSequenceNumber;
   /** The combined size of a sample that is fragmented into multiple subFrames. */
   private int fragmentedSampleSizeBytes;
   private long startTimeOffsetUs;
+  private long sampleTimeUsOfFragmentedSample;
+  private int numSubFrames;
```
claincly, line 57: numberOfSubframes

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
claincly, line 271: should it also include the logic to accommodate RFC3016 Section 4.2:

> Unless specified by an out-of-band means, the resolution of the timestamp is set to its default value of 90 kHz.

claincly: >    If the clock rate has been set to its default value and it is
   necessary to obtain the audio sampling rate, this can be done by
   parsing the "config" parameter (see the following example).

      m=audio 49230 RTP/AVP 96
      a=rtpmap:96 MP4A-LATM/90000
      a=fmtp:96 object=8; cpresent=0; config=9128B1071070

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
claincly: although it's not ideal, you can throw a IAE wrapping the parserException.

Make it's caught in `RtspMediaPeriod.onLoadError()`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
claincly: can you share the spec that defines this? it doesn't seem that it's in RFC3016, maybe it's in ISO14496-3?

rakeshnitb: It is defined in ISO14496-3.
https://csclub.uwaterloo.ca/~ehashman/ISO14496-3-2009.pdf

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
rakeshnitb: After parsing we are getting number of subframes are zero.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
claincly: in which section specifically?

rakeshnitb: You can take a look at "Table 1.41 –Table 1.41 – Syntax of AudioMuxElement()".
From the Sec 1.7.3.2.3 numSubFrames is defined as:
"A data element indicating how many PayloadMux() frames are multiplexed (numSubFrames+1). If more than one PayloadMux() frame is multiplexed, all PayloadMux() share a common StreamMuxConfig().The minimum value is 0 indicating 1 subframe. "

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -208,6 +210,13 @@ public int hashCode() {
       case MimeTypes.AUDIO_AAC:
         checkArgument(channelCount != C.INDEX_UNSET);
         checkArgument(!fmtpParameters.isEmpty());
+        if(mediaEncoding.equals(RtpPayloadFormat.RTP_MEDIA_MPEG4_AUDIO)) {
+          Pair<Integer, Integer> cfgOpts = getSampleRateAndChannelCountFromAudioConfig(
```
claincly, line 214: rename to `sampleRateAndChannelCount` or similar

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
```diff
@@ -0,0 +1,179 @@
...
+        checkArgument(scratchBits.readBits(1) == 1, "Invalid allStreamsSameTimeFraming.");
+        numSubFrames = scratchBits.readBits(6);
+        checkArgument(scratchBits.readBits(4) == 0, "Invalid numProgram.");
+        checkArgument(scratchBits.readBits(3) == 0, "Invalid numLayer.");
+      } else {
+        throw ParserException.createForMalformedDataOfUnknownType(
+            "unsupported audio mux version: " + audioMuxVersion, null);
+      }
+    }
+    return numSubFrames;
```
claincly, line 161: As you mentioned, ISO14496 mentions (1.7.3.2.3) (put this into code comment)

> A data element indicating how many PayloadMux() frames are multiplexed (numSubFrames+1). If more than one PayloadMux() frame is multiplexed, all PayloadMux() share a common StreamMuxConfig().The minimum value is 0 indicating 1 subframe.

So you should return `numSubFrames + 1`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
```diff
@@ -0,0 +1,179 @@
...
+    if(fragmentedSampleSizeBytes > 0 && expectedSequenceNumber < sequenceNumber) {
+      outputSampleMetadataForFragmentedPackets();
+    }
+    int sampleOffset = 0;
+    for (int subFrame = 0; subFrame <= numSubFrames; subFrame++) {
+      int sampleLength = 0;
+
+      /* Each subframe starts with a variable length encoding */
+      for (; sampleOffset < data.bytesLeft(); sampleOffset++) {
+        sampleLength += data.getData()[sampleOffset] & 0xff;
```
claincly, line 103: I would do

```
// rename sampleOffset to audioPayloadOffset
int audioPayloadOffset;
for (audioPayloadOffset = 0; audioPayloadOffset < data.bytesLeft(); audioPayloadOffset++) {
  int payloadMuxLength = data.peekUnsignedByte();
  sampleLength += payloadMuxLength;
  if (payloadMuxLength != 0xff) {
    break;
  }
}
```

claincly (changes requested):

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
```diff
@@ -94,23 +94,30 @@ public void consume(
...
-      for (; sampleOffset < data.bytesLeft(); sampleOffset++) {
-        sampleLength += data.getData()[sampleOffset] & 0xff;
-        if ((data.getData()[sampleOffset] & 0xff) != 0xff) {
+      /**
+       * This implements PayloadLengthInfo() in Chapter 1.7.3.1, it's only support one program and
+       * one layer.
+       * Each subframe starts with a variable length encoding.
+       */
+      for (; audioPayloadOffset < data.bytesLeft(); audioPayloadOffset++) {
+        int payloadMuxLength = data.peekUnsignedByte();
```
claincly, line 107: my bad, you can actually use `.readUnsignedByte()` to save the increment on L112 (and maybe also 115)

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
```diff
@@ -94,23 +94,30 @@ public void consume(
...
+          data.setPosition(audioPayloadOffset + 1);
         }
       }
-      sampleOffset++;
-      data.setPosition(sampleOffset);
+      audioPayloadOffset++;
+      data.setPosition(audioPayloadOffset);
 
-      // Write the audio sample
+      /* Write the audio sample */
```
claincly, line 118: We use `//` for line comment.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -211,11 +212,18 @@ public int hashCode() {
         checkArgument(channelCount != C.INDEX_UNSET);
         checkArgument(!fmtpParameters.isEmpty());
         if(mediaEncoding.equals(RtpPayloadFormat.RTP_MEDIA_MPEG4_AUDIO)) {
-          Pair<Integer, Integer> cfgOpts = getSampleRateAndChannelCountFromAudioConfig(
-              fmtpParameters, channelCount, clockRate);
-          channelCount = cfgOpts.first;
-          clockRate = cfgOpts.second;
-          formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);
+          boolean isCPresent = true;
```
claincly, line 215: maybe rename to `isConfigPresent` for better readability

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -55,6 +55,7 @@
   private static final String PARAMETER_H265_SPROP_VPS = "sprop-vps";
   private static final String PARAMETER_H265_SPROP_MAX_DON_DIFF = "sprop-max-don-diff";
   private static final String PARAMETER_MP4V_CONFIG = "config";
+  private static final String PARAMETER_MP4A_CPRESENT = "cpresent";
```
claincly, line 58: `PARAMETER_MP4A_C_PRESENT`

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -211,11 +212,18 @@ public int hashCode() {
...
+          if (fmtpParameters.get(PARAMETER_MP4A_CPRESENT) != null && fmtpParameters.get(
+              PARAMETER_MP4A_CPRESENT).equals("0")) {
+            isCPresent = false;
+          }
+          @Nullable String configInput = fmtpParameters.get(PARAMETER_MP4V_CONFIG);
+          if (!isCPresent && configInput != null && configInput.length() % 2 == 0) {
+            Pair<Integer, Integer> configParameters = getSampleRateAndChannelCount(configInput);
+            channelCount = configParameters.first;
+            clockRate = configParameters.second;
+            formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);
```
claincly, line 225: would it make sense to do `formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);` in `processAacFmtpAttributes` since the method receives `channelCount` and `clockRate` already?

Have to make sure this change works for LATM (c present or not) and ADTS though

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -311,33 +319,30 @@ private static void processAacFmtpAttribute(
...
-          aacConfig = AacUtil.parseAudioSpecificConfig(scratchBits, false);
-        } catch (ParserException e) {
-          throw new IllegalArgumentException(e);
-        }
-        sampleRate = aacConfig.sampleRateHz;
-        channelCount = aacConfig.channelCount;
+  private static Pair<Integer, Integer> getSampleRateAndChannelCount(String configInput) {
+    int channelCount = 0, sampleRate = 0;
+    byte[] configBuffer = Util.getBytesFromHexString(configInput);
+    ParsableBitArray scratchBits = new ParsableBitArray(configBuffer);
```
claincly, line 331:
```
ParsableBitArray config = new ParsableBitArray(Util.getBytesFromHexString(configInput));
```

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -311,33 +319,30 @@ private static void processAacFmtpAttribute(
...
+    int audioMuxVersion = scratchBits.readBits(1);
+    if (audioMuxVersion == 0) {
+      checkArgument(scratchBits.readBits(1) == 1, "Invalid allStreamsSameTimeFraming.");
+      scratchBits.readBits(6);
+      checkArgument(scratchBits.readBits(4) == 0, "Invalid numProgram.");
```
claincly, line 336: change error message to "Only supports one program", ditto for layer, and `allStreamsSameTimeFraming`.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -311,33 +319,30 @@ private static void processAacFmtpAttribute(
...
-        AacUtil.Config aacConfig = null;
-        try {
-          aacConfig = AacUtil.parseAudioSpecificConfig(scratchBits, false);
-        } catch (ParserException e) {
-          throw new IllegalArgumentException(e);
-        }
-        sampleRate = aacConfig.sampleRateHz;
-        channelCount = aacConfig.channelCount;
+  private static Pair<Integer, Integer> getSampleRateAndChannelCount(String configInput) {
+    int channelCount = 0, sampleRate = 0;
```
claincly, line 329: I don't think you need to create these local vars right?

You can do

```java
if (audioMuxVersion == 0) {
  ...
  return Pair.create(aacConfig.sampleRateHz, aacConfig.channelCount);
}
throw new IAE("Only support audio mux version 0");

```

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -211,11 +212,18 @@ public int hashCode() {
...
-          channelCount = cfgOpts.first;
-          clockRate = cfgOpts.second;
-          formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);
+          boolean isCPresent = true;
+          if (fmtpParameters.get(PARAMETER_MP4A_CPRESENT) != null && fmtpParameters.get(
+              PARAMETER_MP4A_CPRESENT).equals("0")) {
+            isCPresent = false;
+          }
+          @Nullable String configInput = fmtpParameters.get(PARAMETER_MP4V_CONFIG);
+          if (!isCPresent && configInput != null && configInput.length() % 2 == 0) {
```
claincly, line 221: `if(!isCPresent)` only runs when C is **not** present, is that intentional? And `configInput != null && configInput.length() % 2 == 0` suggest C **does** exist

claincly: I would consider writing a test on this.

claincly (changes requested):

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
rakeshnitb: By default "formatBuilder.setSampleRate(clockRate).setChannelCount(channelCount);" is called at line 206. I'm calling this here only to update older values with value i'm getting after parsing.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
rakeshnitb: cpresent: a boolean parameter indicates whether audio payload configuration data has been multiplexed into an RTP payload (see section 4.1).  A 0 indicates the configuration data has not been multiplexed into an RTP payload, a 1 indicates that it has.  The default if the parameter is omitted is 1.

As mentioned in RFC 3016 if cpresent is set then config data will be multiplexed with RTP payload, then mediaDescriptor will have config parameter's values. So, I'm skipping parsing of config in that case.
Let me know if you still want to parse config data irrespective of cpresent is set or not.

claincly: Am I right in understanding that, `RtpMp4aReader` does not currently support the in-band mode (that is, with `cpresent` set to `1`), i.e. `RtpMp4aReader` does not parse the `audioMuxConfig` in RTP packets. 

RFC3016 also mentions in [section 4.1](https://datatracker.ietf.org/doc/html/rfc3016#section-4.1): 

> `muxConfigPresent`: ... In case of SDP, MIME parameter "config" is utilized (see
   [section 5.3](https://datatracker.ietf.org/doc/html/rfc3016#section-5.3)).

Since we are always using SDP, I think we should assert `cpresent == 0` to guarantee that `config` exists in SDP.

Please also add code comment explain that `cpresent == 0` means we need to parse `config`, or else this code segment will be very confusing to future readers.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
claincly: Why doing it in `onReceivingFirstPacket` but not in the constructor? `payloadFormat` is final and it seems `numberOfSubframes` can be final too

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
```diff
@@ -0,0 +1,182 @@
...
+        int payloadMuxLength = data.readUnsignedByte();
+        sampleLength += payloadMuxLength;
+        if (payloadMuxLength != 0xff) {
+          break;
+        }
+      }
+
+      // Write the audio sample
+      trackOutput.sampleData(data, sampleLength);
+      audioPayloadOffset+= sampleLength;
```
claincly, line 116: It seems `audioPayloadOffset` is redundant now

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReader.java
```diff
@@ -78,9 +78,9 @@ public void onReceivingFirstPacket(long timestamp, int sequenceNumber) {
     checkState(firstReceivedTimestamp == C.TIME_UNSET);
     firstReceivedTimestamp = timestamp;
     try {
-      numSubFrames = getNumOfSubframesFromMpeg4AudioConfig(payloadFormat.fmtpParameters);
+      numberOfSubframes = getNumOfSubframesFromMpeg4AudioConfig(payloadFormat.fmtpParameters);
```
claincly, line 81: can this be in the constructor and make `numberOfSubFrames` final?

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aPayloadReader.java
rakeshnitb: I have moved getNumOfSubframesFromMpeg4AudioConfig() function call inside constructor.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
```diff
@@ -208,6 +211,21 @@ public int hashCode() {
...
         checkArgument(!fmtpParameters.isEmpty());
+        if(mediaEncoding.equals(RtpPayloadFormat.RTP_MEDIA_MPEG4_AUDIO)) {
+          boolean isConfigPresent  = true;
+          if (fmtpParameters.get(PARAMETER_MP4A_C_PRESENT) != null && fmtpParameters.get(
+              PARAMETER_MP4A_C_PRESENT).equals("0")) {
+            isConfigPresent  = false;
+          }
+          checkArgument(!isConfigPresent, "cpresent == 0 means we need to parse config");
+          @Nullable String configInput = fmtpParameters.get(PARAMETER_MP4V_CONFIG);
+          if (configInput != null && configInput.length() % 2 == 0) {
```
claincly, line 222: since we require `cpresent=0`, then fmtp must include `config` right? can we assert on `configInput != null && configInput.length() % 2 == 0`? 

I'm not sure if this requirement applies to all ADTS/LATM streams.

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReaderTest.java
```diff
@@ -0,0 +1,175 @@
...
+          new Format.Builder()
+              .setSampleMimeType(MimeTypes.AUDIO_AAC)
+              .setWidth(352)
+              .setHeight(288)
+              .build(),
+          /* rtpPayloadType= */ 96,
+          /* clockRate= */ 90_000,
+          /* fmtpParameters= */ ImmutableMap.of(), RtpPayloadFormat.RTP_MEDIA_MPEG4_AUDIO);
+
+  @Rule public final MockitoRule mockito = MockitoJUnit.rule();
```
claincly, line 99: why do you use mockito in this file?

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReaderTest.java
```diff
@@ -0,0 +1,175 @@
...
+import org.mockito.junit.MockitoRule;
+
+/** Unit test for {@link RtpMp4aReader}. */
+@RunWith(AndroidJUnit4.class)
+public final class RtpMp4aReaderTest {
+  private static final byte[] FRAME_1_FRAGMENT_1_DATA =
+      getBytesFromHexString("0102");
+  private static final RtpPacket FRAME_1_FRAGMENT_1 =
+      new RtpPacket.Builder()
+          .setTimestamp((int) 2599168056L)
```
claincly, line 46: why casting a long to a method that requires long?

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReaderTest.java
```diff
@@ -0,0 +1,175 @@
...
+              /*payload size */ getBytesFromHexString("02"), FRAME_2_FRAGMENT_2_DATA))
+          .build();
+  private static final byte[] FRAME_2_DATA =
+      Bytes.concat(FRAME_2_FRAGMENT_1_DATA, FRAME_2_FRAGMENT_2_DATA);
+
+  private static final RtpPayloadFormat MP4ALATM_FORMAT =
+      new RtpPayloadFormat(
+          new Format.Builder()
+              .setSampleMimeType(MimeTypes.AUDIO_AAC)
+              .setWidth(352)
```
claincly, line 92: why aac audio has width and height

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReaderTest.java
```diff
@@ -0,0 +1,175 @@
...
+              .setHeight(288)
+              .build(),
+          /* rtpPayloadType= */ 96,
+          /* clockRate= */ 90_000,
```
claincly, line 96: `90_000` for audio sample?

claincly (changes requested):
Can you be more thorough with the code??

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReaderTest.java
claincly: Use 44.1k or similar. Remember to update the timestamp checks also

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpMp4aReaderTest.java
rakeshnitb: Sorry, it was mistakenly taken.

umangsaini:
This CL has merged. The Pull request may be marked as completed.

PR #254, opened 2023-02-22
Resolve discontinuities when seeking into fragmented AC-4 audio

hcgil:
Related to google/ExoPlayer#11000

Proof of concept for resolving discontinuities in FMP4 AC-4 playback.

The underlying cause on our end appears to come from the fact that we can't anticipate sync frames when committing them to our sample queue. Once we reach the end of our first fragment, we begin committing samples to our sample queue, even if suitable sync frames are available in later fragments.

For this PR, we would have some sort of intermediate structure to retain samples before committing them, in the event that our format does not guarantee that every frame is a key frame. When we encounter a key frame, we empty the contents of this structure, as we can begin playback from that key frame instead. When we pass our start time, we commit any samples in the intermediate structure to our sample queue, allowing us to play from the nearest key frame.


icbaker:
Please can you sign the CLA, then we can take a look at this.

PR #275, opened 2023-03-22
Dts mpeg2ts update

rahulnmohan:
Added below mentioned features.

- Support for extracting DTS LBR(DTS Express) and DTS UHD Profile 2(DTS:X) descriptor ID from PSI PMT
- The DTSReader class is updated for extracting a DTS LBR frame.
- Newly added DTSUhdReader class for extracting DTS UHD frame.
- The DTSUtil class is updated to parse the DTS LBR or DTS UHD frame and report the format information.

Similar changes are done for the ExoPlayer.
Pull request for the ExoPlayer updates: https://github.com/google/ExoPlayer/pull/11076
Feature request for the ExoPlayer: https://github.com/google/ExoPlayer/issues/11075


rohitjoins:
@rahulnmohan Do you think we can add some tests for this change?

rahulnmohan:
> @rahulnmohan Do you think we can add some tests for this change?

@rohitjoins I will check how to add test cases for the proposed changes.

rahulnmohan:
@rohitjoins 
Is there any documentation available for adding new tests to the framework?

I have added test entries for _sampleWithDts*_ in _TsExtractorTest_ class and copied sample files into _libraries\test_data\src\test\assets\media\ts_ directory. 
While running the test, seeing the below errors.

`Dump file not found. To update the dump file, change DumpFileAsserts#DUMP_FILE_ACTION to WRITE_TO_LOCAL (for Robolectric tests) or WRITE_TO_DEVICE (for instrumentation tests) and re-run the test.
java.io.IOException: Dump file not found. To update the dump file, change DumpFileAsserts#DUMP_FILE_ACTION to WRITE_TO_LOCAL (for Robolectric tests) or WRITE_TO_DEVICE (for instrumentation tests) and re-run the test.
	at androidx.media3.test.utils.DumpFileAsserts.assertOutput(DumpFileAsserts.java:163)
	at androidx.media3.test.utils.DumpFileAsserts.assertOutput(DumpFileAsserts.java:95)
	at androidx.media3.test.utils.ExtractorAsserts.assertOutput(ExtractorAsserts.java:454)
	at androidx.media3.test.utils.ExtractorAsserts.assertBehavior(ExtractorAsserts.java:398)
	at androidx.media3.test.utils.ExtractorAsserts.assertBehavior(ExtractorAsserts.java:354)
	at androidx.media3.extractor.ts.TsExtractorTest.sampleWithDts(TsExtractorTest.java:119)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:591)
	at org.robolectric.internal.SandboxTestRunner$2.lambda$evaluate$0(SandboxTestRunner.java:274)
	at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:88)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.io.FileNotFoundException: extractordumps/ts/sample_dts.ts.0.dump
	at org.robolectric.shadows.ShadowArscAssetManager10.nativeOpenAsset(ShadowArscAssetManager10.java:747)
	at android.content.res.AssetManager.nativeOpenAsset(AssetManager.java)
	at android.content.res.AssetManager.open(AssetManager.java:874)
	at android.content.res.AssetManager.open(AssetManager.java:851)
	at androidx.media3.test.utils.TestUtil.getInputStream(TestUtil.java:171)
	at androidx.media3.test.utils.TestUtil.getByteArray(TestUtil.java:166)
	at androidx.media3.test.utils.TestUtil.getString(TestUtil.java:176)
	at androidx.media3.test.utils.DumpFileAsserts.assertOutput(DumpFileAsserts.java:161)
	at androidx.media3.test.utils.DumpFileAsserts.assertOutput(DumpFileAsserts.java:95)
	at androidx.media3.test.utils.ExtractorAsserts.assertOutput(ExtractorAsserts.java:454)
	at androidx.media3.test.utils.ExtractorAsserts.assertBehavior(ExtractorAsserts.java:398)
	at androidx.media3.test.utils.ExtractorAsserts.assertBehavior(ExtractorAsserts.java:354)
	at androidx.media3.extractor.ts.TsExtractorTest.sampleWithDts(TsExtractorTest.java:119)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)`

rohitjoins:
@rahulnmohan The error is complaining about not finding the dump file for the sample test file you added. As mentioned in the error log, you'll need to temporarily modify [this](https://github.com/androidx/media/blob/850bd69dddca45915199f221d48573fca3fc790b/libraries/test_utils/src/main/java/androidx/media3/test/utils/DumpFileAsserts.java#L75) line in DumpFileAsserts from `COMPARE_WITH_EXISTING` to `WRITE_TO_LOCAL` and run the tests which will add/create the dump files with data for this new file. You can revert back this change once the new dump file is created. Dump file is data against which the test verifies the sample file when tests are run.

rahulnmohan:
@rohitjoins Thanks for the prompt reply. I have added test cases for the TS extractor updates. Would you please review it?

rohitjoins:
@rahulnmohan Thanks for adding the tests and sample files. One more request would be to minimise the size of the sample files(possibly down to single number digit of samples) and still maintain test coverage for them. We try to keep the test media duration reasonably short, so tests run faster and to avoid having a huge directory of binaries that everyone cloning the project externally has to check out.

rahulnmohan:
@rohitjoins Thanks for reviewing the changes.
Currently, I have submitted the same test streams which we added in the Android CTS. I will try for getting shorter streams. Thanks.

rohitjoins:
@rahulnmohan One alternative suggestion could be to consider safely truncating the streams that were included in the tests.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -199,8 +334,262 @@ public static int getDtsFrameSize(byte[] data) {
...
+    int staticFieldsPresent = frameBits.readBits(1);
+    if (staticFieldsPresent == 1) {
+      nRefClockCode = frameBits.readBits(2);
+      nExSSFrameDurationCode = 512 * (frameBits.readBits(3) + 1);
+      if (frameBits.readBits(1) == 1) {
+        frameBits.skipBits(32 + 4);
+      }
+
+      int[] nuActiveExSSMask = new int[8];
+      frameBits.skipBits(3 + 3); // numAudioPresent, numAssets
```
rohitjoins, line 372: Why do we skip the bits for `numAudioPresent` and `numAssests` here? Based on the spec it seems that both `numAudioPresent` and `numAssests` should be 1 only if static fields are not present but in the current code it is always 1 irrespective of whether static fields are present or not.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -199,8 +334,262 @@ public static int getDtsFrameSize(byte[] data) {
...
+        case 0:
+          clockRateHertz = 32000;
+          break;
+        case 1:
+          clockRateHertz = 44100;
+          break;
+        case 2:
+          clockRateHertz = 48000;
+          break;
+      }
```
rohitjoins, line 497: Can we use REF_CLOCK_TABLE here?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: Streaming encoders won't encode streams with multiple audio presentations and assets.
In order to make the parsing logic less complex, I have hardcoded those values as one.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: The 'REF_CLOCK_TABLE' is defined in the 'DTS Core and Extension' specification document (ETSI TS 102 114 V1.6.1 (2019-08), table 7.3). We are following this spec for the Digital Surround and the DTS-HD(LBR) frame header parsing.

For implementing UHD frame header parsing, we have to refer to the spec - ETSI TS 103 491 V1.2.1 (2019-05), Table 6-12.
If you prefer to use a table to access the sampling rate information, I could add another table for this.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -199,8 +334,262 @@ public static int getDtsFrameSize(byte[] data) {
...
+              /* cause= */ null);
+        }
+      }
+      chunkPayloadBytes += nuAudioChunkSize;
+    }
+
+    int frameSize = nFTOCPayloadinBytes + chunkPayloadBytes;
+    return new DtsFormatInfo(
+        MimeTypes.AUDIO_DTS_X,
+        DTS_UHD_CHANNEL_COUNT, // TODO: Need to parse the actual channel count from bitstream
```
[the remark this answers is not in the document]
rahulnmohan: Currently, we have implemented the parsing of the FTOC element in the UHD frame. To get the channel count information, we need to parse the metadata chuck information also (Please see section 4.2.1 in ETSI TS 103 491 V1.2.1 (2019-05) to get a high-level idea of the UHD stream structure).

The process sequence of the UHD reader is given below,
Find sync -> Read 7 bytes data -> Find FTOC header size -> Read FTOC header bytes -> Parse FTOC header and find format information, and frame size -> Read full frame

To find the channel count, we have to parse deep into the bitstream which contains the metadata chunk bytes. So the parsing has to perform after the 'Read full frame'.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
```diff
@@ -0,0 +1,242 @@
...
+            .build();
+        output.format(format);
+      }
+      sampleCount = formatInfo.sampleCount; // Update the sample count only in FTOC Sync frame
+    }
+    sampleSize = formatInfo.frameSize;
+
+    // In this class a sample is an access unit (frame in DTS), but the format's sample rate
+    // specifies the number of PCM audio samples per second.
+    sampleDurationUs = (int)(C.MICROS_PER_SECOND * sampleCount / format.sampleRate);
```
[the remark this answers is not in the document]
rahulnmohan: Thanks for pointing out this potential issue.

The sample rate information in the 'formatInfo' is valid only in the FTOC sync frames. For FTOC non-sync frames, we have to use the stream information reported in the last FTOC sync frame.

I will add some handling in the code to avoid any possible NullPointerException.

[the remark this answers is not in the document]
rahulnmohan: @rohitjoins I have added a fix. Could you review it?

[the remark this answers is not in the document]
rohitjoins: Thanks :)

rahulnmohan:
Hi @rohitjoins, please let me know if any further action is required from my side to merge these changes to the main branch. Thanks.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
```diff
@@ -0,0 +1,246 @@
...
+import androidx.media3.extractor.DtsUtil;
+import androidx.media3.extractor.ExtractorOutput;
+import androidx.media3.extractor.TrackOutput;
+import androidx.media3.extractor.ts.TsPayloadReader.TrackIdGenerator;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+import org.checkerframework.checker.nullness.qual.RequiresNonNull;
+
+/** Parses a continuous DTS UHD byte stream and extracts individual samples. */
+@UnstableApi
+public final class DtsUhdReader implements ElementaryStreamReader {
```
rohitjoins, line 37: Is there a specific reason to declare a new class for UHD? Can we not accommodate  it in `DtsReader`? As far as I can tell from the implementation we can add variables specific to UHD into DtsReader as we have done for HD.

rahulnmohan: @rohitjoins, Please let me know if any further code updates are required on this CL. Thank you.

FongMi: Is there any update?

rahulnmohan: @rohitjoins , would you please suggest further action required to merge these changes?

rahulnmohan:
Hi @rohitjoins ,
a gentle reminder of this pull request.
Would it be possible to share a tentative timeline to have these changes merged into the main branch?
Thank you.

rohitjoins:
@rahulnmohan 

Please refrain from pushing any more changes as it will complicate the internal review. I would try my best to include this change into our next release i.e. 1.2.0. Thanks.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -223,8 +357,262 @@ public static int getDtsFrameSize(byte[] data) {
...
+            /* message= */ "Unsupported metadata chunk size in DTS UHD header: " + nuMDChunkSize,
+            /* cause= */ null);
+      }
+      chunkPayloadBytes += nuMDChunkSize;
+    }
+
+    // Ony one audio chunk is supported
+    int numAudioChunks = 1;
+    for (int nac = 0; nac < numAudioChunks; nac++) {
+      int acID = 256, nuAudioChunkSize;
```
rohitjoins, line 545: where does this 256 come from?

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  /** Find the actual size of DTS UHD frame header. */
+  private void findHeaderSize() {
+    byte[] frameData = new byte[FTOC_HEADER_SIZE_MIN];
+    headerScratchBytes.setPosition(0);
+    headerScratchBytes.readBytes(frameData, 0, FTOC_HEADER_SIZE_MIN);
+    headerSizeToRead = DtsUtil.parseDtsUhdHeaderSize(frameData);
+    if (headerSizeToRead < bytesRead) { // Already read more data than the actual header size
+      headerSizeToRead = bytesRead; // Setting target read length equal to bytesRead
```
rohitjoins, line 207: Why is this needed? The test cases fail when this is removed for `DtsUhdReader` but still passes for `DtsReader`.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -223,8 +357,262 @@ public static int getDtsFrameSize(byte[] data) {
...
+        frameBits.skipBits(32 + 4);
+      }
+      int samplRateMultiplier = (1 << frameBits.readBits(2));
+      sampleRate = clockRateHertz * samplRateMultiplier;
+    }
+
+    int chunkPayloadBytes = 0;
+    int numOfMDChunks = bSyncFrameFlag ? 1 : 0; // Metadata chunks
+    for (int nmdc = 0; nmdc < numOfMDChunks; nmdc++) {
+      int[] ucTable2 = new int[] { 6, 9, 12, 15 };
```
rohitjoins, line 532: AFAICT from the naming conventions in the spec, this is kinda like declaring int integer? Since uc just means unsigned char. Can you please define a better name for these tables?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -223,8 +357,262 @@ public static int getDtsFrameSize(byte[] data) {
...
+    if (bSyncFrameFlag) {
+      // fullChannelBasedMixFlag, ETSI TS 103 491 V1.2.1, Section 6.4.6.1
+      if(frameBits.readBits(1) != 1) {
+        throw ParserException.createForMalformedContainer(
+            /* message= */ "Only supports full channel mask-based audio presentation",
+            /* cause= */ null);
+      }
+
+      int nBaseDurationIndex = frameBits.readBits(2);
+      sampleCount = BASE_DURATION[nBaseDurationIndex] * (frameBits.readBits(3) + 1);
```
rohitjoins, line 507: `m_unFrameDuration` is defined as "The frame duration in clock periods"

Making this equal to sampleCount implies there's exactly one sample per clock period. Is this assumption correct?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -223,8 +357,262 @@ public static int getDtsFrameSize(byte[] data) {
...
+    int nFTOCPayloadinBytes = extractVarLenBitFields(frameBits, ucTable1, true) + 1;
+
+    if (bSyncFrameFlag) {
+      // fullChannelBasedMixFlag, ETSI TS 103 491 V1.2.1, Section 6.4.6.1
+      if(frameBits.readBits(1) != 1) {
+        throw ParserException.createForMalformedContainer(
+            /* message= */ "Only supports full channel mask-based audio presentation",
+            /* cause= */ null);
+      }
+
```
rohitjoins, line 505: should there be a CRC check here? 6.4.6.2

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -223,8 +357,262 @@ public static int getDtsFrameSize(byte[] data) {
...
+        sampleCount,
+        0);
+  }
+
+  /**
+   * Returns the size of frame header in a DTS-HD frame by parsing a few bytes of data
+   * (minimum 8 bytes) from the input bitstream(containing only Extension Sub-stream).
+   *
+   * @param frame The initial 8 bytes(minimum) of a DTS-HD frame(having only Extension Sub-stream).
+   * @return Size of the DTS-HD frame header in bytes.
```
rohitjoins, line 460: Does it mean it's always ok to pass an 8-byte array? It looks like it will read at most 55 bits, so I guess 8 bytes is safe and that is what it means.

But this documentation makes it sound like it will read at least 64 bits, and could read up to an infinite number:

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -223,8 +357,262 @@ public static int getDtsFrameSize(byte[] data) {
...
+        }
+        frameBits.skipBits(5); // nuBitResolution
+
+        sampleRate = SAMPLE_RATE_EXTSS[frameBits.readBits(4)];
+        channelCount = frameBits.readBits(8) + 1;
+      }
+    }
+
+    // Number of audio sample in a compressed DTS frame
+    int sampleCount = nExSSFrameDurationCode * (sampleRate / REF_CLOCK_TABLE[ nRefClockCode]);
```
rohitjoins, line 445: I see the following equation in the spec (page 98):

ExSSFrameDuration = nuExSSFrameDurationCode × RefClockPeriod

Is this line of code we are also multiplying ExSSFrameDuration by the sampleRate? Why is that?

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
rohitjoins: I think it would be better to have a single DTS reader class to handle all types of DTS stream, we won't have to make the same changes in two different places as most of the logic is repeated.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
```diff
@@ -0,0 +1,246 @@
...
+  /**
+   * Constructs a new reader for DTS UHD elementary streams.
+   *
+   * @param language Track language.
+   */
+  public DtsUhdReader(@Nullable String language) {
+    headerScratchBytes = new ParsableByteArray(new byte[FTOC_HEADER_SIZE_MAX]);
+    state = STATE_FINDING_SYNC;
+    timeUs = C.TIME_UNSET;
+    sampleRate = 48000; // initialize to a non-zero sampling rate
```
rohitjoins, line 84: Why do we choose this value 48000 as the default?

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
```diff
@@ -0,0 +1,246 @@
...
+            state = STATE_READING_SAMPLE;
+          }
+          break;
+        case STATE_READING_SAMPLE:
+          int bytesToRead = min(data.bytesLeft(), sampleSize - bytesRead);
+          output.sampleData(data, bytesToRead);
+          bytesRead += bytesToRead;
+          if (bytesRead == sampleSize) {
+            if (timeUs != C.TIME_UNSET) {
+              output.sampleMetadata(timeUs, C.BUFFER_FLAG_KEY_FRAME, sampleSize, 0, null);
```
rohitjoins, line 141: Is this correct? what if it's a frame with syncWord == SYNC_VALUE_UHD_FTOC_NONSYNC_BE? Based on 103 491 section 6.4.4.2 we should only mark frames with syncWord == SYNC_VALUE_UHD_FTOC_SYNC_BE as keyframes.

rohitjoins (commented):
Sorry for replying on this after so long.

rohitjoins:
@rahulnmohan,

Thank you again for bearing with me, I have uploaded the changes I had made based on the internal review and after formatting it with `google-java-format`.

You can see the modifications I have already made(please let us know if you find anything alarming) and make any further changes that would arise from any of the comments above on top of this.

Once all the comments are resolved, I'll send it for another round of internal review before merging the PR.

rahulnmohan:
> @rahulnmohan,
> 
> Thank you again for bearing with me, I have uploaded the changes I had made based on the internal review and after formatting it with `google-java-format`.
> 
> You can see the modifications I have already made(please let us know if you find anything alarming) and make any further changes that would arise from any of the comments above on top of this.
> 
> Once all the comments are resolved, I'll send it for another round of internal review before merging the PR.

@rohitjoins 
Thank you for reviewing it further. 
I just returned from my vacation. I will check your review comments ASAP. Thank you.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: 256 is used as a number which indicates the invalid audio chunck ID. The valid range is up to 255 only.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
rahulnmohan: For the DTS UHD stream, the header size is not constant. To find the actual header size, first, we need to parse a few bytes of data(FTOC_MIN_HEADER_SIZE). Sometimes, the actual header size may be less than the FTOC_MIN_HEADER_SIZE.
i.e.   bytesRead > headerSizeToRead
In this case, if we pass the actual 'headerSizeToRead' to the _continueRead()_ function invoked in the 'STATE_READING_HEADER' state, the _readBytes()_ will fail(the number of bytes to read will become negative).

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: Yes, I will update this.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: The CRC check is not valid for fully channel-based mix(m_bFullChannelBasedMixFlag = TRUE).
Currently, we encode such streams only.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: Yes I agree, I will update the documentation.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: The ExSSFrameDuration is the actual duration in seconds. To convert this value to sampleCount(number of samples), we need to multiply it by sampleRate.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
rahulnmohan: The sampleRate information will be updated only in the FTOC sync frame. For any subsequent FTOC non-sync frames, the sampleRate will be the same as the previous sync frame.
The sampleRate information is used in the below calculation.
 `sampleDurationUs = (int)(C.MICROS_PER_SECOND * sampleCount / sampleRate);`

By any chance, if the extractor encounters a non-sync frame at the very beginning, the sampleRate won't get extracted from the stream. To avoid 'division by zero' in the sampleDurationUs calculation, I have initialized the sampleRate value to a non-zero value. The UHD streams are encoded at 48000 sample rate. I used the same value for initializing sampleRate variable.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
rahulnmohan: The FTOC sync and non-sync frames need to pass to the DTS decoder. 
Which flag do we need to use with the non-sync frame if the BUFFER_FLAG_KEY_FRAME is not suitable for such frames?

rahulnmohan:
Hi @rohitjoins,
I have modified the code to address your review comments. Implemented a single DTS reader class to handle all the DTS stream types(Digital Surround, LBR, and UHD/DTS:X). Could you check the latest code updates and let me know your feedback?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
icbaker: Hey @rahulnmohan, I'm reviewing this PR internally - I thought it would be quicker to come back you with some unresolved questions.

I'm not sure I understand how what you've said here reconciles with the spec that says:
> When `m_bFullChannelBasedMixFlag = TRUE`, the FTOC CRC16 word is transmitted in the stream only in the sync frames. Consequently for this case the CRC check is performed only in sync frames.

It seems that here we **are** in a sync frame (we're inside an `if (syncFrameFlag)` stanza), so according to the spec we should be doing a CRC check here. Can you help me understand why you're saying it would be invalid to do so?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
icbaker: Why is it correct to use an invalid audio chunk ID here if? The spec (ETSI TS 103 491 V1.2.1 table 6-20) makes it look like we should look up the audio chunk ID of the previous sync frame? I think that's what this line is implying:

```
 // Get the audio chunk ID from the stored list
 unAudioChunkID =
m_CAudioChunkNaviLits.m_pCAudioChunksNaviParams[unListIndex].m_unAudioChunkID; 
```

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: Hi @icbaker,
Thank you very much for reviewing the code.
I have added the CRC check for the FTOC sync frames. Could you check the code updates?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: @icbaker, as you mentioned, in non-syn frames we need to use the chunk ID reported in the previous sync frame.
But I am trying to avoid keeping a state inside the utility function.
In any sync frame, the _unAudioChunkID_ extracted from the stream will be a non-zero value(as per the current bitstream stream construction, it will be 0x01). So we can consider that the subsequent non-sync frame will get the stored non-zero value for the _unAudioChunkID_.

icbaker: > But I am trying to avoid keeping a state inside the utility function.

I understand this is a nice goal, but doesn't the spec **require** keeping some form of state?

I'm afraid I'm unclear how the current implementation is compatible with the spec - since the spec says on non-sync frames we need to use the audio chunk ID of the previous sync frame, and the implementation says that non-sync frames have an invalid audio chunk ID.

rahulnmohan: @icbaker , I made an update in the code to handle this. Could you check my last commit([e99a618](https://github.com/androidx/media/pull/275/commits/e99a6185d3f22e376243e1e29524766d7f700c43))?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -47,55 +109,144 @@ public final class DtsUtil {
...
+  /**
+   * Maps index to number of clock cycles in the bfase duration period. See ETSI TS 103 491 V1.2.1
+   * (2019-05) Table 6-13.
+   */
+  private static final int[] BASE_DURATION_BY_INDEX = new int[] {512, 480, 384};
+
+  /**
+   * A state variable that stores the UHD audio chunk ID extracted from the FTOC sync frame.
+   * This value will be used in the subsequent FTOC non-sync frame */
+  private static int storedUhdAudioChunkId;
```
icbaker, line 196: I don't think this can be static.

What if we have two ExoPlayer instances in the same app, both playing UHD content? They will end up cross-talking in this field.

rahulnmohan: @icbaker Would you think that the two instance executions will be affected by the other static variables in the Utils class(unless the instance switches the execution in the frame boundary)?
What would be a use case for having two instances playing UHD content?

In my latest commit([39cd88d](https://github.com/androidx/media/pull/275/commits/39cd88d2d8d7360c42d4927ee6b3f32e96b9b91b), I have updated the logic to allocate the state variable in the DtsReader class instead of the DtsUtil class. I hope that this will ensure an independent state for each of the ExoPlayer instances.

icbaker: > Would you think that the two instance executions will be affected by the other static variables in the Utils class

Which static variables in `DtsUtil` are you referring to? All the ones I can see are either immutable primitives, or arrays that aren't technically immutable but aren't meant to be mutated (and if they're being mutated, that's probably a bug we should fix).

> What would be a use case for having two instances playing UHD content?

One example might be cross-fading between two tracks, which means playing both at the same time for a bit. Or otherwise mixing content together. Lots of reasons we shouldn't assume there's only ever one ExoPlayer instance per app and therefore generally avoid static state.

----

The proposed approach of keeping track of the state in the reader looks better - thanks.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
rahulnmohan: > I think it would be better to have a single DTS reader class to handle all types of DTS stream, we won't have to make the same changes in two different places as most of the logic is repeated.

This is done in the commit [ef41526](https://github.com/androidx/media/pull/275/commits/ef415268e215719ce385f5fb0e3149679a3bd91e)

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: This is done in the commit [ef41526](https://github.com/androidx/media/pull/275/commits/ef415268e215719ce385f5fb0e3149679a3bd91e)

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: This is done in the commit [ef41526](https://github.com/androidx/media/pull/275/commits/ef415268e215719ce385f5fb0e3149679a3bd91e)

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: This is done in the commit [ef41526](https://github.com/androidx/media/pull/275/commits/ef415268e215719ce385f5fb0e3149679a3bd91e)

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: Thank you for explaining the details.
I hope that all the review comments are addressed and no further modification is required on this CL.

rahulnmohan: Hi @icbaker , @rohitjoins 
Would you think this CL is ready to merge or need any updates?
Thank you.

FongMi: We expect to see this in 1.2.0.

rohitjoins: Hi @rahulnmohan,

This is still going through the internal review process. We don't expect this to be released as part of 1.2.0 as the changes are big enough, it cannot be included after 1.2.0-alpha01 release which was done more than 3 months ago.

cedricxperi: @rohitjoins @FongMi We have several customers in China who are using DTS audio with HLS-TS and Exoplayer. This contribution will go a long way in helping us support these efforts. Your kind assistance to expedite this is very much appreciated. Thanks.

FongMi: @cedricxperi Can you help me test this link?
http://150.158.170.214/ishang.php?id=cctv1hd
need china vpn, I recommand 穿梭, and use 全局模式
EXO don't decode the video track, only audio track
IJK can play it.

cedricxperi: @FongMi I tested with Transocks VPN. It can't play video or audio. I am testing from Singapore. May I know what is the content?

FongMi: Sorry, it not about this pull, I have mail to you.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rohitjoins: Because maxSampleRate is:

1. a max value (i.e. not always the same value for every frame and/or every channel)
2. per-channel (p106: "the maximum sample frequency among all channels present in the audio asset.")

Whereas I assume (?) a frame might contain data for multiple(/all?) channels.

So concretely, given:

nuMaxSampleRate = 48_000
nuTotalNumChs = 10
ExSSFrameDuration = 0.05
I would expect (?) a frame could contain 48_000 * 10 * 0.05 = 24_000 samples

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
```diff
@@ -66,9 +97,13 @@ public final class DtsReader implements ElementaryStreamReader {
    * @param language Track language.
    */
   public DtsReader(@Nullable String language) {
-    headerScratchBytes = new ParsableByteArray(new byte[HEADER_SIZE]);
+    int headerSize = max(EXTSS_HEADER_SIZE_MAX, FTOC_MAX_HEADER_SIZE);
+    headerScratchBytes = new ParsableByteArray(new byte[headerSize]);
     state = STATE_FINDING_SYNC;
     timeUs = C.TIME_UNSET;
+    uhdAudioChunkId = new AtomicInteger();
```
rohitjoins, line 104: Shouldn't this `uhdAudioChunkId` also `reset` in the `seek` method below? I tried resetting it but then the tests fails with an error. 

When seeking, the last key frame wouldn't be the same and as this parameter stores the audio chunk ID of the last key frame, I would expect this to be cleared. What do you think?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -223,36 +352,365 @@ public static int getDtsFrameSize(byte[] data) {
...
+    int assetsCount; // nuNumAssets
+    int referenceClockCode; // nuRefClockCode
+    int extensionSubstreamFrameDurationCode; // nuExSSFrameDurationCode
+
+    boolean staticFieldsPresent = frameBits.readBit(); // bStaticFieldsPresent
+    if (staticFieldsPresent) {
+      referenceClockCode = frameBits.readBits(2);
+      extensionSubstreamFrameDurationCode = 512 * (frameBits.readBits(3) + 1);
+
+      if (frameBits.readBit()) { // bTimeStampFlag
```
rohitjoins, line 389: this bit is hard to match against the spec imo

AFAICT the if (frameBits.readBit()) { line here corresponds to m_bParamPresent = (ExtractBits(1)==1) ? TRUE : FALSE; in Table 5-3.

Which then means we're assuming that m_bSyncFramePredefValueExists = false? Because if it's true I think we need to read another bit (m_bUpdateFlag = (m_bSyncFramePredefValueExists) ? (bool) ExtractBits(1) : TRUE;)

Do you know what m_bSyncFramePredefValueExists corresponds to in the code (and thus why it's correct to assume it's true)? Maybe worth a bit more detail in the comment to explain the reasoning here.

rohitjoins (commented):
Hi @rahulmohan,

Please see three new comments added to resolve the final few things which is unclear to us to merge this PR.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
rahulnmohan: @rohitjoins, It seems to be good to reset the _uhdChunkId_ parameter during the seek operation.

I have added this line of code in the seek() function.
`uhdAudioChunkId.set(0);`
I did not see any issues when I tested using a streaming use case. I did seeking the player multiple times from the UI.

Could you share the details of the code changes you added and the tests you did?

rohitjoins: @rahulmohan,

I did not make any code changes, just added `uhdAudioChunkId.set(0);` in the `seek()` method and ran the `TsExtractorTest` tests. 5 tests fails for `sampleWithDtsUhd` test.

rahulnmohan: @rohitjoins 
The issue was caused by the wrong UHD header size value and this resulted in a negative index calculation when the uhdAudioChunkId is zero. I have fixed this issue in [b8f0350](https://github.com/androidx/media/pull/275/commits/b8f0350ddbb50c6d74b758c99d40d9d674c61009)

It should be fine to use the uhdAudioChunkId value from the last sync frame. So, no need to reset the uhdAudioChunkId in the seek() function.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: I have updated the comment. Please refer [b8f0350](https://github.com/androidx/media/pull/275/commits/b8f0350ddbb50c6d74b758c99d40d9d674c61009)

rohitjoins: The comment still doesn't explain why `m_bSyncFramePredefValueExists` is false?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rohitjoins: @rahulnmohan can you please also look at this comment?

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
rahulnmohan: In Table 6-20 (ETSI TS 103 491 V1.2.1 (2019-05)), you can see that the m_unAudioChunkID is extracted from the navigation table. 

```
// Get the audio chunk ID from the stored list
unAudioChunkID =
m_CAudioChunkNaviLits.m_pCAudioChunksNaviParams[unListIndex].m_unAudioChunkID;
```

This list will be cleared in every sync frame and updated with new values.
```
if (m_bSyncFrameFlag) {
m_CAudioChunkNaviLits.ResetList();
}
```
So, we have to reset the uhdAudioChunkId in every sync frame.
Since we are extracting the unAudioChunkID in every sync frame and updating uhdAudioChunkId, we can avoid resetting uhdAudioChunkId explicitly.
 
Currently, the encoder supports only the Audio chunk type 0x1. So the chunk ID will be the same in all the frames.

As per the spec(section 4.2.4), the non-sync frame shouldn't be used as a random access point. So I am adding a logic to establish the synchronization only in the FTOC sync frames.
This will cause failure in the TsExtractorTest for seek test cases. The UHD test vector(sample_dts_uhd.ts) only has one FTOC sync frame. So after the seek operation, synchronization will not happen.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: The m_bSyncFramePredefValueExists is set to false in the encoder for the m_TimeStamp field.

rahulnmohan:
Hi @rohitjoins 
Please let me know if I need to further clarify anything in this thread.

rohitjoins:
Hi @rahulnmohan,

Yes, please look at the comment in L445 in `DtsUtil.Java`. I have also tagged you there to have your attention.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: @rohitjoins 
Sorry, I missed to answer this question.

A frame will contain data for all the channels. If the ExSSFrameDutaion is 0.05, the number of samples decoded per channel will be _Sampling Rate * 0.05_
If the number of channels is 10, then the total sample count will be  _Sampling Rate * 10 * 0.05_

rahulnmohan:
@rohitjoins, Please see my reply here https://github.com/androidx/media/pull/275#discussion_r1426589042

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
rohitjoins: Hi @rahulnmohan,

With your latest set of commits for this, the `TsExtractorTest` tests are failing for seek test cases, do you intend to fix this or should I do something about this? This bit is unclear and we cannot merge a PR with failing test cases right?

rohitjoins:
@rahulnmohan, can you please see the [comment](https://github.com/androidx/media/pull/275#discussion_r1432980203) and reply.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
rahulnmohan: Hi @rohitjoins, I will update the UHD test vector to have at least two sync frames.

rahulnmohan: @rohitjoins 
I have updated the UHD test vector. Now the test stream contains three sync frames.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rohitjoins: @rahulnmohan Agreed that we don't need to multiply this with the number of channels as `ExSSFrameDutaion` is for all channels in the frame. 

Although it is still unclear that the sampling rate we're deriving here is correct, as in the spec it says it is maximum sampling rate (i.e. not always the same value for every frame and/or every channel).

From the spec: `The values that are retrieved from the table are a max sample rate`, How do we justify that?

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
rohitjoins: Agreed that in normal cases we need to feed all (sync and non-sync) frames to the decoder. 

I don't think we can start decoding from a non-sync frame and as such non-sync frames should be marked as `BUFFER_FLAG_KEY_FRAME`. Please also check if changing this would result in test failures while seeking.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
rohitjoins: > Since we are extracting the unAudioChunkID in every sync frame and updating uhdAudioChunkId, we can avoid resetting uhdAudioChunkId explicitly.

Does setting it explicitly breaks any tests? If not, I think we should do it so that we don't use a stale `uhdAudioChunkId` by mistake.

rohitjoins (commented):
Also added three other comments:
[comment1](https://github.com/androidx/media/pull/275#discussion_r1435112875)
[comment2](https://github.com/androidx/media/pull/275#discussion_r1435116309)
[comment3](https://github.com/androidx/media/pull/275#discussion_r1435120432)

Please resolve these last few comments to make further progress in the merge process of this PR.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
rahulnmohan: @rohitjoins, the max sampling frequency shall be used as the sample rate of the stream. The Table 'Section G.3.1, G-1: Common attributes' aligns with this point.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsUhdReader.java
rahulnmohan: The decoding won't start from the non-sync frame. In the seek() function, we are setting 'uhdInSync' to false.
The 'skipToNextSync()' function will return 'true' only if "it is a FTOC sync frame" OR "FTOC non-sync frame and already parsed a FTOC sync".

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
rahulnmohan: Resetting the _uhdAudioChunkId_ won't break any tests. I will add the below code to the seek() function.
`uhdAudioChunkId.set(0)`

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
```diff
@@ -95,20 +133,66 @@ public void packetStarted(long pesTimeUs, @TsPayloadReader.Flags int flags) {
...
+            parseCoreHeader();
             headerScratchBytes.setPosition(0);
-            output.sampleData(headerScratchBytes, HEADER_SIZE);
+            output.sampleData(headerScratchBytes, CORE_HEADER_SIZE);
+            state = STATE_READING_SAMPLE;
+          }
+          break;
+        case STATE_FINDING_EXTSS_HEADER_SIZE:
+          // Read enough bytes to parse the header size information.
+          if (continueRead(data, headerScratchBytes.getData(), /* targetLength= */ 10)) {
```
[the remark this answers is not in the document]
rahulnmohan: @rohitjoins, To parse the ExtSS header size, 7 bytes is enough. The 'continueRead()' function handles the size of any extra bytes read before.
`int bytesToRead = min(source.bytesLeft(), targetLength - bytesRead);`

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
```diff
@@ -95,20 +133,66 @@ public void packetStarted(long pesTimeUs, @TsPayloadReader.Flags int flags) {
...
+            headerScratchBytes.setPosition(0);
+            output.sampleData(headerScratchBytes, extensionSubstreamHeaderSize);
+            state = STATE_READING_SAMPLE;
+          }
+          break;
+        case STATE_FINDING_UHD_HEADER_SIZE:
+          // Read enough bytes to parse the header size information.
+          if (continueRead(data, headerScratchBytes.getData(), /* targetLength= */ 6)) {
+            uhdHeaderSize = DtsUtil.parseDtsUhdHeaderSize(headerScratchBytes.getData());
+            // Adjust the array read position if data read is more than the actual header size.
```
rohitjoins, line 182: I believe this is no longer required as we always start with a SYNC frame. If you'll confirm, I can remove this.

rahulnmohan: The header size extraction needs to be performed both in the FTOC sync frame and FTOC non-sync frames.
 
Every UHD frame contains an FTOC(Frame Table of Contents) component. The main component of the FTOC is the Sync Word (the first 4 bytes of the FTOC), which indicates whether the frame is a sync frame or a non-sync frame. 

In the sync frame(FTOC sync frame), the decoder does not need any information from previous frames to produce output. All parameters necessary to unpack metadata and audio chunks can be found within the payload of this sync
frame.

A non-sync frame(FTOC non-sync frame) permits both metadata chunks and audio chunks to minimize payload size by only sending parameters that have changed in value since the previous frame or sync frame. All parameters that are not re-transmitted are assumed to maintain their previous value. Any value or set of values may be updated in a non-sync frame.

rohitjoins:
@rahulnmohan, the PR was merged as [27c021f](https://github.com/androidx/media/commit/27c021fd7bb827491568dca04b11a9a396d8aee9) to the main branch.

Will be released with 1.3.0.

rahulnmohan:
@rohitjoins, Thank you very much for the great help and the extended review.

PR #326, opened 2023-04-15
Update translations

NUTSUPRA:
Fix the wrong translation.


icbaker:
Thanks for the fix! Unfortunately I can't directly merge this, because it will get overwritten with the old version by our translation machinery next time it runs - I'll work out how to make the change further upstream so that the fix persists.

icbaker:
The fix has been made in https://github.com/androidx/media/commit/bba760f6e5ae8ba6db2cd2e45163698223335c2a, this will be included in the next release.

PR #335, opened 2023-04-18
Dts direct passthrough support

cedricxperi:
This adds DTS Audio Support for Direct Passthrough playback. This is required for Tunnel mode playback on Android TVs that supports tunnel mode playback (e.g. MTK and Realtek based TVs). Test streams will be provided during the review if required.
Thanks.

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
+++ b/libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
@@ -595,6 +595,10 @@ public final class MimeTypes {
         return C.ENCODING_DTS;
       case MimeTypes.AUDIO_DTS_HD:
         return C.ENCODING_DTS_HD;
+      case MimeTypes.AUDIO_DTS_EXPRESS:
+        return C.ENCODING_DTS; // This should be ENCODING_DTS_HD when IC platforms support it.
+      case MimeTypes.AUDIO_DTS_X:
+        return C.ENCODING_DTS; // This should be ENCODING_DTS_UHD_P2 when IC platforms support it.
       case MimeTypes.AUDIO_TRUEHD:
         return C.ENCODING_DOLBY_TRUEHD;
       case MimeTypes.AUDIO_OPUS:
--- a/libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
+++ b/libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
@@ -596,9 +596,9 @@ public final class MimeTypes {
       case MimeTypes.AUDIO_DTS_HD:
         return C.ENCODING_DTS_HD;
       case MimeTypes.AUDIO_DTS_EXPRESS:
-        return C.ENCODING_DTS; // This should be ENCODING_DTS_HD when IC platforms support it.
+        return C.ENCODING_DTS_HD;
       case MimeTypes.AUDIO_DTS_X:
-        return C.ENCODING_DTS; // This should be ENCODING_DTS_UHD_P2 when IC platforms support it.
+        return C.ENCODING_DTS_UHD_P2;
       case MimeTypes.AUDIO_TRUEHD:
         return C.ENCODING_DOLBY_TRUEHD;
       case MimeTypes.AUDIO_OPUS:
--- a/libraries/common/src/main/java/androidx/media3/common/util/Util.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/Util.java
@@ -1884,6 +1884,14 @@ public final class Util {
         return AudioFormat.CHANNEL_OUT_5POINT1 | AudioFormat.CHANNEL_OUT_BACK_CENTER;
       case 8:
         return AudioFormat.CHANNEL_OUT_7POINT1_SURROUND;
+      case 10:
+        if (Util.SDK_INT > 31) {
+          return AudioFormat.CHANNEL_OUT_5POINT1POINT4;
+        } else {
+          // This is used by DTS:X P2 with Direct Passthrough Playback.
+          // Specifying the audio format as 7.1 for 10 channels does not affect the playback.
+          return AudioFormat.CHANNEL_OUT_7POINT1_SURROUND;
+        }
       case 12:
         return AudioFormat.CHANNEL_OUT_7POINT1POINT4;
       default:
--- a/libraries/common/src/main/java/androidx/media3/common/util/Util.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/Util.java
@@ -1885,11 +1885,12 @@ public final class Util {
       case 8:
         return AudioFormat.CHANNEL_OUT_7POINT1_SURROUND;
       case 10:
-        if (Util.SDK_INT > 31) {
+        if (Util.SDK_INT >= 32) {
           return AudioFormat.CHANNEL_OUT_5POINT1POINT4;
         } else {
-          // This is used by DTS:X P2 with Direct Passthrough Playback.
-          // Specifying the audio format as 7.1 for 10 channels does not affect the playback.
+          // Before API 32, height channel masks are not available. For those 10-channel streams
+          // supported on the audio output devices (e.g. DTS:X P2), we will use 7.1-surround
+          // instead.
           return AudioFormat.CHANNEL_OUT_7POINT1_SURROUND;
         }
       case 12:
--- a/libraries/common/src/main/java/androidx/media3/common/util/Util.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/Util.java
@@ -537,6 +537,30 @@ public final class Util {
   }
 
   /**
+   * Remove duplicates from an Integer array.
+   *
+   * @param inputs The input integer array.
+   * @return The integer array without duplicates.
+   */
+  @UnstableApi
+  @SuppressWarnings("nullness:assignment")
+  public static int[] nullSafeIntegerArrayDistinct(int[] inputs) {
+    int end = inputs.length;
+    for (int i = 0; i < end; i++) {
+      for (int j = i + 1; j < end; j++) {
+        if (inputs[i] == inputs[j]) {
+          inputs[j] = inputs[end - 1];
+          end--;
+          j--;
+        }
+      }
+    }
+    int[] newlist = new int[end];
+    System.arraycopy(inputs, 0, newlist, 0, end);
+    return newlist;
+  }
+
+  /**
    * Copies the contents of {@code list} into {@code array}.
    *
    * <p>{@code list.size()} must be the same as {@code array.length} to ensure the contents can be
@@ -550,6 +574,23 @@ public final class Util {
   public static <T> void nullSafeListToArray(List<T> list, T[] array) {
     Assertions.checkState(list.size() == array.length);
     list.toArray(array);
+  }
+
+  /**
+   * Creates a new Integer array from a List of Integers.
+   *
+   * @param list The list of integers to convert.
+   * @return Created array of integers.
+   */
+  @UnstableApi
+  @SuppressWarnings("nullness:assignment")
+  public static int[] nullSafeIntegerListToArray(List<Integer> list) {
+    int[] intList = new int[list.size()];
+
+    for (int i = 0; i < list.size(); i++) {
+      intList[i] = list.get(i);
+    }
+    return intList;
   }
 
   /**
--- a/libraries/common/src/main/java/androidx/media3/common/util/Util.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/Util.java
@@ -517,6 +517,26 @@ public final class Util {
   }
 
   /**
+   * Creates a new Integer array containing the concatenation of two non-null Integer arrays.
+   *
+   * @param first  The first array.
+   * @param second The second array.
+   * @return The concatenated result.
+   */
+  @UnstableApi
+  @SuppressWarnings("nullness:assignment")
+  public static int[] nullSafeIntegerArrayConcatenation(int[] first, int[] second) {
+    int[] concatenation = Arrays.copyOf(first, first.length + second.length);
+    System.arraycopy(
+        /* src= */ second,
+        +        /* srcPos= */ 0,
+        /* dest= */ concatenation,
+        /* destPos= */ first.length,
+        /* length= */ second.length);
+    return concatenation;
+  }
+
+  /**
    * Copies the contents of {@code list} into {@code array}.
    *
    * <p>{@code list.size()} must be the same as {@code array.length} to ensure the contents can be
--- a/libraries/common/src/main/java/androidx/media3/common/util/Util.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/Util.java
@@ -517,30 +517,6 @@ public final class Util {
   }
 
   /**
-   * Remove duplicates from an Integer array.
-   *
-   * @param inputs The input integer array.
-   * @return The integer array without duplicates.
-   */
-  @UnstableApi
-  @SuppressWarnings("nullness:assignment")
-  public static int[] nullSafeIntegerArrayDistinct(int[] inputs) {
-    int end = inputs.length;
-    for (int i = 0; i < end; i++) {
-      for (int j = i + 1; j < end; j++) {
-        if (inputs[i] == inputs[j]) {
-          inputs[j] = inputs[end - 1];
-          end--;
-          j--;
-        }
-      }
-    }
-    int[] newlist = new int[end];
-    System.arraycopy(inputs, 0, newlist, 0, end);
-    return newlist;
-  }
-
-  /**
    * Copies the contents of {@code list} into {@code array}.
    *
    * <p>{@code list.size()} must be the same as {@code array.length} to ensure the contents can be
--- a/libraries/common/src/main/java/androidx/media3/common/util/Util.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/Util.java
@@ -1889,8 +1889,7 @@ public final class Util {
           return AudioFormat.CHANNEL_OUT_5POINT1POINT4;
         } else {
           // Before API 32, height channel masks are not available. For those 10-channel streams
-          // supported on the audio output devices (e.g. DTS:X P2), we will use 7.1-surround
-          // instead.
+          // supported on the audio output devices (e.g. DTS:X P2), we use 7.1-surround instead.
           return AudioFormat.CHANNEL_OUT_7POINT1_SURROUND;
         }
       case 12:
--- a/libraries/common/src/main/java/androidx/media3/common/util/Util.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/Util.java
@@ -517,26 +517,6 @@ public final class Util {
   }
 
   /**
-   * Creates a new Integer array containing the concatenation of two non-null Integer arrays.
-   *
-   * @param first  The first array.
-   * @param second The second array.
-   * @return The concatenated result.
-   */
-  @UnstableApi
-  @SuppressWarnings("nullness:assignment")
-  public static int[] nullSafeIntegerArrayConcatenation(int[] first, int[] second) {
-    int[] concatenation = Arrays.copyOf(first, first.length + second.length);
-    System.arraycopy(
-        /* src= */ second,
-        +        /* srcPos= */ 0,
-        /* dest= */ concatenation,
-        /* destPos= */ first.length,
-        /* length= */ second.length);
-    return concatenation;
-  }
-
-  /**
    * Remove duplicates from an Integer array.
    *
    * @param inputs The input integer array.
@@ -574,23 +554,6 @@ public final class Util {
   public static <T> void nullSafeListToArray(List<T> list, T[] array) {
     Assertions.checkState(list.size() == array.length);
     list.toArray(array);
-  }
-
-  /**
-   * Creates a new Integer array from a List of Integers.
-   *
-   * @param list The list of integers to convert.
-   * @return Created array of integers.
-   */
-  @UnstableApi
-  @SuppressWarnings("nullness:assignment")
-  public static int[] nullSafeIntegerListToArray(List<Integer> list) {
-    int[] intList = new int[list.size()];
-
-    for (int i = 0; i < list.size(); i++) {
-      intList[i] = list.get(i);
-    }
-    return intList;
   }
 
   /**
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -48,6 +48,8 @@ import java.util.Arrays;
 @UnstableApi
 public final class AudioCapabilities {
 
+  // TODO(internal b/283945513): Have separate default max channel counts in `AudioCapabilities`
+  // for PCM and compressed audio.
   private static final int DEFAULT_MAX_CHANNEL_COUNT = 10;
   @VisibleForTesting /* package */ static final int DEFAULT_SAMPLE_RATE_HZ = 48_000;
 
@@ -374,9 +376,12 @@ public final class AudioCapabilities {
 
     @DoNotInline
     public static ImmutableList<Integer> getDirectPlaybackSupportedEncodings() {
-      ImmutableList<Integer> encodings = Api29.getAllSurroundEncodingsMaybeSupported();
       ImmutableList.Builder<Integer> supportedEncodingsListBuilder = ImmutableList.builder();
-      for (int encoding : encodings) {
+      for (int encoding : ALL_SURROUND_ENCODINGS_AND_MAX_CHANNELS.keySet()) {
+        // AudioFormat.ENCODING_DTS_UHD_P2 is supported from API 34.
+        if (Util.SDK_INT < 34 && encoding == C.ENCODING_DTS_UHD_P2) {
+          continue;
+        }
         if (AudioTrack.isDirectPlaybackSupported(
             new AudioFormat.Builder()
                 .setChannelMask(AudioFormat.CHANNEL_OUT_STEREO)
@@ -412,20 +417,6 @@ public final class AudioCapabilities {
         }
       }
       return 0;
-    }
-
-    /** Returns a list of surround encodings that maybe supported. */
-    @DoNotInline
-    private static ImmutableList<Integer> getAllSurroundEncodingsMaybeSupported() {
-      ImmutableList.Builder<Integer> encodings = new ImmutableList.Builder<>();
-      for (int encoding : ALL_SURROUND_ENCODINGS_AND_MAX_CHANNELS.keySet()) {
-        // AudioFormat.ENCODING_DTS_UHD_P2 is supported from API 34.
-        if (Util.SDK_INT < 34 && encoding == C.ENCODING_DTS_UHD_P2) {
-          continue;
-        }
-        encodings.add(encoding);
-      }
-      return encodings.build();
     }
   }
 }
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -101,8 +101,8 @@ public final class AudioCapabilities {
     if (Util.SDK_INT >= 23 && Api23.isBluetoothConnected(context)) {
       return DEFAULT_AUDIO_CAPABILITIES;
     }
-    ImmutableSet.Builder supportedEncodings = new ImmutableSet.Builder<>();
 
+    ImmutableSet.Builder supportedEncodings = new ImmutableSet.Builder<>();
     if (deviceMaySetExternalSurroundSoundGlobalSetting()
         && Global.getInt(context.getContentResolver(), EXTERNAL_SURROUND_SOUND_KEY, 0) == 1) {
       supportedEncodings.addAll(Ints.asList(EXTERNAL_SURROUND_SOUND_ENCODINGS));
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -119,7 +119,9 @@ public final class AudioCapabilities {
       supportedEncodings.addAll(Ints.asList(
           intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)));
       return new AudioCapabilities(
-          Ints.toArray(supportedEncodings.build()), /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT);
+          Ints.toArray(supportedEncodings.build()),
+          intent.getIntExtra(AudioManager.EXTRA_MAX_CHANNEL_COUNT, /* defaultValue= */
+              DEFAULT_MAX_CHANNEL_COUNT));
     }
 
     if (supportedEncodings.build().isEmpty()) {
@@ -227,7 +229,13 @@ public final class AudioCapabilities {
       channelCount = getMaxSupportedChannelCountForPassthrough(encoding, sampleRate);
     } else {
       channelCount = format.channelCount;
-      if (channelCount > maxChannelCount) {
+      // To file a Bug: Some DTS:X TVs reports ACTION_HDMI_AUDIO_PLUG.EXTRA_MAX_CHANNEL_COUNT as 8
+      // instead of 10.
+      if (format.sampleMimeType == MimeTypes.AUDIO_DTS_X) {
+        if (channelCount > 10) {
+          return null;
+        }
+      } else if (channelCount > maxChannelCount) {
         return null;
       }
     }
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -44,6 +44,8 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.primitives.Ints;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 /** Represents the set of audio formats that a device is capable of playing. */
 @UnstableApi
@@ -402,7 +404,7 @@ public final class AudioCapabilities {
     /**
      * Returns an array list of surround encodings that maybe supported.
      */
-    private static ArrayList<Integer> getAllSurroundEncodingsMaybeSupported() {
+    private static List<Integer> getAllSurroundEncodingsMaybeSupported() {
       ArrayList<Integer> encodings = new ArrayList<>();
       for (int encoding : ALL_SURROUND_ENCODINGS_AND_MAX_CHANNELS.keySet()) {
         // AudioFormat.ENCODING_DTS_UHD_P2 is supported from API 34.
@@ -411,7 +413,7 @@ public final class AudioCapabilities {
         }
         encodings.add(encoding);
       }
-      return encodings;
+      return Collections.unmodifiableList(encodings);
     }
   }
 }
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -59,8 +59,9 @@ public final class AudioCapabilities {
   @SuppressWarnings("InlinedApi")
   private static final AudioCapabilities EXTERNAL_SURROUND_SOUND_CAPABILITIES =
       new AudioCapabilities(
-          new int[] {
-            AudioFormat.ENCODING_PCM_16BIT, AudioFormat.ENCODING_AC3, AudioFormat.ENCODING_E_AC3
+          new int[]{
+              AudioFormat.ENCODING_PCM_16BIT, AudioFormat.ENCODING_AC3, AudioFormat.ENCODING_E_AC3,
+              AudioFormat.ENCODING_DTS, AudioFormat.ENCODING_DTS_HD
           },
           DEFAULT_MAX_CHANNEL_COUNT);
 
@@ -220,7 +221,13 @@ public final class AudioCapabilities {
       channelCount = getMaxSupportedChannelCountForPassthrough(encoding, sampleRate);
     } else {
       channelCount = format.channelCount;
-      if (channelCount > maxChannelCount) {
+      if (format.sampleMimeType == MimeTypes.AUDIO_DTS_X) {
+        if (channelCount > 10) {
+          // To fix wrong reporting from device. ChannelCount is reported as 8 for DTS:X P2
+          // on some devices.
+          return null;
+        }
+      } else if (channelCount > maxChannelCount) {
         return null;
       }
     }
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -123,11 +123,12 @@ public final class AudioCapabilities {
               AudioManager.EXTRA_MAX_CHANNEL_COUNT, /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT));
     }
 
-    if (supportedEncodings.build().isEmpty()) {
+    ImmutableSet supportedEncodingsSet = supportedEncodings.build();
+    if (supportedEncodingsSet.isEmpty()) {
       return DEFAULT_AUDIO_CAPABILITIES;
     } else {
       return new AudioCapabilities(
-          Ints.toArray(supportedEncodings.build()), /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT);
+          Ints.toArray(supportedEncodingsSet), /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT);
     }
   }
 
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -58,9 +58,7 @@ public final class AudioCapabilities {
   /** Encodings supported when the device specifies external surround sound. */
   private static final ImmutableList<Integer> EXTERNAL_SURROUND_SOUND_ENCODINGS =
       ImmutableList.of(
-          AudioFormat.ENCODING_PCM_16BIT,
-          AudioFormat.ENCODING_AC3,
-          AudioFormat.ENCODING_E_AC3);
+          AudioFormat.ENCODING_PCM_16BIT, AudioFormat.ENCODING_AC3, AudioFormat.ENCODING_E_AC3);
 
   /**
    * All surround sound encodings that a device may be capable of playing mapped to a maximum
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -49,7 +49,7 @@ import java.util.Arrays;
 @UnstableApi
 public final class AudioCapabilities {
 
-  private static final int DEFAULT_MAX_CHANNEL_COUNT = 8;
+  private static final int DEFAULT_MAX_CHANNEL_COUNT = 10;
   @VisibleForTesting /* package */ static final int DEFAULT_SAMPLE_RATE_HZ = 48_000;
 
   /** The minimum audio capabilities supported by all devices. */
@@ -224,12 +224,7 @@ public final class AudioCapabilities {
       channelCount = getMaxSupportedChannelCountForPassthrough(encoding, sampleRate);
     } else {
       channelCount = format.channelCount;
-      if (format.sampleMimeType == MimeTypes.AUDIO_DTS_X) {
-        if (channelCount > 10) {
-          // To fix wrong reporting from device. ChannelCount is reported as 8 for DTS:X P2 on some devices
-          return null;
-        }
-      } else if (channelCount > maxChannelCount) {
+      if (channelCount > maxChannelCount) {
         return null;
       }
     }
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -115,19 +115,19 @@ public final class AudioCapabilities {
       supportedEncodings.addAll(Ints.asList(Api29.getDirectPlaybackSupportedEncodings()));
     }
 
-    if (intent == null || intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 0) {
-      if (supportedEncodings.build().isEmpty()) {
-        return DEFAULT_AUDIO_CAPABILITIES;
-      } else {
-        return new AudioCapabilities(Ints.toArray(supportedEncodings.build()), /* defaultValue= */
-            DEFAULT_MAX_CHANNEL_COUNT);
-      }
+    if (intent == null || intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 1) {
+      supportedEncodings.addAll(Ints.asList(
+          intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)));
+      return new AudioCapabilities(
+          Ints.toArray(supportedEncodings.build()), /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT);
     }
 
-    supportedEncodings.addAll(Ints.asList(
-        intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)));
-    return new AudioCapabilities(
-        Ints.toArray(supportedEncodings.build()), /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT);
+    if (supportedEncodings.build().isEmpty()) {
+      return DEFAULT_AUDIO_CAPABILITIES;
+    } else {
+      return new AudioCapabilities(Ints.toArray(supportedEncodings.build()), /* defaultValue= */
+          DEFAULT_MAX_CHANNEL_COUNT);
+    }
   }
 
   /**
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -115,7 +115,7 @@ public final class AudioCapabilities {
       supportedEncodings.addAll(Ints.asList(Api29.getDirectPlaybackSupportedEncodings()));
     }
 
-    if (intent == null || intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 1) {
+    if (intent != null && intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 1) {
       supportedEncodings.addAll(Ints.asList(
           intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)));
       return new AudioCapabilities(
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -57,8 +57,8 @@ public final class AudioCapabilities {
 
   /** Encodings supported when the device specifies external surround sound. */
   private static final int[] EXTERNAL_SURROUND_SOUND_ENCODINGS =
-      new int[]{
-          AudioFormat.ENCODING_PCM_16BIT, AudioFormat.ENCODING_AC3, AudioFormat.ENCODING_E_AC3
+      new int[] {
+        AudioFormat.ENCODING_PCM_16BIT, AudioFormat.ENCODING_AC3, AudioFormat.ENCODING_E_AC3
       };
 
   /**
@@ -116,19 +116,18 @@ public final class AudioCapabilities {
     }
 
     if (intent != null && intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 1) {
-      supportedEncodings.addAll(Ints.asList(
-          intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)));
+      supportedEncodings.addAll(Ints.asList(intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)));
       return new AudioCapabilities(
           Ints.toArray(supportedEncodings.build()),
-          intent.getIntExtra(AudioManager.EXTRA_MAX_CHANNEL_COUNT, /* defaultValue= */
-              DEFAULT_MAX_CHANNEL_COUNT));
+          intent.getIntExtra(
+              AudioManager.EXTRA_MAX_CHANNEL_COUNT, /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT));
     }
 
     if (supportedEncodings.build().isEmpty()) {
       return DEFAULT_AUDIO_CAPABILITIES;
     } else {
-      return new AudioCapabilities(Ints.toArray(supportedEncodings.build()), /* defaultValue= */
-          DEFAULT_MAX_CHANNEL_COUNT);
+      return new AudioCapabilities(
+          Ints.toArray(supportedEncodings.build()), /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT);
     }
   }
 
@@ -211,8 +210,8 @@ public final class AudioCapabilities {
     if (encoding == C.ENCODING_E_AC3_JOC && !supportsEncoding(C.ENCODING_E_AC3_JOC)) {
       // E-AC3 receivers support E-AC3 JOC streams (but decode only the base layer).
       encoding = C.ENCODING_E_AC3;
-    } else if ((encoding == C.ENCODING_DTS_HD && !supportsEncoding(C.ENCODING_DTS_HD)) ||
-        (encoding == C.ENCODING_DTS_UHD_P2 && !supportsEncoding(C.ENCODING_DTS_UHD_P2))) {
+    } else if ((encoding == C.ENCODING_DTS_HD && !supportsEncoding(C.ENCODING_DTS_HD))
+        || (encoding == C.ENCODING_DTS_UHD_P2 && !supportsEncoding(C.ENCODING_DTS_UHD_P2))) {
       // DTS receivers support DTS-HD streams (but decode only the core layer).
       encoding = C.ENCODING_DTS;
     }
@@ -411,9 +410,7 @@ public final class AudioCapabilities {
       return 0;
     }
 
-    /**
-     * Returns an array list of surround encodings that maybe supported.
-     */
+    /** Returns an array list of surround encodings that maybe supported. */
     @DoNotInline
     private static int[] getAllSurroundEncodingsMaybeSupported() {
       ImmutableList.Builder<Integer> encodings = new ImmutableList.Builder<>();
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -214,8 +214,7 @@ public final class AudioCapabilities {
       return null;
     }
     int channelCount;
-    if (format.channelCount == Format.NO_VALUE || encoding == C.ENCODING_E_AC3_JOC
-        || encoding == C.ENCODING_DTS) {
+    if (format.channelCount == Format.NO_VALUE || encoding == C.ENCODING_E_AC3_JOC) {
       // In HLS chunkless preparation, the format channel count and sample rate may be unset. See
       // https://github.com/google/ExoPlayer/issues/10204 and b/222127949 for more details.
       // For E-AC3 JOC, the format is object based so the format channel count is arbitrary.
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
@@ -102,7 +102,7 @@ public final class AudioCapabilities {
       return DEFAULT_AUDIO_CAPABILITIES;
     }
 
-    ImmutableSet.Builder supportedEncodings = new ImmutableSet.Builder<>();
+    ImmutableSet.Builder<Integer> supportedEncodings = new ImmutableSet.Builder<>();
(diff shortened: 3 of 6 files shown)


cedricxperi:
Resolved some of the conflicts in the readme files. Hope it's correct. Thanks.

cedricxperi:
@christosts Have been waiting on a response from @tianyif for some time now. Am not sure how we can move forward on this pull requets. Please let me know if there is anything I need to do. Thanks.

cedricxperi:
> > @tianyif Hi, thanks for reviewing this pull request. Kindly let me know if there is anything I can do to help move this along. Thank you.
> 
> Hi @cedricxperi, sorry for the late reply. I was on the other stuffs last week, and I'm starting reviewing this now.

@tianyif No Worries. Thanks for reviewing this!

cedricxperi:
@tianyif Just FYI. I am in Asian (GMT + 8) timezone. Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -58,8 +58,9 @@ public final class AudioCapabilities {
   @SuppressWarnings("InlinedApi")
   private static final AudioCapabilities EXTERNAL_SURROUND_SOUND_CAPABILITIES =
       new AudioCapabilities(
-          new int[] {
-            AudioFormat.ENCODING_PCM_16BIT, AudioFormat.ENCODING_AC3, AudioFormat.ENCODING_E_AC3
+          new int[]{
+              AudioFormat.ENCODING_PCM_16BIT, AudioFormat.ENCODING_AC3, AudioFormat.ENCODING_E_AC3,
+              AudioFormat.ENCODING_DTS, AudioFormat.ENCODING_DTS_HD
```
tianyif, line 63: The `EXTERNAL_SURROUND_SOUND_CAPABILITIES` that the new encodings are added to, is available to Amazon or Xiaomi device. Did the manufacturers confirm that they support those new encodings?

cedricxperi: @tianyif Yes. Xperi worked with both manufacturers to make sure that they support those encodings. Thanks.

libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
```diff
@@ -590,6 +590,10 @@ public static String getMimeTypeFromMp4ObjectType(int objectType) {
         return C.ENCODING_DTS;
       case MimeTypes.AUDIO_DTS_HD:
         return C.ENCODING_DTS_HD;
+      case MimeTypes.AUDIO_DTS_EXPRESS:
+        return C.ENCODING_DTS; // This should be ENCODING_DTS_HD when IC platforms support it.
```
tianyif, line 594: Does comment mean if the platform doesn't support `ENCODING_DTS_HD`, then it can be treated as `ENCODING_DTS`? 

Maybe we can directly return `C.ENCODING_DTS_HD` here, because when doing the actual format support querying, we will look at if `C.ENCODING_DTS_HD` is supported according to the audio capabilities. If not, we will treat it as the `C.ENCODING_DTS` - https://github.com/androidx/media/blob/1ef427f640277dc8e0ad09ebf86be972f216eb82/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java#L198-L200

libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
```diff
@@ -590,6 +590,10 @@ public static String getMimeTypeFromMp4ObjectType(int objectType) {
...
+      case MimeTypes.AUDIO_DTS_X:
+        return C.ENCODING_DTS; // This should be ENCODING_DTS_UHD_P2 when IC platforms support it.
```
tianyif, line 596: This is similar to the above one but slightly different:

`AudioFormat.ENCODING_DTS_UHD_P2` is [in Android 14 Beta](https://developer.android.com/reference/android/media/AudioFormat#ENCODING_DTS_UHD_P2), so currently we cannot refer to it in the constant class `C`. Could you please add `ENCODING_DTS_UHD_P2` in the `C` class and refer to its constant value (`0x0000001e`), then we can change it to `AudioFormat.ENCODING_DTS_UHD_P2` once it is released.

libraries/common/src/main/java/androidx/media3/common/util/Util.java
```diff
@@ -1882,6 +1882,14 @@ public static int getAudioTrackChannelConfig(int channelCount) {
...
       case 8:
         return AudioFormat.CHANNEL_OUT_7POINT1_SURROUND;
+      case 10:
+        if (Util.SDK_INT > 31) {
+          return AudioFormat.CHANNEL_OUT_5POINT1POINT4;
+        } else {
+          // This is used by DTS:X P2 with Direct Passthrough Playback.
+          // Specifying the audio format as 7.1 for 10 channels does not affect the playback.
+          return AudioFormat.CHANNEL_OUT_7POINT1_SURROUND;
+        }
```
tianyif, line 1892: This change conflicts with an ongoing pull request from Dolby in which the case 10 will always return 5.1.4. I will consult with the platform team today what can happen (for SDK <= 31) when creating `AudioTrack` with channel count 5.1.4 for DTS:X P2 stream, or with channel count 7.1 for DD+JOC stream.

If there is any potential problem with that, I'm afraid that we should refactor this method `getAudioTrackChannelConfig` first to take the mime type into consideration.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
tianyif: Acknowledged, thanks!

libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
cedricxperi: Currently, to playback DTS audio in tunnel mode via Direct playback mode, it is better to create an AudioTrack with ENCODING_DTS regardless of whether the incoming DTS stream is  DTS Digital surround, DTS Express or DTS:X Profile 2. I know creating AudioTrack on some MTK platforms with ENCODING_DTS_HD will cause a crash. We are working with IC vendors to resolve this. So in future we should be using ENCODING_DTS_HD for DTS Express. But for now, it should be only ENCODING_DTS. Thanks.

libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
cedricxperi: I think this could be a problem because creation of AudioTrack with 0x0000001e will fail on the TV platforms that we are testing. I think it is because these Android TVs are not Android 14. So for now can we stick to ENCODING_DTS?

libraries/common/src/main/java/androidx/media3/common/util/Util.java
cedricxperi: Creating an AudioTrack for DTS:X P2 with AudioFormat.CHANNEL_OUT_5POINT1POINT4 (taking the integer value) will fail on MTK platforms. That is why I switched it to AudioFormat.CHANNEL_OUT_7POINT1_SURROUND for API <=31. :)

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -41,6 +41,9 @@ public final class DtsUtil {
   private static final int SYNC_VALUE_14B_BE = 0x1FFFE800;
   private static final int SYNC_VALUE_LE = 0xFE7F0180;
   private static final int SYNC_VALUE_14B_LE = 0xFF1F00E8;
+  private static final int SYNC_EXT_SUB_LE = 0x25205864;
```
tianyif, line 44: Do we have a specification pointing to these new values?

libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
tianyif: I see. And I saw the offload logic is also [calling this method].(https://github.com/androidx/media/blob/release/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L1716), so any change may potentially impact the offload case.

If we return `ENCODING_DTS_HD` and it is supported by the platform for offload playback, then an `AudioTrack` will be [tried to create with `ENCODING_DTS_HD`](https://github.com/androidx/media/blob/release/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L762-L763), and leads to the MTK platform issue you mentioned. The current implementation may be the simplest way to care about both passthrough and offload cases.

cc: @christosts

libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
tianyif: Good point! I think we should look into this carefully, again, this method is also called by offload logic.

When looking at [if offload is supported or not](https://github.com/androidx/media/blob/release/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L1716). The platform gets the int value returned by this method when evaluating the offload support, it tries to [convert the value](https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/android13-release/core/jni/android_media_AudioSystem.cpp#2300) to the native constant. If we pass an `ENCODING_DTS_UHD_P2` value, for Android <= 13, in the absence of definition of `ENCODING_DTS_UHD_P2`, the native constant converted to will be `AUDIO_FORMAT_INVALID`. Then I supposed that will lead to a "offload unsupported" result, then the DTS:X P2 stream cannot be offloaded played, when the device support `ENCODING_DTS` for offload playback.

cc @christosts - If we return `ENCODING_DTS` for mime type DTS:X P2, the offload logic will check against the support of `ENCODING_DTS` from the platform, if the platform said `ENCODING_DTS` is supported, an `AudioTrack` is created with `ENCODING_DTS`, then is it ok that DTS_UHD_P2 is actually played on that `AudioTrack`?

libraries/common/src/main/java/androidx/media3/common/util/Util.java
tianyif: Acknowledged, we will discuss internally tomorrow on the potential refactorization of `getAudioTrackChannelConfig`. Thanks!

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
cedricxperi: 1) Extension Substream Syncword (in different Endianness) are defined in Section 7.4.1:
https://www.etsi.org/deliver/etsi_ts/102100_102199/102114/01.06.01_60/ts_102114v010601p.pdf

2) The 2 FTOC Sync words (in different Endianness) are defined in Section 6.4.4.1 of
https://www.etsi.org/deliver/etsi_ts/103400_103499/103491/01.02.01_60/ts_103491v010201p.pdf

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -219,7 +220,13 @@ public Pair<Integer, Integer> getEncodingAndChannelConfigForPassthrough(Format f
       channelCount = getMaxSupportedChannelCountForPassthrough(encoding, sampleRate);
     } else {
       channelCount = format.channelCount;
-      if (channelCount > maxChannelCount) {
+      if (format.sampleMimeType == MimeTypes.AUDIO_DTS_X) {
+        if (channelCount > 10) {
+          // To fix wrong reporting from device. ChannelCount is reported as 8 for DTS:X P2
```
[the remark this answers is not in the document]
cedricxperi: @tianyif Please see my latest commit aecc8dd858ea9fa4fbbbe35dae73e65ba4c84718 for the change to use getMaxSupportedChannelCountForPassthrough(encoding, sample). So we don't really need to do 
"if (format.sampleMimeType == MimeTypes.AUDIO_DTS_X) {
        if (channelCount > 10) {" etc.
       
Thanks.

libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
cedricxperi: @tianyif Anyway, DTS_Express is supposed to be transcoded to DTS Digital Surround. We do not allow passthrough of DTS Express to audio sink devices directly. In my experience, the support for DTS Express using ENCODING_DTS_HD AudioTrack is inconsistent. To solve all these issues, we use tunnel mode with direct playback for DTS Express. The downstream platform's DSP will take care of decoding/passthrough/transcoding. So an application such as Exoplayer just needs to send the frames downstream to the platform's DSP (Decoder) via AudioTrack.

libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
christosts: Similar to previous comments, I think it's better to have logic to handle MimeTypes.AUDIO_DTS_X as ENCODING_DTS in components closer to the audio rather than putting it in MimeTypes, because putting this here means we map MimeTypes.AUDIO_DTS_X to ENCODING_DTS which is not accurate. 

> I think this could be a problem because creation of AudioTrack with 0x0000001e will fail on the TV platforms that we are testing. 

For ExoPlayer to create an AudioTrack with 0x0000001e it means the platform reported that such an encoding is supported in the first place, which should not happen.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
christosts: on what API versions did Amazon and Xiaomi include the new format? The code currently assumes they are on API 17+. Do all Amazon and Xiaomi devices with API 17+ support ENCODING_DTS and ENCODING_DTS_HD? Do we need to add further API checks?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -71,7 +72,7 @@ public final class AudioCapabilities {
       new ImmutableMap.Builder<Integer, Integer>()
           .put(C.ENCODING_AC3, 6)
           .put(C.ENCODING_AC4, 6)
-          .put(C.ENCODING_DTS, 6)
+          .put(C.ENCODING_DTS, 10)
```
christosts, line 75: is this per-DTS spec or is it because the pull-request has MimeTypes to map AUDIO_DTS_EXPRESS and AUDIO_DTS_X to ENCODING_DTS?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -210,7 +211,8 @@ public Pair<Integer, Integer> getEncodingAndChannelConfigForPassthrough(Format f
       return null;
     }
     int channelCount;
-    if (format.channelCount == Format.NO_VALUE || encoding == C.ENCODING_E_AC3_JOC) {
+    if (format.channelCount == Format.NO_VALUE || encoding == C.ENCODING_E_AC3_JOC
+        || encoding == C.ENCODING_DTS) {
       // In HLS chunkless preparation, the format channel count and sample rate may be unset. See
       // https://github.com/google/ExoPlayer/issues/10204 and b/222127949 for more details.
       // For E-AC3 JOC, the format is object based so the format channel count is arbitrary.
```
christosts, line 218: We need to extend the comment here and explain why we don't care about the container's (compressed bitsteam) channel count for ENCODING_DTS too. 

Also, this is probably *not* true for ENCODING_DTS but one of AUDIO_DTS_EXPRESS or AUDIO_DTS_X, right? This goes back to the other comment that we need to change the logic here to handle AUDIO_DTS_EXPRESS and AUDIO_DTS_X instead of MimeTypes

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: It's because the "pull-request has MimeTypes to map AUDIO_DTS_EXPRESS and AUDIO_DTS_X to ENCODING_DTS". This mapping is necessary for Video tunnel mode + Audio Direct Passthrough Playback. All DTS audio types will be sent to MTK/Realtek platform via AudioTrack with ENCODING_DTS. Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @christosts I agree. I think it is better to add API checks. Let me check which API level and get back. Thanks.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -149,6 +152,19 @@ public static int parseDtsAudioSampleCount(byte[] data) {
...
+    // DTS:X Profile 2's FrameSize = 1024.
+    if ((buffer.getInt(0) == SYNC_FTOC_LE) ||
+        (buffer.getInt(0) == SYNC_FTOC_NON_SYNC_LE)) {
+      return 1024;
+    }
+    // Check for sync word for DTS Express.
+    // DTS Express's FrameSize = 4096.
+    else if (buffer.getInt(0) == SYNC_EXT_SUB_LE) {
+      return 4096;
+    }
```
tianyif, line 166: Could you please add the specification where suggests the frame sizes to the comment?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
tianyif: Agree. Gathering the idea from our comments so far, I think this should be something like `format.channelCount == Format.NO_VALUE || encoding == C.ENCODING_E_AC3_JOC || encoding == C.ENCODING_DTS_HD || encoding == C.ENCODING_DTS_UHD_P2` later.

And the comment explaining the new cases is definitely needed here :)

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
cedricxperi: @tianyif We are controlling the possible frame sizes from the encoder. Right now, these are the only possible frame sizes that are in use for legitimate streaming. The enforcement of these frame sizes are not listed in the specs.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
cedricxperi: @tianyif Do I need to add these as comments? Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @christosts It will be pretty hard to do the API checks as some of the Xiaomi devices may not be consistent in terms of DTS support across the same API. We will stream to newer versions of XIaomi TV that should have support for ENCODING_DTS and ENCODING_DTS_HD. Is it ok if we leave it at API 17+? Thanks.

libraries/common/src/main/java/androidx/media3/common/MimeTypes.java
cedricxperi: Please see 61a6b995104a3e40a9e27d223eceb8e79452ae2d.  C.ENCODING_DTS_UHD_P2 == 0x1e has been created.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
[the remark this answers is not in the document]
cedricxperi: @tianyif I mis-intepreted the meaning of arbitrary to mean whether channelcount can be any value when using direct passthrough mode.  
In this case, the platform has mistakenly reported max channel count as 8 for DTS:X P2. The correct maximum channel count should be 10. Therefore, I think is it better to stick to the original code that fixes this platform mistake. Please see commit 61a6b995104a3e40a9e27d223eceb8e79452ae2d. Thanks

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @christosts I have reset the value of C.ENCODING_DTS back to 6 and added C.ENCODING_DTS_UHD_P2=10 . Hope this is more correct. Kindly see 61a6b995104a3e40a9e27d223eceb8e79452ae2d. Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif This change has been reverted. DTS Encoding's format.channelcount should not be skipped. Please see dc85ffb17456fe7cfa6eb135025665800948de1a  Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
christosts: This is going to regress on apps that already have DTS or DTS_HD encodings included in DASH/HLS content that are played on Amazon and Xiaomi devices that do not support the encodings. Once these apps upgrade to latest of ExoPlayer, the player will attempt to play DTS or DTS_HD via passthrough, which likely is going to fail during creation of AudioTrack later.

I do not know how many apps may fall in this category, so I can't evaluate the impact. Do you have any data on this?

On the other hand, do you know if the Amazon and Xiaomi devices that support the new audio formats are after API 29? It seems safer to change the logic here and query the platform for direct support of DTS and DTS_HD with AudioTrack.isDirectAudioPlaybackSupported(), similar to how it's done for other devices.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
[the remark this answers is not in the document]
christosts: Is there a issue on buganizer or any other issue ID we can add as a comment as to this? It's not uncommon we go back to pieces of code that have assumptions like this one, and having the ability to get more context helps.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @christosts I was trying to let isDirectAudioPlaybackSupported() do it's job as well. The issue was deviceMaySetExternalSurroundSoundGlobalSetting() will return EXTERNAL_SURROUND_SOUND_CAPABILITIES (which originally did not include DTS encodings) for all Xiaomi TVs, before isDirectAudioPlaybackSupported() has a chance to run. The Xiaomi TV that I have does support ENCODING_DTS and ENCODING_DTS_HD. But the playback stopped because EXTERNAL_SURROUND_SOUND_CAPABILITIES did not include DTS encodings.
So this part of the code is to rectify that.

If there is another way to solve this problem, I would be most happy to try it out. 

I think Amazon and Xiaomi Android 8.0 API Level 26 and above should support these encodings. ENCODING_DTS and ENCODING_DTS_HD were defined in API 23 onwards.

Regarding the impact, I can give you more data through email as this involves some business information which is best communicated via email or some private channel. 

Thanks

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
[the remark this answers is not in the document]
cedricxperi: @christosts Thanks for the tip. Will file a bug tomorrow and add a comment.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
tianyif: Yes, that would be appreciated! Thanks!

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
tianyif: Yes, thanks!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
christosts: > If there is another way to solve this problem, I would be most happy to try it out.

For API 29+, we can refactor the code so that for DTS and DTS_HD we query the platform, re-using Api29.getDirectPlaybackSupportedEncodings(). This means we need to change the method to take an argument for the queried encodings. At the moment this method is querying the platform with some fixed properties for audio format and attributes. If the same assumptions work for the DTS and DTS_HD, then we can change the method to only accept an int[] as argument

> I think Amazon and Xiaomi Android 8.0 API Level 26 and above should support these encodings. ENCODING_DTS and ENCODING_DTS_HD were defined in API 23 onwards.

Do the TVs advertise support for DTS/DTS_HD over HDMI? If yes, we pick the encodings from `intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)` which is used further below, right?

So basically, the code here will return

`EXTERNAL_SURROUND_SOUND_CAPABILITIES (=> PCM, AC3, EAC3) +`
+ `if API >= 29: use directPlaybackSupported([DTS, DTS_HD])`
+ `else look at intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)`


> Regarding the impact, I can give you more data through email as this involves some business information which is best communicated via email or some private channel.

Please send us an email at dev.exoplayer@gmail.com with subject "Issue #335". As I said we need to evaluate the risk of causing a regression to existing apps on Amazon and Xiaomi devices that don't support DTS and DTS_HD

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -357,6 +365,10 @@ private Api29() {}
     public static int[] getDirectPlaybackSupportedEncodings() {
       ImmutableList.Builder<Integer> supportedEncodingsListBuilder = ImmutableList.builder();
       for (int encoding : ALL_SURROUND_ENCODINGS_AND_MAX_CHANNELS.keySet()) {
+        // Skip ENCODING_DTS_UHD_P2 if API < 34. Otherwise setEncoding will crash.
```
tianyif, line 368: I'm thinking of this check is likely to be extended longer as more encodings are added with the platform evolving. How about we create another private static function under Api29 like

```
private static final ArrayList<Integer> getAllSurroundEncodingsMaybeSupported() {
      ArrayList<Integer> encodings = new ArrayList<Integer>();
      for (int encoding : ALL_SURROUND_ENCODINGS_AND_MAX_CHANNELS.keySet()) {
        // AudioFormat.ENCODING_DTS_UHD_P2 is supported from API 34.
        if (Util.SDK_INT >= 34 && encoding == C.ENCODING_DTS_UHD_P2) {
          continue;
        }
        encodings.add(encoding);
      }
      return encodings;
    }
```
And we iterate the array `getAllSurroundEncodingsMaybeSupported()` instead of the current `ALL_SURROUND_ENCODINGS_AND_MAX_CHANNELS` in line 367. This will introduce one more iteration when calling `getAllSurroundEncodingsMaybeSupported()` , considering `ALL_SURROUND_ENCODINGS_AND_MAX_CHANNELS` is in a quite limited size, so should be fine?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
cedricxperi: @tianyif Done. Please see 5c84b879f3837b289fd7af9c70eb55078adb705b.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
cedricxperi: @tianyif Done. Please see 5c84b879f3837b289fd7af9c70eb55078adb705b. Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif Add getAllSurroundEncodingsMaybeSupported() as suggested. Please see d9129c2e0bebeb930e269408595e2d5da7043062  Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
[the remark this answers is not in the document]
cedricxperi: @tianyif Set AudioCapabilities.DEFAULT_MAX_CHANNEL_COUNT to 10. Please see a9274b975621139ec61035408a04aa994a3480ce). Let me know if it is ok. Thanks

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @christosts  API 29+ for getDirectPlaybackSupportedEncodings() is safe for ENCODING_DTS and ENCODING_DTS_HD. Also, yes, TVs will advertise support for DTS/DTS_HD over HDMI. We check all Android based TVs for this during certification of the TV. So I think your idea of 

EXTERNAL_SURROUND_SOUND_CAPABILITIES (=> PCM, AC3, EAC3) +
- if API >= 29: use directPlaybackSupported([DTS, DTS_HD])
- else look at intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)

should work.

cedricxperi: @christosts Are you referring to something like below? There is still bugs that I need to work through. E.g. The concatenated list of encodings contain duplicates which needs to be removed.

static AudioCapabilities getCapabilities(Context context, @Nullable Intent intent) {
    
    if (Util.SDK_INT >= 23 && Api23.isBluetoothConnected(context)) {
      return DEFAULT_AUDIO_CAPABILITIES;
    }
    int[] encodings = {};
    if (deviceMaySetExternalSurroundSoundGlobalSetting()
        && Global.getInt(context.getContentResolver(), EXTERNAL_SURROUND_SOUND_KEY, 0) == 1) {
      encodings = EXTERNAL_SURROUND_SOUND_CAPABILITIES;
    }
    
    int[] encodings2 = {};
    if (Util.SDK_INT >= 29 && (Util.isTv(context) || Util.isAutomotive(context))) {
      encodings2 = Util.nullSafeIntegerArrayConcatenation(encodings,
          Api29.getDirectPlaybackSupportedEncodings());
    }
    if (intent == null || intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 0) {
      if (encodings2.length == 0) {
        return DEFAULT_AUDIO_CAPABILITIES;
      } else {
        return new AudioCapabilities(encodings2, /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT);
      }
    }
    encodings = Util.nullSafeIntegerArrayConcatenation(encodings2,
        intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS));
    return new AudioCapabilities(
        encodings,
        intent.getIntExtra(
            AudioManager.EXTRA_MAX_CHANNEL_COUNT, /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT));
  }

cedricxperi: @christosts If we use the above code, EXTERNAL_SURROUND_SOUND_CAPABILITIES can remain unchanged (without ENCODING_DTS and ENCODING_DTS_HD).

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -41,6 +41,17 @@ public final class DtsUtil {
   private static final int SYNC_VALUE_14B_BE = 0x1FFFE800;
   private static final int SYNC_VALUE_LE = 0xFE7F0180;
   private static final int SYNC_VALUE_14B_LE = 0xFF1F00E8;
+  /**
+   * DTS Extension Substream Syncword (in different Endianness) is defined in Section 7.4.1 of:
```
tianyif, line 45: Could you please reword this to align it with the other comments in this file? For example:

"DTS Extension Substream Syncword (in different endianness). See ETSI TS 103 491 section 7.4.1."

No need to put the whole link I think. The same applies to the other new comment below.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
tianyif: Thanks for the change! Could we return a immutable list instead of the mutable one? Sorry my bad, I should have included this in the above snippet.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
christosts: With the suggested change, you need to figure the duplicates as you said, and also you query the platform twice on devices with API 29+.

I was thinking that we'd only change the if-block for the Amazon-XIaomi case, from
```
if (deviceMaySetExternalSurroundSoundGlobalSetting()
        && Global.getInt(context.getContentResolver(), EXTERNAL_SURROUND_SOUND_KEY, 0) == 1) {
  return EXTERNAL_SURROUND_SOUND_CAPABILITIES;
}
```

to something like

```
if (deviceMaySetExternalSurroundSoundGlobalSetting()
        && Global.getInt(context.getContentResolver(), EXTERNAL_SURROUND_SOUND_KEY, 0) == 1) {
  return getExternalSurroundCapabilities(intent);
}

// pseudo-code
private static getExternalSurroundCapabilities(@Nullable Intent intent) {
int [] supportedEncodings;
if (Util.SDK_INT >= 29) {
  supportedEncodings = Api29.getDirectPlaybackSupportedEncodings(DTS, DTS_HD);
 return EXTERNAL_SURROUND_SOUND_CAPABILITIES + supportedEncodings;
}

if (intent != null && intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 0)) {
 ...
}
```

I had in mind that we can change  Api29.getDirectPlaybackSupportedEncodings() to take an argument for which encodings to query. For the Amazon/Xiaomi branch, we only query for DTS/DTS_HD where as where we query for all surround encodings furtherdow, we can pass the argument `ALL_SURROUND_ENCODINGS_AND_MAX_CHANNELS.keySet()` (the type of the argument can be int[], or Set<Integer>, I don't think it matters too much)

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif Done. Please see 4a9c804f844dcb062cf1bd1a45021c3f536b78dd. Thanks.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
cedricxperi: @tianyif Done. Please see 9a1bde9e5f4e4fdfa407d8630d555bd563b5c1be. Thanks

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @christosts  Please see 2bbd9891c2797932a7ee3b7a8bf59dda21e09d2f. I tried to stick as close as possible to your recommendations above.  ALL_SURROUND_ENCODINGS_AND_MAX_CHANNELS.keySet() was refactored to return from getAllSurroundEncodingsMaybeSupported() as proposed by @tianyif  earlier. Let me know. Thanks.

libraries/common/src/main/java/androidx/media3/common/util/Util.java
cedricxperi: @tianyif Any updates on this? Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -41,27 +41,27 @@
...
-  /** Audio capabilities when the device specifies external surround sound. */
-  @SuppressWarnings("InlinedApi")
-  private static final AudioCapabilities EXTERNAL_SURROUND_SOUND_CAPABILITIES =
-      new AudioCapabilities(
-          new int[] {
-            AudioFormat.ENCODING_PCM_16BIT, AudioFormat.ENCODING_AC3, AudioFormat.ENCODING_E_AC3
-          },
-          DEFAULT_MAX_CHANNEL_COUNT);
+  /** Encodings supported when the device specifies external surround sound. */
+  private static final int[] EXTERNAL_SURROUND_SOUND_CAPABILITIES =
```
tianyif, line 61: Since it is just an array of encodings, could we change it to `EXTERNAL_SURROUND_SOUND_ENCODINGS`?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -95,6 +96,19 @@ public static AudioCapabilities getCapabilities(Context context) {
     return getCapabilities(context, intent);
   }
 
+  private static AudioCapabilities getExternalSurroundCapabilities(@Nullable Intent intent) {
+    int[] supportedEncodings = EXTERNAL_SURROUND_SOUND_CAPABILITIES;
+
+    if (Util.SDK_INT >= 29) {
+      // Check if DTS Encodings are supported via Direct Playback.
+      int[] dtsTypes = {AudioFormat.ENCODING_DTS, AudioFormat.ENCODING_DTS_HD};
```
tianyif, line 104: nit: Better to rename `dtsTypes` to `dtsEncodings`?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -95,6 +96,19 @@ public static AudioCapabilities getCapabilities(Context context) {
...
+      supportedEncodings = Util.nullSafeIntegerArrayConcatenation(supportedEncodings,
+          Api29.getDirectPlaybackSupportedEncodings(dtsTypes));
+    }
+    supportedEncodings = Arrays.stream(supportedEncodings).distinct().toArray();
```
tianyif, line 108: I'm afraid that we currently cannot use Java 8 methods (`Arrays.stream(int[])`). And do we even need to de-duplication here?

libraries/common/src/main/java/androidx/media3/common/util/Util.java
tianyif: We got a hint from the audio framework team that the height channel masks were added in SDK 32 (e.g. [CHANNEL_OUT_TOP_BACK_LEFT](https://developer.android.com/reference/android/media/AudioFormat#CHANNEL_OUT_TOP_BACK_LEFT)), so we are now reaching to Dolby to see AudioTrack creation with 5.1.4 will also fail for them when API <= 31. If that's the case, the conflicts won't be there. But we are also waiting for a concrete answer from audio framework team as well.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -393,5 +410,20 @@ public static int getMaxSupportedChannelCountForPassthrough(
       }
       return 0;
     }
+
+    /**
+     * Returns an array list of surround encodings that maybe supported.
+     */
+    private static List<Integer> getAllSurroundEncodingsMaybeSupported() {
```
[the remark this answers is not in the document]
cedricxperi: @tianyif Done. Please see 8072d63acd4cbbe6666c68f5865bfb439fbeb23c Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif Done. Please see 8072d63acd4cbbe6666c68f5865bfb439fbeb23c Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif Done. Please see 8072d63acd4cbbe6666c68f5865bfb439fbeb23c Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: Added duplicated encodings check in one more place: 4637d5f832de88efc500274d3c2e7587a4f2c216. Thanks.

libraries/common/src/main/java/androidx/media3/common/util/Util.java
```diff
@@ -530,6 +574,23 @@ public static <T> void nullSafeListToArray(List<T> list, T[] array) {
...
 
+  /**
+   * Creates a new Integer array from a List of Integers.
+   *
+   * @param list The list of integers to convert.
+   * @return Created array of integers.
+   */
+  @UnstableApi
+  @SuppressWarnings("nullness:assignment")
+  public static int[] nullSafeIntegerListToArray(List<Integer> list) {
```
tianyif, line 585: Would it be better to use [Ints.toArray(Collection)](https://guava.dev/releases/19.0/api/docs/com/google/common/primitives/Ints.html#toArray(java.util.Collection))?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -393,5 +404,20 @@ public static int getMaxSupportedChannelCountForPassthrough(
       }
       return 0;
     }
+
+    /**
+     * Returns an array list of surround encodings that maybe supported.
+     */
+    private static ImmutableList<Integer> getAllSurroundEncodingsMaybeSupported() {
```
tianyif, line 411: Ok, I see we will convert the return type to an array whenever we use it. 

Probably we can return the array type `int[]` here, so at the end of this method we do 
```
return Ints.toArray(encodings.build());
```
instead of 
```
return encodings.build();
```
Sorry for the back and forth!

cedricxperi: @tianyif  No worries! please see d4f8b7f7ed1cae89d725dfbf4d74492d3808356a  Thanks

libraries/common/src/main/java/androidx/media3/common/util/Util.java
cedricxperi: @tianyif  Done. Please see d4f8b7f7ed1cae89d725dfbf4d74492d3808356a  Thanks

libraries/common/src/main/java/androidx/media3/common/util/Util.java
```diff
@@ -514,6 +514,50 @@ public static <T> T[] nullSafeArrayConcatenation(T[] first, T[] second) {
...
+  /**
+   * Creates a new Integer array containing the concatenation of two non-null Integer arrays.
+   *
+   * @param first  The first array.
+   * @param second The second array.
+   * @return The concatenated result.
+   */
+  @UnstableApi
+  @SuppressWarnings("nullness:assignment")
+  public static int[] nullSafeIntegerArrayConcatenation(int[] first, int[] second) {
```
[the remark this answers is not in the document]
cedricxperi: @tianyif  Done. Please see d4f8b7f7ed1cae89d725dfbf4d74492d3808356a  Thanks

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -102,25 +100,36 @@ public static AudioCapabilities getCapabilities(Context context) {
     if (Util.SDK_INT >= 23 && Api23.isBluetoothConnected(context)) {
       return DEFAULT_AUDIO_CAPABILITIES;
     }
+    int[] supportedEncodings = {};
```
tianyif, line 103: I have feeling that if we initialize this as a ImmutableSet.Builder, then add the encodings each time by:

```
// Below are all pseudo code

ImmutableSet.Builder supportedEncodings = new ImmutableSet.Builder<>();

supportedEncodings.addAll(Ints.asList(EXTERNAL_SURROUND_SOUND_ENCODINGS));
// Add other encodings

return AudioCapabilities(Ints.toArray(supportedEncodings.build()), DEFAULT_MAX_CHANNEL_COUNT);
```

In this way you don't need to add the de-duplication method. How do you think?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -354,9 +364,9 @@ private static final class Api29 {
     private Api29() {}
 
     @DoNotInline
-    public static int[] getDirectPlaybackSupportedEncodings() {
+    public static int[] getDirectPlaybackSupportedEncodings(int[] encodings) {
```
tianyif, line 367: I saw in the latest revision of this pull request, this method is only called at one place, where we pass the return array from `getAllSurroundEncodingsMaybeSupported`. How about we remove the parameter, and call `getAllSurroundEncodingsMaybeSupported` inside of this method?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: One of the  duplications comes from the pcm 16 bit encoding that exists in extra_surround_sound_encoding and the directplaybacksupported encodings. Will the above pseudo code solve this case?

tianyif: We will add all the encodings (whether it is from `extra_surround_sound_encoding` or `directplaybacksupported`) to the `ImmutableSet.Builder` via `addAll` method right? And the [javadoc](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableSet.Builder.html#addAll(java.lang.Iterable)) of the method has such explanation:

```
Adds each element of elements to the ImmutableSet, ignoring duplicate elements (only the first duplicate element is added).
```
So as long as we add everything into the `ImmutableSet.Builder` and the `ImmutableSet.Builder.build()` produces a set (every element is unique), then this should solve the duplication?

cedricxperi: @tianyif I see. Will try that tomorrow.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif  Done. Please see bc197dbd14499569708b0822bc1c9c016c6eabf3. Thanks

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -393,5 +403,20 @@ public static int getMaxSupportedChannelCountForPassthrough(
       }
       return 0;
     }
+
+    /**
```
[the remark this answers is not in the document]
cedricxperi: @tianyif  Done. Please see bc197dbd14499569708b0822bc1c9c016c6eabf3. Thanks

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif  Done. Please see bc197dbd14499569708b0822bc1c9c016c6eabf3. Thanks

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -102,25 +100,33 @@ public static AudioCapabilities getCapabilities(Context context) {
     if (Util.SDK_INT >= 23 && Api23.isBluetoothConnected(context)) {
       return DEFAULT_AUDIO_CAPABILITIES;
     }
+    ImmutableSet.Builder supportedEncodings = new ImmutableSet.Builder<>();
+
```
tianyif, line 104: nit: Remove the empty line 104 or swap the line 103 and 104, to make the construction of the ImmutableSet.Builder and its usage closer.

cedricxperi: @tianyif Done. 51b2201f3ae79c09f5b60b11a6bdf4361e00481a

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -102,25 +100,33 @@ public static AudioCapabilities getCapabilities(Context context) {
...
     // encodings.
     if (Util.SDK_INT >= 29 && (Util.isTv(context) || Util.isAutomotive(context))) {
-      return new AudioCapabilities(
-          Api29.getDirectPlaybackSupportedEncodings(), DEFAULT_MAX_CHANNEL_COUNT);
+      supportedEncodings.addAll(Ints.asList(Api29.getDirectPlaybackSupportedEncodings()));
     }
+
     if (intent == null || intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 0) {
-      return DEFAULT_AUDIO_CAPABILITIES;
+      if (supportedEncodings.build().isEmpty()) {
```
tianyif, line 118: 1. How do you think if we handle the case when HDMI is connected first, since currently the non-HDMI case has more nested logics. For example:

```
if (intent != null) && intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 1 {
    // HDMI case. Add the encodings from AudioManager.EXTRA_ENCODINGS, and return the audio capabilities based on the supportedEncodings and AudioManager.EXTRA_MAX_CHANNEL_COUNT
}

// Non HDMI case

```
2. For non-HDMI case, could we build the supportedEncodings set **just for once**, and check if it is empty, and then convert the **built set** to array to construct audio capabilities?

cedricxperi: @tianyif I think it is better to check getDirectPlaybackSupportedEncodings() before checking connected HDMI. I remember coming across cases where the device do not report ENCODING_DTS when it is connected to an AVR that supports it. Yet it is able to playback because of Direct passthrough playback. The SOC takes care of everything including transcoding and sending out the correct HDMI bitstream to the AVR. If we return immediately after checking the connected HDMI (as the first check), it will not know that ENCODING_DTS is actually supported by direct passthrough playback. It may revert to using mediacodec DTS decoder and then fail to transcode. The output could be PCM even though it is connected to an AVR. What do you think?

cedricxperi: @tianyif The way I am looking at this is:  getCapabilities() should return the union of all the encodings that are supported by direct playback and connected audio sink devices. Let me know if you think otherwise.

tianyif: Oh sorry for the confusion! I mean after all the other checks remain in their current place, and then just make the logics in the line 126 - 129 (HDMI case) **inside** of the if condition, and make the logics in the lines 118 - 123 **after** the if condition, that is:

```
// Bluetooth case, External surround sound settings case, direct playback case remain in their current place

if (intent != null) && intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 1 {
    // HDMI case. Add the encodings from AudioManager.EXTRA_ENCODINGS, and return the audio capabilities based on the supportedEncodings and AudioManager.EXTRA_MAX_CHANNEL_COUNT
}

// No HDMI case, the final return
```

cedricxperi: @tianyif No worries. Think I mis-interpreted your intentions as well. I agree, better to swap those 2 blocks of code. Please see df0058109857f35c36e4366c549ff1f26c43d04a  Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -102,25 +100,33 @@ public static AudioCapabilities getCapabilities(Context context) {
...
     }
     // AudioTrack.isDirectPlaybackSupported returns true for encodings that are supported for audio
     // offload, as well as for encodings we want to list for passthrough mode. Therefore we only use
     // it on TV and automotive devices, which generally shouldn't support audio offload for surround
     // encodings.
     if (Util.SDK_INT >= 29 && (Util.isTv(context) || Util.isAutomotive(context))) {
+      supportedEncodings.addAll(Ints.asList(Api29.getDirectPlaybackSupportedEncodings()));
+    }
+
+    if (intent == null || intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 1) {
```
tianyif, line 117: Should this be `if (intent != null && intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 1))`?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -102,25 +100,33 @@ public static AudioCapabilities getCapabilities(Context context) {
...
+      supportedEncodings.addAll(Ints.asList(
+          intent.getIntArrayExtra(AudioManager.EXTRA_ENCODINGS)));
       return new AudioCapabilities(
-          Api29.getDirectPlaybackSupportedEncodings(), DEFAULT_MAX_CHANNEL_COUNT);
+          Ints.toArray(supportedEncodings.build()), /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT);
```
tianyif, line 121: Should we pass `intent.getIntExtra(AudioManager.EXTRA_MAX_CHANNEL_COUNT, /* defaultValue= */ DEFAULT_MAX_CHANNEL_COUNT))` as the `maxChannelCount` in this case?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -102,25 +100,33 @@ public static AudioCapabilities getCapabilities(Context context) {
...
     }
-    if (intent == null || intent.getIntExtra(AudioManager.EXTRA_AUDIO_PLUG_STATE, 0) == 0) {
+
+    if (supportedEncodings.build().isEmpty()) {
```
tianyif, line 124: For lines 124 - 128, can we call the `supportedEncodings.build()` for just **once**?

```
ImmutableSet supportedEncodingsSet = supportedEncodings.build();
if (supportedEncodingsSet.isEmpty()) {
    return DEFAULT_AUDIO_CAPABILITIES;
} else {
    return new AudioCapabilities(Ints.toArray(supportedEncodingsSet), /* defaultValue= */
          DEFAULT_MAX_CHANNEL_COUNT);
}
```

I'm not sure if the naming `supportedEncodingsSet` is good though, we can get more suggestions in the internal review process.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif Oops. My bad. Fixed in 56ff091e9f58a25e57ecefbd49164df08773b46e  Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif This will caused problems with some Xiaomi DTS:X TV as it reports AudioManager.EXTRA_MAX_CHANNEL_COUNT as 8 instead of the correct value of 10 (For layout 5.1.4). This will cause DTS:X profile2 playback to fail.

cedricxperi: @tianyif I confirm that there are some DTS:X devices that are reporting  ACTION_HDMI_AUDIO_PLUG.EXTRA_MAX_CHANNEL_COUNT as 8 instead of the required 10 channels. 

If the AudioCapabilities is constructed with maxChannelCount==8, it will cause playback to stop in 
Method getEncodingAndChannelConfigForPassthrough()
...
channelCount = getMaxSupportedChannelCountForPassthrough(encoding, sampleRate);
    } else {
      channelCount = format.channelCount;
      **_if (channelCount > maxChannelCount) {_**

This is related to a similar issue about 2 weeks ago. I have unresolved that conversation so that you can see it. Since this is a device issue, how about we proceed with constructing AudioCapabilities with the maxChannlCount from the device as you have suggested. We then handle the channelCount=format.channelCount comparison as a special case and file a bug to come back to this when the devices are fixed (as suggested by @christosts in the earlier conversation that I just unresolved). What do you think? Please see recommended changes here: 78a0385b8d721d0a38947e7ffa0b616e44217983 Thanks.

P/S: I will file the DTS:X Device maxchannelcount reporting bug in Exoplayer and update the comment in the code once you ok this.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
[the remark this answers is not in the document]
cedricxperi: @tianyif This is the conversation that is related to the commit 78a0385b8d721d0a38947e7ffa0b616e44217983 Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
tianyif: Acknowledged. I will accept this pull request and move it into the internal review stage, during which our team will discuss how to address the mentioned concern.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
tianyif: No need to address this for now. We will do this change in the internal review stage.

tianyif:
Hi @cedricxperi,

As mentioned in the above comment, we are going to internally review this pull request. Please note that we are likely to do more changes based on the internal review feedbacks. Whenever the changes are needed, we will update them on Github and let you know. Could you please check it out then and help us to do the tests?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
cedricxperi: @tianyif Have filed an issue for this. See https://github.com/androidx/media/issues/396 Comments updated. Please see: 4973af30d3dde53bab5e2dd47b192914765eaa94

tianyif:
@cedricxperi I asked some questions inline in the code, and besides that, this revision has resolved all the internal review comments. Could you please test it on your side? I may do some rebase meanwhile though.

cedricxperi:
> @cedricxperi I asked some questions inline in the code, and besides that, this revision has resolved all the internal review comments. Could you please test it on your side? I may do some rebase meanwhile though.

@tianyif Thanks a lot. Will answer the questions and do the tests tomorrow my time (Asian time). Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
```diff
@@ -220,7 +234,13 @@ public Pair<Integer, Integer> getEncodingAndChannelConfigForPassthrough(Format f
       channelCount = getMaxSupportedChannelCountForPassthrough(encoding, sampleRate);
     } else {
       channelCount = format.channelCount;
-      if (channelCount > maxChannelCount) {
+      // Some DTS:X TVs reports ACTION_HDMI_AUDIO_PLUG.EXTRA_MAX_CHANNEL_COUNT as 8
+      // instead of 10. See https://github.com/androidx/media/issues/396
+      if (format.sampleMimeType.equals(MimeTypes.AUDIO_DTS_X)) {
```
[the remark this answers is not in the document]
cedricxperi: @tianyif I do not have access to other device at the moment but it is reasonable to think that this would be reproduceable for all DTS:X TVs based on MTK platforms. It will take sometime for this issue to be fixed on the platform. Meanwhile, Xperi will have to go to each and every streaming service to tell them to add the above so that DTS:X can be played back. I think it is better to add this patch in the exoplayer code base so that everyone will be aware of it.  When the issue get fixed on the platform (which could take a long time), we can remove this patch in another exoplayer contribution. Thanks.

cedricxperi:
> > @cedricxperi I asked some questions inline in the code, and besides that, this revision has resolved all the internal review comments. Could you please test it on your side? I may do some rebase meanwhile though.
> 
> @tianyif Thanks a lot. Will answer the questions and do the tests tomorrow my time (Asian time). Thanks.


@tianyif I have tested 1a77a927eab11f788431a14ab6a7368b5c1cd2d5 with various DTS test streams (speaker + bitstream out) and the playback is ok. Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java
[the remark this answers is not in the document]
tianyif: I will keep this patch. Thanks!

cedricxperi:
@tianyif I see this branch has some conflicts that need to be resovled. Let me know if you need me to resolve the conflicts. Thanks.

tianyif:
I already merged the pull request internally. And the change will be available next time new commits being pushed to `main` branch. No action item at this point. Thanks all for the contribution and discussion!

lowlatencystreamer:
> @tianyif Is it ok if I delete my forked androidx/media repository for this contribution? I want to branch off from "main" instead of "release" for the next pull request. But am not able to create another fork of androidx/media until I delete the current fork that I have. Thanks.

@tianyif I found out how to copy the main branch into this fork. So I am good. Thanks. Please ignore this message.

PR #348, opened 2023-04-21
extract mp3 bitrate from synchronized header

chientrm:
This helps determining bitrate of the first frame synchronized.


icbaker:
As explained in https://github.com/google/ExoPlayer/issues/11124#issuecomment-1516490095 we won't accept this into the library, because it assumes **all** MP3 files are CBR, which is not a valid assumption. Propagating the bitrate in a way that either works for both VBR and CBR, or **only** affects truly CBR files is much harder as described in that comment.

My suggestion in https://github.com/google/ExoPlayer/issues/11124#issuecomment-1516490095 is that if **you** want to make this assumption you can make (and keep) this change locally.

PR #368, opened 2023-05-01
Strings must be escaped in raw SQL queries.

Tolriq:
Without this fix, using "'" in the cacheKey will break cache and allow SQL injections.

### DIFF
--- a/libraries/datasource/src/main/java/androidx/media3/datasource/cache/CacheFileMetadataIndex.java
+++ b/libraries/datasource/src/main/java/androidx/media3/datasource/cache/CacheFileMetadataIndex.java
@@ -19,6 +19,7 @@ import static androidx.media3.common.util.Assertions.checkNotNull;
 
 import android.content.ContentValues;
 import android.database.Cursor;
+import android.database.DatabaseUtils;
 import android.database.SQLException;
 import android.database.sqlite.SQLiteDatabase;
 import androidx.annotation.WorkerThread;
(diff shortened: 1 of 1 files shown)


marcbaechinger:
I'm unsure if we need that. Have you tried this out and can you provide a test fixture of this being able to inject SQL that leads to something else than deleting the table? That would be an exploit of a bug in the `SqLiteDatabase` implementation. That's possible but without such a test fixture I'd say we are fine as per documentation and as per my source code investigation (see below).

I'm asking because I think that when using the `whereArgs` arguments we are save. It's actually the idea of these methods to be save as opposed to concatenating SQL queries by hand.

When reading the [documentation](https://developer.android.com/topic/security/risks/sql-injection#use_query_methods) it says it's save to use query methods like the `delete` method you patched. 

It also says it's fine to use prepared statements and when I follow the code of the `delete` method you patched, I end up here: https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/database/sqlite/SQLiteConnection.java?pli=1#1159  Which explains why query methods are save in case `PreparedStatements` are save as [documented here](https://developer.android.com/topic/security/risks/sql-injection#use_preparedstatement_objects).

Tolriq:
Hum yes you're right, this was a coincidence and still trying to figure out all the cases and recovery from https://github.com/google/ExoPlayer/issues/10773.

Seems I reached a state where `cache.removeResource(cacheKey)` did not work out and ran the `SimpleCache.delete` but seems that one can't be called when the cache is loaded.

1) What would be the proper way to try to debug those cache corruptions?
2) Is there a safe way to clear the cache and recover without closing and restarting the cache? (Since it's tied to the service, this requires a major rewrite to be able to unload it before clean)

microkatz:
Hi @Tolriq.

Thank you for your continued persistence! I'm going to close this Github pull request as it seems you decided to rescind it? Reopen if that is not the case.

Would you be able to either repost your questions in https://github.com/google/ExoPlayer/issues/10773 or create a new Github issue with these questions?

PR #369, opened 2023-05-01
Catch errors and OOM when decoding ID3 frames.

Tolriq:
Invalid frames have no impact on ExoPlayer ability to play the media and should not fail on errors. 

Some tools can add 100Mb images in the tags that will trigger recoverable OOM with this fix, previously the load would fail and the OOM caught too late.

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/metadata/id3/Id3Decoder.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/metadata/id3/Id3Decoder.java
@@ -372,8 +372,9 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
       frameSize = removeUnsynchronization(id3Data, frameSize);
     }
 
+    String error = "";
+    Id3Frame frame = null;
     try {
-      Id3Frame frame;
       if (frameId0 == 'T'
           && frameId1 == 'X'
           && frameId2 == 'X'
@@ -430,18 +431,24 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
         String id = getFrameId(majorVersion, frameId0, frameId1, frameId2, frameId3);
         frame = decodeBinaryFrame(id3Data, frameSize, id);
       }
-      if (frame == null) {
-        Log.w(
-            TAG,
-            "Failed to decode frame: id="
-                + getFrameId(majorVersion, frameId0, frameId1, frameId2, frameId3)
-                + ", frameSize="
-                + frameSize);
-      }
-      return frame;
-    } finally {
+    } catch (Exception e) {
+      error = ",error=" + e.getMessage();
+    } catch (OutOfMemoryError e) {
+      error = ",error=" + e.getMessage();
+    }
+    finally {
       id3Data.setPosition(nextFramePosition);
     }
+    if (frame == null) {
+      Log.w(
+          TAG,
+          "Failed to decode frame: id="
+              + getFrameId(majorVersion, frameId0, frameId1, frameId2, frameId3)
+              + ", frameSize="
+              + frameSize
+              + error);
+    }
+    return frame;
   }
 
   @Nullable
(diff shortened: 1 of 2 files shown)


libraries/extractor/src/main/java/androidx/media3/extractor/metadata/id3/Id3Decoder.java
```diff
@@ -428,18 +429,24 @@ private static Id3Frame decodeFrame(
...
-        Log.w(
-            TAG,
-            "Failed to decode frame: id="
-                + getFrameId(majorVersion, frameId0, frameId1, frameId2, frameId3)
-                + ", frameSize="
-                + frameSize);
-      }
-      return frame;
-    } finally {
+    } catch (Exception e) {
```
icbaker, line 432: What happens if this `catch Exception` isn't present (i.e. we only catch `OutOfMemoryError` here)?

icbaker (commented):
Can you please provide a test asset in this PR with a problematically large ID3 tag that reproduces the failure this PR is fixing? You can probably do this by either:
* Adding a test case to `Id3DecoderTest` with the bytes written out in java (probably only if it's very short, which seems like it might be unlikely since we're testing an OOM :p)
* Or copying and modifying one of the [existing assets](https://github.com/androidx/media/tree/release/libraries/test_data/src/test/assets/media/id3) and using that in a new test case in `Id3DecoderTest` (see how they're currently used in `Id3PeekerTest`).

libraries/extractor/src/main/java/androidx/media3/extractor/metadata/id3/Id3Decoder.java
Tolriq: For this specific case the Exception catch is not necessary, I have some analytics logs about errors but no reproduction file provided by any users.

Tolriq:
I can provide a test asset in private but it's probably copyrighted including the image. (Got the file from an user). That file crash most of the id3 tools I have so can't even extract the tags. (The other tools refuse to read them)  So not sure how to be able to build one similar.

Tolriq:
Bump on this one, I'm not able to generate the test asset with the tools I have at my disposal. I can still send the copyrighted file by mail if anyone is able to extract the tags and reproduce in a non copyrighted way.

If you can generate and send me the test asset then I can finish this PR.

libraries/extractor/src/main/java/androidx/media3/extractor/metadata/id3/Id3Decoder.java
tonihei: This part if probably fine given we return null in other cases where we specifically detect that the frame is malformed.

libraries/extractor/src/main/java/androidx/media3/extractor/metadata/id3/Id3Decoder.java
```diff
@@ -428,18 +429,24 @@ private static Id3Frame decodeFrame(
...
+      error = ",error=" + e.getMessage();
+    } catch (OutOfMemoryError e) {
```
tonihei, line 434: Can you combine both catch clauses together? (as `Exception | OutOfMemoryError`)
This would allow to move the `Log.w` line and `return null` inside the catch clause and make the code a lot cleaner.

tonihei:
After some further discussion we said it's probably alright not to have a test file for this case. We have other OOM error clauses that guard against invalid media without a unit test, so this doesn't feel a lot different.

libraries/extractor/src/main/java/androidx/media3/extractor/metadata/id3/Id3Decoder.java
Tolriq: @tonihei The Log.w was split to help debug other issues, `decodeTxxxFrame` and others can return null and we have no log or information to help debug things when users complains about missing data.

If possible I'd like to keep it here. If you do not want the error field I can add 2 different log calls.

icbaker: I've kept the logging for `frame == null` but improved the logs to include the full exception stack trace.

icbaker:

PR #420, opened 2023-05-25
Save state throwsWhenUsingWrongThread when ListenerSet::copy()

changxiangzhong:
Our code
```
DefaultAnalyticsCollector collector = new DefaultAnalyticsCollector();
collector.setThrowsWhenUsingWrongThread(false);

// This function call ListenerSet.copy(), which ignores the state set in the above line
collector.setPlayer(myPlayer, myLooper);
```

The implementation of DefaultAnalyticsCollector.setPlayer(), which calls `ListenerSet.copy()`
image

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/util/ListenerSet.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/ListenerSet.java
@@ -99,7 +99,12 @@ public final class ListenerSet<T extends @NonNull Object> {
    *     during one {@link Looper} message queue iteration were handled by the listeners.
    */
   public ListenerSet(Looper looper, Clock clock, IterationFinishedEvent<T> iterationFinishedEvent) {
-    this(/* listeners= */ new CopyOnWriteArraySet<>(), looper, clock, iterationFinishedEvent, true);
+    this(
+        /* listeners= */ new CopyOnWriteArraySet<>(),
+        looper,
+        clock,
+        iterationFinishedEvent,
+        /* throwsWhenUsingWrongThread= */ true);
   }
 
   private ListenerSet(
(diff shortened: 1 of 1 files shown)


icbaker:
Thanks for the contribution! I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks!

tonihei:
Just following up on my question in https://github.com/google/ExoPlayer/pull/11181: I can now see how you set the flag in `DefaultAnalyticsCollector`. But I still don't understand how you accidentally use the `DefaultAnalyticsCollector` on a wrong thread. If it's used with an `ExoPlayer` instance, the correct threading should be enforced already by the player integration (and it's not possible to disable these checks). Do you have a stack trace or reproduction steps for a problem when `ThrowsWhenUsingWrongThread` is kept at its default of `true`?

In case you manually trigger some methods on this class, it's also worth highlighting that the implementation is not thread-safe and using it from multiple threads may result in hard-to-debug problems in the underlying listener invocation code.

changxiangzhong:
Thank you for your reminder in paragraph #2. We'll try to review the thread safety issue.

We are using the delegate pattern to implement our own CorePlayer, like the following snippet. 
image

The suspect stack trace is as following, it's running in a background thread, which cause the exception.
```
BasePlayer.addAnalyticsListener(AnalyticsListener) (se.b17g.player.player.base)
    |->CorePluginHandler.onInit(Ticket) (se.b17g.player.plugins)
	    |->PlaybackApiPlayer.onApiLoaded(PlaybackConfig) (se.b17g.player.player)
```

tonihei:
Ah, thanks for the explanation! 

It's definitely best to avoid the threading issue if at all possible (e.g. by posting updates to the right thread instead of triggering them directly). Some further notes for these specific methods:
 - We deliberately added the check to `remove(AnalyticsListener)` because the `AnalyticsListener.onEvents` callbacks might be called on the wrong thread as a result of the `remove` call. This may surprising or potentially dangerous for listener implementations who expect all listener callbacks to happen on the specified thread.
 - `add(AnalyticsListener)` is the only exception to the threading rule because it's inherently thread-safe as we implemented it and it helps with setup flows where the listener is added on another thread when a player is created (e.g. when using dependency injection frameworks like Dagger).

PR #426, opened 2023-05-30
Fix DTS:X P2 Direct Passthrough Playback.

cedricxperi:
DTS:X P2 audio will currently fail in direct passthrough playback. 

It will fail at DefaultAudioOffloadSupportProvide.java,  around line 96
_**AudioFormat audioFormat = Util.getAudioFormat(format.sampleRate, channelConfig, encoding);**_ when trying to build an AudioFormat with encoding == ENCODING_DTS_UHD_P2.
 
 AudioFormat.ENCODING_DTS_UHD_P2 is only defined from API 34 onwards. Before that encoding should
 be set to C.ENCODING_DTS, which is supported by TV ICs for direct passthrough playback. 
@tianyif

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioOffloadSupportProvider.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioOffloadSupportProvider.java
@@ -81,17 +81,16 @@ public final class DefaultAudioOffloadSupportProvider
 
     @C.Encoding
     int encoding = MimeTypes.getEncoding(checkNotNull(format.sampleMimeType), format.codecs);
-    if (encoding == C.ENCODING_INVALID) {
+    // AudioFormat.ENCODING_DTS_UHD_P2 is defined from API 34 onwards. We return offload
+    // unsupported to prevent crash in Util.getAudioFormat() below when it tries to create
+    // an AudioFormat with ENCODING_DTS_UHD_P2.
+    if ((encoding == C.ENCODING_INVALID) ||
+        ((Util.SDK_INT < 34) && (encoding == C.ENCODING_DTS_UHD_P2))) {
       return AudioOffloadSupport.DEFAULT_UNSUPPORTED;
     }
     int channelConfig = Util.getAudioTrackChannelConfig(format.channelCount);
     if (channelConfig == AudioFormat.CHANNEL_INVALID) {
       return AudioOffloadSupport.DEFAULT_UNSUPPORTED;
-    }
-    // AudioFormat.ENCODING_DTS_UHD_P2 is defined from API 34 onwards. Before that encoding should
-    // be set to C.ENCODING_DTS, which is supported by TV ICs for offload playback.
-    if ((Util.SDK_INT < 34) && (encoding == C.ENCODING_DTS_UHD_P2)) {
-      encoding = C.ENCODING_DTS;
     }
     AudioFormat audioFormat = Util.getAudioFormat(format.sampleRate, channelConfig, encoding);
     if (Util.SDK_INT >= 31) {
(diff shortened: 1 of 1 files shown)


microkatz:
Hello @cedricxperi,

Thank you for proposing a code change to ExoPlayer!

I understand that its the building of an AudioFormat object with the unreleased format is crashing. However, I believe that your suggested changes just punt the problem down the line.

Your proposed change will cause the following behavior. For SDK < 34, offload support for `ENCODING_DTS_UHD_P2` will be checked by checking offload support for `ENCODING_DTS`. If offload is supported for `ENCODING_DTS` and the renderers are configured for offload, then `DefaultAudioSink::configure` is called with Format with `ENCODING_DTS_UHD_P2`. 

Unlike in `::configure` with passthrough, no method is called that sets the configuration's output encoding to `ENCODING_DTS`. In `::createAudioTrackV29` we'll call `Util.getAudioFormat` again and that will crash with the same error you are facing now.

Have you tested this code change? Do you have a device that supports `ENCODING_DTS` with offload?

cedricxperi:
> 

Hi @microkatz 

Thanks for the quick response. Appreciate it.

- Perhaps I have should have changed the title to "Fix DTS:X P2 Direct Passthrough playback" instead of calling it offload. What is really happening here is the creation of AudioFormat with an unsupported type (ENCODING_DTS_UHD_P2) during the checking of offload support is causing the DTS:X P2 direct passthrough playback to stop abruptly. Ultimately, the playback will be through direct passthrough. For playing back DTS audio on TVs, MTK has recommended that Direct Passthrough playback should always be used so that the application do not have to worry about things like transcoding.
- I do not have a device that supports offload playback for DTS audio. But I do have a Xiaomi v435H TV that supports Tunnel mode with Direct passthrough playback for audio and that is what we are testing and is a focus for DTS as far as DTS:X Profile 2 audio is concerned.
- At present, there is no expectation for offload playback for DTS:X P2 on any existing device. For futures devices with API 34 and above, this would not be an issue as AudioFormat.ENCODING_DTS_UHD_P2 will be defined. How about we just returned offload unsupported when it is DTS:X P2 and API < 34? This will allow Direct Passthrough playback to proceed and not affect existing offload for ENCODING_DTS. For your consideration.
  
   if ((Util.SDK_INT < 34) && (encoding == C.ENCODING_DTS_UHD_P2)) {
      return AudioOffloadSupport.DEFAULT_UNSUPPORTED;
    }

microkatz:
Gotcha. I understand your reasoning and that seems more acceptable. Would you also move it to right below the check on encoding == C.ENCODING_INVALID, about line 86? Thanks! And then I'll review.

cedricxperi:
@microkatz Please see 17e47f78f61f3b2bb00d4f7133a724e59b442bb4   Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioOffloadSupportProvider.java
```diff
@@ -81,14 +81,17 @@ public AudioOffloadSupport getAudioOffloadSupport(
 
     @C.Encoding
     int encoding = MimeTypes.getEncoding(checkNotNull(format.sampleMimeType), format.codecs);
-    if (encoding == C.ENCODING_INVALID) {
+    // AudioFormat.ENCODING_DTS_UHD_P2 is defined from API 34 onwards. We return offload
+    // unsupported to prevent crash in Util.getAudioFormat() below when it tries to create
+    // an AudioFormat with ENCODING_DTS_UHD_P2.
+    if ((encoding == C.ENCODING_INVALID) ||
+        ((Util.SDK_INT < 34) && (encoding == C.ENCODING_DTS_UHD_P2))) {
```
microkatz, line 88: I understand that that both conditions would lead to returning AudioOffloadSupport.DEFAULT_UNSUPPORTED. However, since this is a specific workaround with adjoining comment, could you split into two conditionals?

cedricxperi: @microkatz Done. Thanks.

microkatz:

PR #441, opened 2023-06-06
Dont cache in OkHttpDataSource

yschimke:
ExoPlayer has it's own cache, so OkHttp caching should be disabled to avoid storing twice.

### DIFF
--- a/libraries/datasource_okhttp/src/main/java/androidx/media3/datasource/okhttp/OkHttpDataSource.java
+++ b/libraries/datasource_okhttp/src/main/java/androidx/media3/datasource/okhttp/OkHttpDataSource.java
@@ -198,6 +198,9 @@ public class OkHttpDataSource extends BaseDataSource implements HttpDataSource {
   private long bytesToRead;
   private long bytesRead;
 
+  private static final CacheControl DISABLE_CACHE = new CacheControl.Builder().noCache().noStore()
+      .build();
+
   /**
    * @deprecated Use {@link OkHttpDataSource.Factory} instead.
    */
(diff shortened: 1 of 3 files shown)


libraries/datasource_okhttp/src/main/java/androidx/media3/datasource/okhttp/OkHttpDataSource.java
```diff
@@ -458,6 +461,10 @@ private Request makeRequest(DataSpec dataSpec) throws HttpDataSourceException {
       requestBody = RequestBody.create(null, Util.EMPTY_BYTE_ARRAY);
     }
     builder.method(dataSpec.getHttpMethodString(), requestBody);
+
+    // ExoPlayer implements it's own caching, so avoid a second cache.
```
oceanjules, line 465: Could you fix the typo to "its own caching"?

marcbaechinger:
Thanks!

icbaker:
CC @tonihei 

> ExoPlayer has it's own cache

Is this referring to `CacheDataSource`? Usage of that is optional, so I'm not sure we should disable OkHttp caching for **all** users of `OkHttpDataSource` - what about exposing this as an option on `OkHttpDataSource`? In fact, going to look at what that would look like, i think it [already exists](https://developer.android.com/reference/androidx/media3/datasource/okhttp/OkHttpDataSource.Factory#setCacheControl(okhttp3.CacheControl)) - and this PR as implemented actually breaks that setter (because it unconditionally overwrites `request.setCacheControl` lower down in the `makeRequest` method).

yschimke:
@icbaker ooops, I'll take a look again.  Or close it. Thanks.

yschimke:
Closing in favour of https://github.com/google/horologist/pull/1364/files

Still not sure how I missed that, but now painfully obvious.

I would suggest changing the default, but suspect this would be considered a breaking change for you.

But currently I suspect some number of apps will be double caching.

PR #475, opened 2023-06-24
Support cast live seekable ranges and seek to the live position.

dfoelster:
Currently the seekable range of live streams is not exposed in the CastPlayer.
The duration is always reported as C.TIME_UNSET.

This PR checks the media status of the active item for a LiveSeekableRange and exposes it over the `CastTimeline`.  
Because the Timeline is only updated on status changed callbacks from the cast sdk (which are not send for livestream range changes) it would be outdated immediately.
To avoid this issue the `CastTimeline` stores it's creation timestamp and adjusts the reported window according to the elapsed time. This is similar to the internal behavior of the `RemoteMediaClient`.

### DIFF
--- a/libraries/cast/src/main/java/androidx/media3/cast/CastPlayer.java
+++ b/libraries/cast/src/main/java/androidx/media3/cast/CastPlayer.java
@@ -425,15 +425,21 @@ public final class CastPlayer extends BasePlayer {
     // We assume the default position is 0. There is no support for seeking to the default position
     // in RemoteMediaClient.
     positionMs = positionMs != C.TIME_UNSET ? positionMs : 0;
-    if (mediaStatus != null) {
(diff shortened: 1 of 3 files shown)


microkatz:
@dfoelster 

Hi you'll need to sign the CLA agreement before we can review your pull request. Thank you!

albinus-sc:
@microkatz @dfoelster Hey, this pull request seems like it might solve some of the issues we are facing regarding casting live content with a buffer. Any updates on this?

albinus-sc:
@microkatz I'm afraid I don't understand the changes well enough to create my own PR, let alone create tests for it.

I just wanted to let you know that this is a problem we're struggling with and if you at some point have the time to look at it yourself it would be very much appreciated!

Issue #485, opened 2023-06-28
Make adaptation set ID a double


vishnuchilakala:
### [REQUIRED] Use case description
There are some contents like [this](https://virtual-channel.unified-streaming.com/demo_channel-stable.isml/.mpd) that have adaptation set IDs set to a value bigger than integer, which is making Exoplayer unable to play these contents.
This is different from Shaka player, which has id set to `number` and thus does not throw any issues.

### Proposed solution
Change id of `AdaptationSet` to `java.lang.Double`.

vishnuchilakala:
Raised a PR for the proposed change: https://github.com/androidx/media/pull/486

tonihei:
The DASH spec (ISO 23009-1) defines the schema of `AdaptationSet` and has `<xs:attribute name="id" type="xs:unsignedInt"/>`, which looks to me as if it must always be an unsigned integer. 

> There are some contents like [this](https://virtual-channel.unified-streaming.com/demo_channel-stable.isml/.mpd) that have adaptation set IDs set to a value bigger than integer

Do you have control over this content? If so, I'd recommend to fix the ids to fit the DASH spec schema. Alternatively, please try to contact the owners of the content to remedy the issue.

vishnuchilakala:
@tonihei As far as I understand, it can be up to 64 bits, which Javascript's `number` does support and hence Shaka player does not have a problem. But Java's `Integer` can hold only upto 32 bits, which is causing this issue.

Can you please confirm this?

tonihei:
The XML schema definition for `xs:unsignedInt` is [relatively clear](https://www.oreilly.com/library/view/xml-schema/0596002521/re97.html) that this is 32-bit. There is a small issue still that we currently use `Integer.parseInt` although we should technically use `Integer.parseUnsignedInt` to cover the additional range by the one extra bit we get for unsigned integers. I can change this (unless you prefer to send a PR for this?) to be fully spec-compliant, but I don't think we should allow values other than what the spec allows.

vishnuchilakala:
@tonihei that makes sense. I have used that and tested the said content and it works without any issues, so that should be good enough. I have created a new PR with that change: https://github.com/androidx/media/pull/487
Please let me know what you think. Thanks for the quick reply!

tonihei:
Ah, interesting. So your test content is actually spec-compliant because the `id` is the range of `2^31` to `2^32-1`?

Will try to merge the PR now, thanks!

tonihei:
[...]

PR #486, opened 2023-06-28
Changed type of id of adaptation set to double

vishnuchilakala:
Fixes https://github.com/androidx/media/issues/485


tonihei:
Closing (see discussion in https://github.com/androidx/media/issues/485)

PR #487, opened 2023-06-28
Allow unsigned integers in adaptation set id

### LINKED ISSUE #485: Make adaptation set ID a double

vishnuchilakala:
Fixes #485

### DIFF
--- a/libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/DashMediaPeriod.java
+++ b/libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/DashMediaPeriod.java
@@ -20,7 +20,6 @@ import static java.lang.annotation.ElementType.TYPE_USE;
 
 import android.util.Pair;
 import android.util.SparseArray;
-import android.util.SparseIntArray;
 import androidx.annotation.IntDef;
 import androidx.annotation.Nullable;
 import androidx.media3.common.C;
@@ -58,6 +57,7 @@ import androidx.media3.exoplayer.upstream.Allocator;
 import androidx.media3.exoplayer.upstream.CmcdConfiguration;
 import androidx.media3.exoplayer.upstream.LoadErrorHandlingPolicy;
 import androidx.media3.exoplayer.upstream.LoaderErrorThrower;
+import com.google.common.collect.Maps;
 import com.google.common.primitives.Ints;
 import java.io.IOException;
 import java.lang.annotation.Documented;
@@ -66,6 +66,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.regex.Matcher;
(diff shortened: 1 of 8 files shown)


tonihei:
After some further discussion during internal review, we switched the parsing to `long` after all (that is, full 64-bit integer). Please have a look at the additional changes I uploaded to this PR. This helps to avoid the negative parsed numbers, which may be confusing. Also, `Integer.parseUnsignedInt` turned out not to be available on all Android versions we support.

vishnuchilakala:
@tonihei thanks for making the changes. looks good to me 👍

vishnuchilakala:
@tonihei are we good with these changes? If you need any further information from my side, please do let me know.

tonihei:
Yes, it's all merged in our internal codebase and you'll see it merged here too the next time we update our changes (~once a week).

vishnuchilakala:
@tonihei Please help us understand the Media3 vs Exoplayer situation.

We are still on the old Exoplayer and the migration is not planned any time soon because of some other dependencies. So how would we proceed here?

Would all the changes that are being contributed to Media3 also be ported to Exoplayer, and vice versa? How long would the team continue to supporting the old Exoplayer? Is there any difference in the timelines that the changes would be released, between the two? Would appreciate any information that you can share with us to handle this situation. Thanks!

tonihei:
The code base is 100% identical and we push the same code to both repositories (just that we rewrite package names for ExoPlayer in the process).

 > We are still on the old Exoplayer and the migration is not planned any time soon because of some other dependencies.

I'm curious about this part. What kind of dependency prevents you from migrating? The migration step for your code should be very straight-forward with the [migration script](https://developer.android.com/guide/topics/media/media3/getting-started/migration-guide#exoplayer) that does everything for you. As pointed out above, the code is fully identical, so this is really just about renaming package imports.

> the timelines that the changes would be released,

Given this is a bug fix, it will be part of any Media3 1.1.1 release and ExoPlayer 2.19.1  (which will be identical again). There is no guarantee we actually do such a bug fix release, but it's not too unlikely because there will always be a few things worth fixing in a release. 

We will also continue to push changes to both repositories for a while, but we are not planning to release ExoPlayer 2.20 (matching Media3 1.2.0).

vishnuchilakala:
> What kind of dependency prevents you from migrating?

We depend on some third parties for their proprietary analytics and SSAI, who we work with using Exoplayer components. They are not planning to migrate yet, effectively blocking us too from doing it.

> it will be part of any Media3 1.1.1 release and ExoPlayer 2.19.1

I just saw that 2.19.0 has been released and everything has been marked as deprecated. So can you please confirm that we will get another minor release that will have this change?

> We will also continue to push changes to both repositories for a while

We are trying to understand your timelines so that we can bring those third parties to the table and plan the migrations accordingly. Would it be fair to assume that any fixes / enhancements you make for **another month** will be released on the old Exoplayer also?

Appreciate the help, thanks!

tonihei:
> They are not planning to migrate yet, effectively blocking us too from doing it.

That's interesting. But I guess it's the same as them not updating from ExoPlayer version 2.x to 2.y, meaning you are blocked on 2.x? 

> So can you please confirm that we will get another minor release that will have this change?

I can't really confirm it, but we usually wait a bit to see what kind of bug fixes we accumulate and make a new bug fix release if needed. We've almost always done one in the past, so I'm fairly sure we end up doing Media3 1.1.1. and ExoPlayer 2.19.1 too. Your change isn't included in the main 2.19.0 yet because it happened after the internal branch cut off.

> Would it be fair to assume that any fixes / enhancements you make for another month will be released on the old Exoplayer also?

We just push to the `dev-v2` branch. We are not planning any further ExoPlayer releases except for `2.19.x` bugfix releases (which are all aligned with the corresponding Media3 `1.1.x` bugfix release).

vishnuchilakala:
> But I guess it's the same as them not updating from ExoPlayer version 2.x to 2.y, meaning you are blocked on 2.x?

Yep, that's basically it.

Thanks a lot for the information @tonihei ! This is really helpful for us to plan our roadmap.

Banks666 (changes requested):
Dd

PR #491, opened 2023-06-29
Add luma & chroma bit depths to Format

dsparano:
**Issue**
Videos may be encoded with different bit depths (even between luma and chroma), not just the traditional 8 bpp.

Currently the Format class carries Color metadata, but not bit depths. The capability to decode wide bit depth videos is down to the codec profile support from the device but not always the video profile is associated with a unique bit depth (for ex. not for H266 where Main 10 is for both 8 and 10 bpp). Besides working out the bitdepth from the codec specific profile strings would be cumbersome.

We think it would be useful for e.g. renderers, transformers etc to know the video bit depth. We had this issue internally while working on use cases such as 8bpp base and 10bpp LCEVC enhanced output, but we think it could very well benefit other scenarios.

**Proposed solution**
Add luma and chroma bit depths to the Format class, set them from the various parsers (see changes).

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/ColorInfo.java
+++ b/libraries/common/src/main/java/androidx/media3/common/ColorInfo.java
@@ -288,10 +288,27 @@ public final class ColorInfo implements Bundleable {
   /**
    * Returns whether this instance is valid.
    *
-   * <p>This instance is valid if no color members are {@link Format#NO_VALUE}, while bit depths may
-   * be unset.
+   * <p>This instance is valid if at least one between bitdepths and color info are valid.
    */
   public boolean isValid() {
+    return isBitdepthValid() || isColorValid();
+  }
+
+  /**
+   * Returns whether this instance has valid bitdepths.
+   *
+   * <p>This instance has valid bitdepths if none of them is {@link Format#NO_VALUE}.
+   */
+  public boolean isBitdepthValid() {
+    return lumaBitdepth != Format.NO_VALUE && chromaBitdepth != Format.NO_VALUE;
+  }
+
+  /**
+   * Returns whether this instance is color valid.
+   *
+   * <p>This instance is valid if no color members are {@link Format#NO_VALUE}.
+   */
+  public boolean isColorValid() {
     return colorSpace != Format.NO_VALUE
         && colorRange != Format.NO_VALUE
         && colorTransfer != Format.NO_VALUE;
@@ -302,29 +319,17 @@ public final class ColorInfo implements Bundleable {
    *
    * @see Format#toLogString(Format)
    */
-  public String toColorString() {
+  public String toLogString() {
     if (!isValid()) {
       return "NA";
     }
 
-    return Util.formatInvariant(
-        "%s/%s/%s",
+    String bitdepthsString = isBitdepthValid() ? lumaBitdepth + "/" + chromaBitdepth : "NA";
+    String colorString = isColorValid() ? Util.formatInvariant("%s/%s/%s",
         colorSpaceToString(colorSpace),
         colorRangeToString(colorRange),
-        colorTransferToString(colorTransfer));
-  }
-
-  /**
-   * Returns whether this instance has valid bitdepths.
-   *
-   * <p>This instance has valid bitdepths if none of them is {@link Format#NO_VALUE}.
-   */
-  public boolean isBppValid() {
-    return lumaBitdepth != Format.NO_VALUE && chromaBitdepth != Format.NO_VALUE;
-  }
-
-  public String toBppString() {
-    return isBppValid() ? lumaBitdepth + "," + chromaBitdepth : "NA";
+        colorTransferToString(colorTransfer)) : "NA";
+    return bitdepthsString + "/" + colorString;
   }
 
   @Override
@@ -347,9 +352,9 @@ public final class ColorInfo implements Bundleable {
   @Override
   public String toString() {
     return "ColorInfo("
-        + lumaBitdepth
+        + lumaBitdepthToString(lumaBitdepth)
         + ", "
-        + chromaBitdepth
+        + chromaBitdepthToString(chromaBitdepth)
         + ", "
         + colorSpaceToString(colorSpace)
         + ", "
@@ -359,6 +364,14 @@ public final class ColorInfo implements Bundleable {
         + ", "
         + (hdrStaticInfo != null)
         + ")";
+  }
+
+  private static String lumaBitdepthToString(int val) {
+    return val != Format.NO_VALUE ? val + "bit Luma" : "NA";
+  }
+
+  private static String chromaBitdepthToString(int val) {
+    return val != Format.NO_VALUE ? val + "bit Chroma" : "NA";
   }
 
   private static String colorSpaceToString(@C.ColorSpace int colorSpace) {
--- a/libraries/common/src/main/java/androidx/media3/common/ColorInfo.java
+++ b/libraries/common/src/main/java/androidx/media3/common/ColorInfo.java
@@ -22,6 +22,9 @@ import androidx.media3.common.util.Util;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.util.Arrays;
 import org.checkerframework.dataflow.qual.Pure;
+// copybara:exo-only import com.google.android.exoplayer2.Bundleable;
+// copybara:exo-only import com.google.android.exoplayer2.C;
+// copybara:exo-only import com.google.android.exoplayer2.Format;
 
 /**
  * Stores color info.
@@ -320,11 +323,6 @@ public final class ColorInfo implements Bundleable {
    * @see Format#toLogString(Format)
    */
   public String toLogString() {
-    if (!isValid()) {
-      return "NA";
-    }
-
-    String bitdepthsString = isBitdepthValid() ? lumaBitdepth + "/" + chromaBitdepth : "NA";
     String dataspaceString =
         isDataSpaceValid()
             ? Util.formatInvariant(
@@ -332,8 +330,9 @@ public final class ColorInfo implements Bundleable {
                 colorSpaceToString(colorSpace),
                 colorRangeToString(colorRange),
                 colorTransferToString(colorTransfer))
-            : "NA";
-    return bitdepthsString + "/" + dataspaceString;
+            : "NA/NA/NA";
+    String bitdepthsString = isBitdepthValid() ? lumaBitdepth + "/" + chromaBitdepth : "NA/NA";
+    return dataspaceString + "/" + bitdepthsString;
   }
 
   @Override
@@ -354,12 +353,23 @@ public final class ColorInfo implements Bundleable {
   }
 
   @Override
+  public int hashCode() {
+    if (hashCode == 0) {
+      int result = 17;
+      result = 31 * result + colorSpace;
+      result = 31 * result + colorRange;
+      result = 31 * result + colorTransfer;
+      result = 31 * result + Arrays.hashCode(hdrStaticInfo);
+      result = 31 * result + lumaBitdepth;
+      result = 31 * result + chromaBitdepth;
+      hashCode = result;
+    }
+    return hashCode;
+  }
+
+  @Override
   public String toString() {
     return "ColorInfo("
-        + lumaBitdepthToString(lumaBitdepth)
-        + ", "
-        + chromaBitdepthToString(chromaBitdepth)
-        + ", "
         + colorSpaceToString(colorSpace)
         + ", "
         + colorRangeToString(colorRange)
@@ -367,6 +377,10 @@ public final class ColorInfo implements Bundleable {
         + colorTransferToString(colorTransfer)
         + ", "
         + (hdrStaticInfo != null)
+        + ", "
+        + lumaBitdepthToString(lumaBitdepth)
+        + ", "
+        + chromaBitdepthToString(chromaBitdepth)
         + ")";
   }
 
@@ -392,6 +406,7 @@ public final class ColorInfo implements Bundleable {
       default:
         return "Undefined color space";
     }
+    // LINT.ThenChange(C.java:color_space)
   }
 
   private static String colorTransferToString(@C.ColorTransfer int colorTransfer) {
@@ -414,6 +429,7 @@ public final class ColorInfo implements Bundleable {
       default:
         return "Undefined color transfer";
     }
+    // LINT.ThenChange(C.java:color_transfer)
   }
 
   private static String colorRangeToString(@C.ColorRange int colorRange) {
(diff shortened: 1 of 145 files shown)


dsparano:
Hi Michael, please let me know if you need any additional info from me for this PR, thanks.

dsparano:
Hi @microkatz any progress on this?

microkatz:
@dsparano

Thank you for your patience. I'm still reviewing the CL. I understand that you would like to have this data through H264 and H265 streams. In that regard, would you update the processH264FmtpAttribute and processH265FmtpAttribute methods in RtspMediaTrack.java to include your luma and chroma data in the format building?

dsparano:
Thanks Michael for reviewing this. I've added setting bit depths for H264 and H265 in RtspMediaTrack.java, I've also rebased against today's tip of main, just in case.

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
```diff
@@ -1237,6 +1247,22 @@ private static void parseVideoSampleEntry(
       } else if (childAtomType == Atom.TYPE_av1C) {
         ExtractorUtil.checkContainerInput(mimeType == null, /* message= */ null);
         mimeType = MimeTypes.VIDEO_AV1;
+        parent.setPosition(childStartPosition + Atom.HEADER_SIZE);
+        if (childAtomSize > Atom.HEADER_SIZE) {
+          parent.skipBytes(1); ;  // marker(1), version(7)
+          int byte2 = parent.readUnsignedByte();
+          int seqProfile = byte2 >> 5;
+          int byte3 = parent.readUnsignedByte();
+          boolean highBitdepth = ((byte3 >> 6) & 0b1) != 0;
```
microkatz, line 1256: The twelveBit boolean condition is only needed if (seqProfile == 2 && highBitdepth). If you could move this assignment into that segment.

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
```diff
@@ -1237,6 +1247,22 @@ private static void parseVideoSampleEntry(
       } else if (childAtomType == Atom.TYPE_av1C) {
         ExtractorUtil.checkContainerInput(mimeType == null, /* message= */ null);
         mimeType = MimeTypes.VIDEO_AV1;
+        parent.setPosition(childStartPosition + Atom.HEADER_SIZE);
+        if (childAtomSize > Atom.HEADER_SIZE) {
```
microkatz, line 1251: What is the reason behind using this as the condition to see if an AV1CodecConfigurationBox is included?

libraries/common/src/main/java/androidx/media3/common/Format.java
```diff
@@ -160,6 +162,8 @@ public static final class Builder {
     @Nullable private byte[] projectionData;
     private @C.StereoMode int stereoMode;
     @Nullable private ColorInfo colorInfo;
+    private int lumaBitdepth;
```
microkatz, line 165: Could you please instead add this bit depth info to the ColorInfo class?

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
microkatz, line 1: Since you are adding the luma and chroma information to vp9 in the mp4 AtomParsers, if you could please at the parsing logic for vp9 for rtsp in this file as well. It will require parsing the fmtp parameters if they exist.

dway123:
Hello @dsparano , would the Format.ColorInfo's ColorTransfer and ColorSpace values work to identify the bit depth?

My understanding is that 10-bit color is typically used with HDR, whereas 8-bit is typically used with SDR. Therefore, you could identify HDR-ness or 10-bit usage by checking whether the color transfer is COLOR_TRANSFER_ST2084 || COLOR_TRANSER_HLG, or ColorTransfer.isTransferHdr.

ColorSpace may also be able to identify the bit depth, but I'm less certain on that.

dsparano:
Hi @dway123, I think we cannot deny there is a correlation between colour metadata and bit depth, however an SDR content can still be on 10 bit or even more, and also HDR content can be encoded with 10, 12 or more (even 8, depending on how you define "HDR"). So I don't think the transfer characteristics itself, e.g. PQ or HLG, or color primaries should be used to assign a bit depth, not to mention that both H264 and HEVC allow for chroma to have a different bit depth than luma.

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
dsparano: The AV1 spec is mentioned at line 1258 to describe how to set `bitdepthLuma` from `seqProfile` and `highBitdepth`, however these two are read from `AV1CodecConfigurationRecord` (from https://aomediacodec.github.io/av1-isobmff/#av1codecconfigurationbox-syntax) which looks like this:
```
aligned (8) class AV1CodecConfigurationRecord {
  unsigned int (1) marker = 1;
  unsigned int (7) version = 1;
  unsigned int (3) seq_profile;
  unsigned int (5) seq_level_idx_0;
  unsigned int (1) seq_tier_0;
  unsigned int (1) high_bitdepth;
  unsigned int (1) twelve_bit;
  unsigned int (1) monochrome;
  unsigned int (1) chroma_subsampling_x;
  unsigned int (1) chroma_subsampling_y;
  unsigned int (2) chroma_sample_position;
  unsigned int (3) reserved = 0;

  unsigned int (1) initial_presentation_delay_present;
  if (initial_presentation_delay_present) {
    unsigned int (4) initial_presentation_delay_minus_one;
  } else {
    unsigned int (4) reserved = 0;
  }

  unsigned int (8) configOBUs[];
}
```
so I think the added lines here are correct. I could still add the link to the AV1CodecConfigurationRecord spec.

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
```diff
@@ -1237,6 +1247,22 @@ private static void parseVideoSampleEntry(
       } else if (childAtomType == Atom.TYPE_av1C) {
         ExtractorUtil.checkContainerInput(mimeType == null, /* message= */ null);
         mimeType = MimeTypes.VIDEO_AV1;
+        parent.setPosition(childStartPosition + Atom.HEADER_SIZE);
```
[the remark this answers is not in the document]
dsparano: This comes from a fix we made in our fork, where we found encodes that had missing HEVCDecoderConfigurationRecord so we added that check for all such atoms. However given that the specs, both ISO 14496-15 and the AV1 one, say "sample entry **shall** contain an ...Configuration Box" it seems such encodes would not be compliant.
Long story short I think you're right, I should remove the check at line 1251.

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
dsparano: This comes from a fix we made in our fork, where we found encodes that had missing HEVCDecoderConfigurationRecord so we added that check for all such atoms. However given that the specs, both ISO 14496-15 and the AV1 one, say "sample entry **shall** contain an ...Configuration Box" it seems such encodes would not be compliant.
Long story short I think you're right, I should remove the check at line 1251.

dway123:
Thanks @dsparano. Yeah, I'm aware that 8-bit HDR content can exist (but admittedly just haven't seen it generated myself yet). Have you seen HDR content with a bit depth other than 10-bit? 

I wasn't sure if bit-depth might be a proxy for HDR-ness here, but it seems this isn't the case. I don't think this is duplicating already-signalled info then.

I'm also unsure whether this benefits ExoPlayer/Transformer yet, as I haven't seen a use-case myself (the [`MediaFormat`](https://developer.android.com/reference/android/media/MediaFormat) used to configure `MediaCodec` doesn't use this), but I'm happy to defer to @microkatz on that.

libraries/common/src/main/java/androidx/media3/common/Format.java
dway123: I don't have a strong opinion on where this value should go, but I figure both the luma and chroma bitDepth should be in the same class (either ColorInfo or Format)

microkatz: Yes, my apology, the comment was to move both into ColorInfo. Could not attach the comment to multiple lines.

dsparano: Tbh it feels slightly wrong to me seeing luma bitdepth in ColorInfo, and since luma & chroma bitdepth are logically in pair
I would think they should go in Format, however I'm happy to move them to ColorInfo if you guys prefer so.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
dsparano: Sure, looking into this...

dsparano: If I'm correct from https://www.[ietf.org/archive/id/draft-ietf-payload-vp9-16.txt](https://www.ietf.org/archive/id/draft-ietf-payload-vp9-16.txt) the RTP payload for VP9 is not final yet. In the current draft the fmtp parameters for VP9 seem only optional: "max-fr", "max-fs" and "profile-id". None of them allows setting the bit depths. Incidentally I've played with mediamtx and I see it doesn't add fmtp at all for VP9.
Anyway, I can add a `processVP9FmtpAttribute()` method.

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
dsparano: Sorry, misunderstood your comment. Yes, twelveBit can be set before is needed.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMediaTrack.java
dsparano: Well, in more detail: I could add it, but it seems to me we couldn't get anything useful out of it, the only possibly relevant bit is the profile-id however for values 0 and 1 the bit depth is 8, which is already the default in Format.Builder, while for values of 2 and 3 the bit depth can be either 10 or 12, so we couldn't set any value. So @microkatz can we skip this?

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
dsparano: Done

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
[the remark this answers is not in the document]
dsparano: Conditional block removed, anyway (see other comment).

libraries/common/src/main/java/androidx/media3/common/Format.java
dsparano: @microkatz final decision ColorInfo or Format?

microkatz:
@dsparano 
No apology needed. We really appreciate your contribution to the library. I still need to review your PR but I have not forgotten!

As far as answers to your question.
1) ColorInfo please.
2) You can skip the processVP9FmtpAttribute()

dsparano:
Cool, no problem. I've moved the bit depths to ColorInfo, which opens the question of how to consider it "valid", atm I've tried to keep existing behaviour and at the same time allow for bit depths to be still reported in the debug text in case they are available and the colour metadata are not. Please check if you're happy with it.

microkatz:
Hello @dsparano, I was working on setting this PR up for internal review and merging. Unfortunately it seems that this commit is being made from a topic branch off of a private organization-owned fork? Would it be possible to open a new PR from an individual-owned fork? We can't push changes to organization-owned forks like this one. If that's not possible then we can still merge this PR but it will result in an ['evil' merge](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefevilmergeaevilmerge). See more info here: https://github.com/androidx/media/blob/release/CONTRIBUTING.md#push-access-to-pr-branches

libraries/common/src/main/java/androidx/media3/common/ColorInfo.java
```diff
@@ -258,7 +302,7 @@ public boolean isValid() {
    *
    * @see Format#toLogString(Format)
    */
-  public String toLogString() {
+  public String toColorString() {
```
microkatz, line 305: Can you keep this as toLogString? The purpose being is that it is used in debug/logging and used by parent Format#toLogString(Format). 

Please change the method to combine the bit depth info into the printing. If you want, you can have that if only the bit depth info is valid, then just print the bit depths. If none at all are valid, then print NA. If all are valid then print concatenation of the information.

dsparano: I think I've modified as requested.
In VideoSampleExporter and DefaultVideoFrameProcessor I thought I should keep the existing behaviour and only check for presence of Color metadata, therefore I've added isColorValid(). isValid() is now true if at least one between bitdepth and Color info is set, so that it will enable the logging.

libraries/common/src/main/java/androidx/media3/common/ColorInfo.java
```diff
@@ -288,10 +288,27 @@ public Builder buildUpon() {
...
+  public boolean isBitdepthValid() {
+    return lumaBitdepth != Format.NO_VALUE && chromaBitdepth != Format.NO_VALUE;
+  }
+
+  /**
+   * Returns whether this instance is color valid.
+   *
+   * <p>This instance is valid if no color members are {@link Format#NO_VALUE}.
+   */
+  public boolean isColorValid() {
```
microkatz, line 311: Please change this method to be called isDataspaceValid(). Name comes from the Android context combining this information: https://developer.android.com/reference/android/hardware/DataSpace.

libraries/common/src/main/java/androidx/media3/common/ColorInfo.java
microkatz: Thank you for your patience. One last comment which was to change isColorValid() to isDataSpaceValid() based on https://developer.android.com/reference/android/hardware/DataSpace

libraries/common/src/main/java/androidx/media3/common/ColorInfo.java
dsparano: Done

libraries/common/src/main/java/androidx/media3/common/ColorInfo.java
dsparano: Np, it's done.

microkatz: Thank you!

dsparano:
@microkatz It's been quiet for a while, is anything else needed for this to go in?

microkatz:

dsparano:
@microkatz has the internal review completed?

microkatz:
@dsparano. Not just yet. Much of team is currently at a summit and will return next week. We are still working on it. Thank you for your patience!

dsparano:
@microkatz Hi there, is anything else needed from our side on this PR?

PR #525, opened 2023-07-17
Update dash chunk source to handle in-progressive recording dash stream.

zgzong:
Using calculated segments timeline duration to compare with period duration to update whether period has reach it the end of stream.


tonihei:
Thanks for the PR! Could you change the target branch to `main` and not `release`?

PR #528, opened 2023-07-18
Update dash chunk source to handle in-progressive recording dash stream.

zgzong:
Using calculated segments timeline duration to compare with period duration to update whether period has reach it the end of stream.


zgzong:
It seems a common issue for Vod stream. The declared MPD@ mediaPresentationDuration is not accurate, it is bigger than one adaptationSet's duration. The stream packager may ignore the last hundreds milliseconds because it is too short to generate a segment.  
The multiple periods did meet same issue for these periods which ends already.

tonihei:
> The multiple periods did meet same issue for these periods which ends already.

Are you saying that the case you are trying to solve also has multiple periods that all have their duration set but are not yet available? Or is this more a generic comment agreeing with my change? 

The commit I was talking is uploaded (see above), so please take a look and leave any comments if needed.

zgzong:
> > The multiple periods did meet same issue for these periods which ends already.
> 
> Are you saying that the case you are trying to solve also has multiple periods that all have their duration set but are not yet available? Or is this more a generic comment agreeing with my change?
> 
> The commit I was talking is uploaded (see above), so please take a look and leave any comments if needed.

The issue we are facing is single period live stream. 
My original PR miss out the multi-periods cases, from theory they are same as the issue we meet in the vod streams. 
Vote up for your change.

tonihei:
Thanks for the confirmation! We'll work on submitting the change internally and then you should see the PR marked as merged soon.

PR #574, opened 2023-08-10
DASH UrlTemplate parses template if it contains more than one time ea…

hugohlln:
With DASH, it's possible to have multiple times each DASH identifier in a template (for example by using a proxy, see [Chaos streaming proxy](https://github.com/Eyevinn/chaos-stream-proxy)).
This commit allows to compile templates containing more than 4 identifiers (it was not possible before)

### DIFF
--- a/libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/manifest/UrlTemplate.java
+++ b/libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/manifest/UrlTemplate.java
@@ -17,6 +17,7 @@ package androidx.media3.exoplayer.dash.manifest;
 
 import androidx.media3.common.util.UnstableApi;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 
 /**
@@ -39,9 +40,9 @@ public final class UrlTemplate {
   private static final int BANDWIDTH_ID = 3;
   private static final int TIME_ID = 4;
 
-  private final ArrayList<String> urlPieces;
-  private final ArrayList<Integer> identifiers;
-  private final ArrayList<String> identifierFormatTags;
+  private final List<String> urlPieces;
+  private final List<Integer> identifiers;
+  private final List<String> identifierFormatTags;
   private final int identifierCount;
 
   /**
@@ -52,9 +53,9 @@ public final class UrlTemplate {
    * @throws IllegalArgumentException If the template string is malformed.
    */
   public static UrlTemplate compile(String template) {
-    ArrayList<String> urlPieces = new ArrayList<>();
-    ArrayList<Integer> identifiers = new ArrayList<>();
-    ArrayList<String> identifierFormatTags = new ArrayList<>();
+    List<String> urlPieces = new ArrayList<>();
+    List<Integer> identifiers = new ArrayList<>();
+    List<String> identifierFormatTags = new ArrayList<>();
 
     int identifierCount = parseTemplate(template, urlPieces, identifiers, identifierFormatTags);
     return new UrlTemplate(urlPieces, identifiers, identifierFormatTags, identifierCount);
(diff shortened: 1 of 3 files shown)


microkatz:
Hello @hugohlln,

Thank you for looking to contribute to ExoPlayer. Before we review your Pull Request, you'll need to sign the Contributor License Agreement (CLA). Thank you!

hugohlln:
Hi @microkatz,

Thanks for yout reply.

I've already sign the CLA :

image

It's just that checks are not updating maybe ?

brignolij:
@hugohlln thanks for the fix, will be very helpful.

sebimag:
@hugohlln  Wonderful fix this will help my team and allow us to continue working with exoPlayer

microkatz:
Hi @hugohlln,

I really appreciate your patience. Also thank you to the other developers for adding your support to this pull request in the comments!

@hugohlln, would you mind adding some unit tests in UrlTemplateTest.java with some segments that have more than one identifier?

libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/manifest/UrlTemplate.java
```diff
@@ -51,11 +51,16 @@ public final class UrlTemplate {
...
-    // most once in the template, which seems like a reasonable assumption.
-    String[] urlPieces = new String[5];
-    int[] identifiers = new int[4];
-    String[] identifierFormatTags = new String[4];
+    // We can have more than once each identifier in the template so we need
+    // to count the total number of identifiers present in the template
+    int identifiersCount = 0;
+    String[] identifiersNames = { REPRESENTATION, NUMBER, BANDWIDTH, TIME };
+    for(String identifierName : identifiersNames){
+      identifiersCount += template.split("\\$" + identifierName + "\\$").length - 1;
```
microkatz, line 59: Your regex pattern does not take into account the escape sequence $$. 

If your segment template was as follows "media/$$Bandwidth$$_c$$$Number$$$_mpd.m4s" with Bandwidth=650000 and Number=1, the solution would be "$Bandwidth$_c$1$_mpd.m4s". Your regex would mistake the $$Bandwidth$$ as an identifier.

hugohlln:
@microkatz 

Thanks for your reply.

I've just added tests in UrlTemplateTest and also changed my regex to match identifiers with escaped dollars.

All tests are sucessfully running

libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/manifest/UrlTemplate.java
```diff
@@ -51,11 +53,19 @@ public final class UrlTemplate {
...
+    // We can have more than once each identifier in the template so we need
+    // to count the total number of identifiers present in the template
+    int identifiersCount = 0;
+    String[] identifiersNames = { REPRESENTATION, NUMBER, BANDWIDTH, TIME };
+    for(String identifierName : identifiersNames){
+      Pattern pattern = Pattern.compile("(\\$" + identifierName + "\\$|" + ESCAPED_DOLLAR + "\\$" + identifierName + "\\$" + ESCAPED_DOLLAR +")");
+      Matcher matcher = pattern.matcher(template);
+      while(matcher.find())
+        identifiersCount++;
+    }
```
microkatz, line 65: Still incorrect regex as it matches $$Number$$ when it should not.

The parseTemplate method is already running through the string and pattern matching. 

I would suggest altering the class to use ArrayList for urlPieces, identifiers, and identifierFormat tags just so that it can grow dynamically.

hugohlln: Ok, sorry. I replaced arrays by ArrayLists. That should be ok now

libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/manifest/UrlTemplate.java
```diff
@@ -39,7 +40,7 @@ public final class UrlTemplate {
   private static final int TIME_ID = 4;
 
   private final String[] urlPieces;
-  private final int[] identifiers;
+  private final Integer[] identifiers;
```
microkatz, line 43: Might as well make these fields ArrayLists as you have changed the typing in all the methods. That way you don't need to do a conversion in the constructor.

microkatz:

PR #594, opened 2023-08-21
Format passed to audioSink.configure is missing information.

Tolriq:
AudioSink might require the identifying data and metadata to be able to take decisions.


tonihei:
Thanks for the PR! Will work on merging it internally and it will then be marked as merged here too.

Tolriq:
@tonihei thanks.
Anyway you can help me move on with https://github.com/androidx/media/pull/369 to generate the test asset?

While I can have a simple personal workaround extractor to address https://github.com/google/ExoPlayer/pull/10807, the id3decoder issue force me to keep a full fork of the extractor library for something really simple to fix :(

PR #633, opened 2023-09-03
GL: Fix non-standard call to glEnableVertexAttribArray

zhouweihui:
When setting up vertex arrays, glEnableVertexAttribArray should use the location parameter instead of the externally passed index parameter, to be consistent with the above glVertexAttribPointer.


droid-girl:
Hi @zhouweihui, based on your pull request proposal we made some changes in this file, including this particular line. I will close this one. These new changes will likely be included in the next release.
Thank you for your contribution.

PR #648, opened 2023-09-13
:bento: Rollback to the old exoplayer notifications icons in xml format

Didier116:
On Android 12, the default notification icons seem to be blurred.

Pixel 3 - Android 12 - Screenshot :
Pixel 3 - Android 12 - Screenshot

The problem seems to be that we're using default android png icons (which are surrounded by grey).
So I went back and added the exoplayer xml icons again (and renamed them at the same time).


consp1racy:
Vector drawables are only supported starting Android 5. This won't work as expected on Android 4.

marcbaechinger:
Thanks! I think something went wrong here.

If you want us to look into this PR, I think easiest would be to create a new one. I'd expect to see a single commit only that contains your changes. 

You also need to sign the CLA before we start looking into a PR.

If you still want to pursue this change it would be great if you can spend some time testing the XML icons with Android 21. I will do this also once you have created a new PR and signed the CLS. I'm not sure how far down XML icons are supported, so it's worth testing to find out why the @android:drawable/ic_media_play variants are bitmaps and not XML.

PR #650, opened 2023-09-13
Fix DTS Express Audio Buffer Underflow Issue.

cedricxperi:
Problem Description
DASH DTS Express audio playback results in buffer underrun issues. This can be heard as regular blips in the output.

Solution
DTS Express audio for streaming, uses a frame size (number of audio samples per channel per frame) of 4096. This requires a higher multiple for the buffersize computation.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
@@ -64,8 +64,7 @@ public class DefaultAudioTrackBufferSizeProvider
   private static final int AC3_BUFFER_MULTIPLICATION_FACTOR = 2;
 
   /**
-   * Default multiplication factor to apply to DTS Express passthrough buffer to avoid underruns on
-   * some devices (e.g., Xiaomi A2 TV).
+   * Default multiplication factor to apply to DTS Express passthrough buffer to avoid underruns.
    */
   private static final int DTSHD_BUFFER_MULTIPLICATION_FACTOR = 4;
 
@@ -153,8 +152,7 @@ public class DefaultAudioTrackBufferSizeProvider
 
     /**
      * Sets the multiplication factor to apply to the passthrough buffer for DTS-HD (DTS Express) to
-     * avoid underruns on some devices (e.g., Xiaomi A2 TV). Default is {@link
-     * #DTSHD_BUFFER_MULTIPLICATION_FACTOR}.
+     * avoid underruns. Default is {@link #DTSHD_BUFFER_MULTIPLICATION_FACTOR}.
      */
     @CanIgnoreReturnValue
     public Builder setDtshdBufferMultiplicationFactor(int dtshdBufferMultiplicationFactor) {
@@ -191,7 +189,7 @@ public class DefaultAudioTrackBufferSizeProvider
 
   /**
    * The multiplication factor to apply to DTS-HD (DTS Express) passthrough buffer to avoid
-   * underruns on some devices (e.g., Xiaomi A2 TV).
+   * underruns.
    */
   public final int dtshdBufferMultiplicationFactor;
 
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
@@ -26,6 +26,7 @@ import static java.lang.Math.max;
 import android.media.AudioTrack;
 import androidx.media3.common.C;
 import androidx.media3.common.Format;
+import androidx.media3.common.MimeTypes;
 import androidx.media3.common.util.UnstableApi;
 import androidx.media3.exoplayer.audio.DefaultAudioSink.OutputMode;
 import androidx.media3.extractor.AacUtil;
@@ -62,7 +63,11 @@ public class DefaultAudioTrackBufferSizeProvider
    * devices (e.g., Broadcom 7271).
    */
   private static final int AC3_BUFFER_MULTIPLICATION_FACTOR = 2;
-
+  /**
+   * Default multiplication factor to apply to DTS Express passthrough buffer to avoid underruns
+   * on some devices (e.g., Xiaomi A2 TV).
+   */
+  private static final int DTSE_BUFFER_MULTIPLICATION_FACTOR = 4;
   /** A builder to create {@link DefaultAudioTrackBufferSizeProvider} instances. */
   public static class Builder {
 
@@ -72,7 +77,7 @@ public class DefaultAudioTrackBufferSizeProvider
     private int passthroughBufferDurationUs;
     private int offloadBufferDurationUs;
     private int ac3BufferMultiplicationFactor;
-
+    private int dtseBufferMultiplicationFactor;
     /** Creates a new builder. */
     public Builder() {
       minPcmBufferDurationUs = MIN_PCM_BUFFER_DURATION_US;
@@ -81,6 +86,7 @@ public class DefaultAudioTrackBufferSizeProvider
       passthroughBufferDurationUs = PASSTHROUGH_BUFFER_DURATION_US;
       offloadBufferDurationUs = OFFLOAD_BUFFER_DURATION_US;
       ac3BufferMultiplicationFactor = AC3_BUFFER_MULTIPLICATION_FACTOR;
+      dtseBufferMultiplicationFactor = DTSE_BUFFER_MULTIPLICATION_FACTOR;
     }
 
     /**
@@ -143,6 +149,17 @@ public class DefaultAudioTrackBufferSizeProvider
       return this;
     }
 
+    /**
+     * Sets the multiplication factor to apply to the passthrough buffer for DTS Express to avoid
+     * underruns on some devices (e.g., Xiaomi A2 TV). Default is
+     * {@value #DTSE_BUFFER_MULTIPLICATION_FACTOR}.
+     */
+    @CanIgnoreReturnValue
+    public Builder setDtseBufferMultiplicationFactor(int dtseBufferMultiplicationFactor) {
+      this.dtseBufferMultiplicationFactor = dtseBufferMultiplicationFactor;
+      return this;
+    }
+
     /** Build the {@link DefaultAudioTrackBufferSizeProvider}. */
     public DefaultAudioTrackBufferSizeProvider build() {
       return new DefaultAudioTrackBufferSizeProvider(this);
@@ -169,7 +186,11 @@ public class DefaultAudioTrackBufferSizeProvider
    * (e.g., Broadcom 7271).
    */
   public final int ac3BufferMultiplicationFactor;
-
+  /**
+   * The multiplication factor to apply to DTS Express passthrough buffer to avoid underruns on some
+   * devices (e.g., Xiaomi A2 TV).
+   */
+  public final int dtseBufferMultiplicationFactor;
   protected DefaultAudioTrackBufferSizeProvider(Builder builder) {
     minPcmBufferDurationUs = builder.minPcmBufferDurationUs;
     maxPcmBufferDurationUs = builder.maxPcmBufferDurationUs;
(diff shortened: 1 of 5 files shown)


cedricxperi:
@tianyif This is the same pull request as #450. I re-applied the changes to the latest main branch. Thanks.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
```diff
@@ -76,6 +76,19 @@ public final class DtsUtil {
         64, 112, 128, 192, 224, 256, 384, 448, 512, 640, 768, 896, 1024, 1152, 1280, 1536, 1920,
         2048, 2304, 2560, 2688, 2816, 2823, 2944, 3072, 3840, 4096, 6144, 7680
       };
+  private static String currentMimeType;
```
tianyif, line 79: It's error-prone to set static property on a util class as calling `clearCurrentMimeType` can be easily forgotten. 

It looks like we only need this information in `DefaultAudioTrackBufferSizeProvider`, so I think we have below options:

1. In `DefaultAudioSink.configure`, when calling method [`DefaultAudioTrackBufferSizeProvider.getBufferSizeInBytes`](https://github.com/androidx/media/blob/main/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L751), pass the encoding as `ENCODING_DTS_HD` if: `outputEncoding == ENCODING_DTS_HD` OR `outputEncoding == ENCODING_DTS && inputFormat.sampleMimeType.contentEquals(MimeTypes.AUDIO_DTS_EXPRESS)`. Note that this "modified" encoding can also flow to the offload branch, if this only applies to passthrough, we also need to check the output mode before passing the "modified" encoding.

2. Add a parameter of `sampleMimeType` to `DefaultAudioTrackBufferSizeProvider.getBufferSizeInBytes`, this requires the customized implementation of `AudioTrackBufferSizeProvider` to change accordingly though.

How do you think of these two options?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
```diff
@@ -232,7 +254,20 @@ protected int getPassthroughBufferSizeInBytes(@C.Encoding int encoding, int bitr
     int bufferSizeUs = passthroughBufferDurationUs;
     if (encoding == C.ENCODING_AC3) {
       bufferSizeUs *= ac3BufferMultiplicationFactor;
+    } else if ((DtsUtil.getCurrentMimeType().contentEquals(MimeTypes.AUDIO_DTS_EXPRESS) && (bitrate
```
tianyif, line 257: Could you please add tests for the DTS Express case? [`DefaultAudioTrackBufferSizeProviderAC3Test`](https://github.com/androidx/media/blob/main/libraries/exoplayer/src/test/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProviderAC3Test.java) can be a good example.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
```diff
@@ -72,7 +77,7 @@ public static class Builder {
     private int passthroughBufferDurationUs;
     private int offloadBufferDurationUs;
     private int ac3BufferMultiplicationFactor;
-
+    private int dtseBufferMultiplicationFactor;
```
tianyif, line 80: nit: Add a blank line after this line.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
```diff
@@ -62,7 +63,11 @@ public class DefaultAudioTrackBufferSizeProvider
    * devices (e.g., Broadcom 7271).
    */
   private static final int AC3_BUFFER_MULTIPLICATION_FACTOR = 2;
-
+  /**
+   * Default multiplication factor to apply to DTS Express passthrough buffer to avoid underruns
+   * on some devices (e.g., Xiaomi A2 TV).
+   */
+  private static final int DTSE_BUFFER_MULTIPLICATION_FACTOR = 4;
```
tianyif, line 70: nit: Add a blank line after this line.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
cedricxperi: Done.

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
cedricxperi: @tianyif I double checked with my TV SOC colleague and turns out that ENCODING_DTS_HD should be supported by all implementations for DTS Express. So I think there is no need to remap anything. The presence of ENCODING_DTS_HD should be enough to signify that the input stream is actually DTS Express and we can proceed to multiple the buffer size by dtseBufferMultiplicationFactor. I have removed currentMimeType etc in the new commit. Let me know if it is ok. Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
```diff
@@ -62,8 +63,13 @@ public class DefaultAudioTrackBufferSizeProvider
...
```
tianyif, line 70: nit: How do you think of changing the naming to `DTSHD_BUFFER_MULTIPLICATION_FACTOR`?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
```diff
@@ -72,6 +78,7 @@ public static class Builder {
     private int passthroughBufferDurationUs;
     private int offloadBufferDurationUs;
     private int ac3BufferMultiplicationFactor;
+    private int dtseBufferMultiplicationFactor;
```
tianyif, line 81: nit: How do you think of changing the naming to `dtshdBufferMutiplicationFactor`? 

If you agree with me, could you please also make the according change to the setter's name?

libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
tianyif: Sounds good. Thanks for letting me know!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
cedricxperi: I think it is a good idea, I changed all the names from dtse to dtshd. Thanks.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
cedricxperi: Good idea. Done.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
```diff
@@ -143,6 +153,17 @@ public Builder setAc3BufferMultiplicationFactor(int ac3BufferMultiplicationFacto
       return this;
     }
 
+    /**
+     * Sets the multiplication factor to apply to the passthrough buffer for DTS-HD (DTS Express)
+     * to avoid underruns on some devices (e.g., Xiaomi A2 TV). Default is
+     * {@value #DTSHD_BUFFER_MULTIPLICATION_FACTOR}.
```
tianyif, line 159: nit: `{@link #DTSHD_BUFFER_MULTIPLICATION_FACTOR}` instead of `@value`.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
```diff
@@ -232,7 +258,13 @@ protected int getPassthroughBufferSizeInBytes(@C.Encoding int encoding, int bitr
     int bufferSizeUs = passthroughBufferDurationUs;
     if (encoding == C.ENCODING_AC3) {
       bufferSizeUs *= ac3BufferMultiplicationFactor;
+    } else if (encoding == C.ENCODING_DTS_HD) {
+      // DTSHD (DTS Express) for streaming uses a frame size (number of audio samples per channel
```
tianyif, line 262: super-nit: DTS-HD (a dash "-" in between)?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
```diff
@@ -232,7 +258,13 @@ protected int getPassthroughBufferSizeInBytes(@C.Encoding int encoding, int bitr
...
+      // per frame of 4096. This requires a higher multiple for the buffersize computation.
```
tianyif, line 263: nit: A missing closing `)` after the second word "frame"? The corresponding open `(` is in the last line.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
```diff
@@ -62,8 +62,16 @@ public class DefaultAudioTrackBufferSizeProvider
...
    */
   private static final int AC3_BUFFER_MULTIPLICATION_FACTOR = 2;
+  /**
+   * Default multiplication factor to apply to DTS Express passthrough buffer to avoid underruns
+   * on some devices (e.g., Xiaomi A2 TV).
+   */
+  private static final int DTSHD_BUFFER_MULTIPLICATION_FACTOR = 4;
+
+  /**
+   * A builder to create {@link DefaultAudioTrackBufferSizeProvider} instances.
```
tianyif, line 72: nit: Since there is no actual required diff at lines 71-74, could you please revert these lines back to the original state?

tianyif (commented):
Thanks for the changes! I only left some nit comments, and after addressing them I will bring this PR to the internal review stage.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
cedricxperi: Done

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
cedricxperi: Done

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
cedricxperi: Done

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioTrackBufferSizeProvider.java
cedricxperi: Done

tianyif:

tianyif:
@cedricxperi Sorry for the delay on this PR. I was just back from the holiday :) 

Quick question for double check: Does this buffer underrun issues apply to all devices or only apply to some of them like Xiaomi A2 TV? I'm asking to see if we have to mention the devices in the comment lines.

cedricxperi:
> @cedricxperi Sorry for the delay on this PR. I was just back from the holiday :)
> 
> Quick question for double check: Does this buffer underrun issues apply to all devices or only applies to some of them like Xiaomi A2 TV? I'm asking to see if we have to mention the devices in the comment lines.

Hi @tianyif no worries. This will happen for all devices as it is not a processor speed issue. Thanks

Issue #659, opened 2023-09-18
Get access to public functions in MediaDrm that are not exposed to ExoMediaDrm


shanujshekhar:
Hey,
I'm trying to understand how does one use a function from APIs in `MediaDrm` that are not part of `ExoMediaDrm` .
- This is in regards to [MediaDrm#removeOfflineLicense](https://developer.android.com/reference/android/media/MediaDrm#removeOfflineLicense(byte[])) API that does not exist in ExoMediaDrm for our APIs to consume.

icbaker:
Would it make sense to add a method like `OfflineLicenseHelper.removeLicense(byte[])`? In general I would expect most apps to use `OfflineLicenseHelper` for managing downloaded licenses, rather than directly interacting with an `ExoMediaDrm` instance. If you're using `ExoMediaDrm` directly, could you add a bit of detail about exactlly what your setup looks like?

(this may in turn require adding a similar method to `ExoMediaDrm` as part of the implementation, but I just want to make sure we're working downwards from the API requirements rather than upwards from the API offered by the platform `MediaDrm` class)

shanujshekhar:
Yes, it would help adding `OfflineLicenseHelper.removeLicense(byte[])` in OfflineLicenseHelper but it would also be helpful if we have `removeOfflineLicenses` in `ExoMediaDrm` too so that we can leverage that API incase we already have access to `ExoMediaDrm`.

The setup for our `ExoMediaDrm` is that we are offering the custom `ExoMediaDrm` using [DefaultDrmSessionManager.Builderl#setUuidAndExoMediaDrmProvider](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/drm/DefaultDrmSessionManager.Builder.html#setUuidAndExoMediaDrmProvider(java.util.UUID,com.google.android.exoplayer2.drm.ExoMediaDrm.Provider)) in the `DefaultDrmSessionManager` which the `OfflineLicenseHelper` uses.

On a side note, I was also looking at this API to get offline licenses IDs ([MediaDrm#getOfflineLicenseKeySetIds()](https://developer.android.com/reference/android/media/MediaDrm#getOfflineLicenseKeySetIds())) which would be helpful for our use case. This API is also not available in `ExoMediaDrm` or `OfflineLicenseHelper`, so it would be really helpful if we can have access to it also

shanujshekhar:
We use Widevine L1 and alternate Widevine integration and we would like to have the functionality of removing offline licenses for both integration. 

[...]

PR #675, opened 2023-09-25
Add new APIs to ExoMediaDrm

shanujshekhar:
Changes
---
- Added `removeOfflineLicense(byte[])` and `getOfflineLicenseKeySetIds` and consumed them in their implementations

Background
---
- These APIs will help in addressing an increasing amount of `java.lang.IllegalArgumentException: Failed to restore keys: BAD_VALUE` which is our top playback error in our app
	- Based on our discussion with Widevine team and [this exoplayer issue](https://github.com/google/ExoPlayer/issues/11202#issuecomment-1708792594)
		- TL;DR: The failure occurs on startup if the user has 200+ offline licenses, we would like to add the functionality to remove offline licenses

**Note: Why we want these APIs in ExoMediaDrm and not in OfflineLicenseHelper**

- As per the issue above, we would like to access these 2 public APIs in MediaDrm that don’t exist in `OfflineLicenseHelper` or `ExoMediaDrm`
	- APIs interested in:
		- [MediaDrm#removeOfflineLicense()](https://developer.android.com/reference/android/media/MediaDrm#removeOfflineLicense(byte%5B%5D)): To remove offline license
		- [MediaDrm#getOfflineLicenseKeySetIds()](https://developer.android.com/reference/android/media/MediaDrm#getOfflineLicenseKeySetIds()): To see number of offline licenses on startup

	- We use `OfflineLicenseHelper` to download license for L1 and we don't interact with `ExoMediaDrm` directly. But for the alternate Widevine integration, we directly depend on `ExoMediaDrm` APIs to override and call CDM Native APIs.
	- We would like to have the functionality of removing offline licenses for both integration which would need access to above APIs in `ExoMediaDrm`.

Links
---
- https://github.com/androidx/media/issues/659

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
@@ -24,7 +24,7 @@ import androidx.media3.common.DrmInitData;
 import androidx.media3.common.util.UnstableApi;
 import androidx.media3.common.util.Util;
 import androidx.media3.decoder.CryptoConfig;
-import java.util.Collections;
+import com.google.common.collect.ImmutableList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -122,6 +122,17 @@ public final class DummyExoMediaDrm implements ExoMediaDrm {
   }
 
   @Override
+  public void removeOfflineLicense(byte[] keySetId) {
+    // Should not be invoked. No session should exist.
+    throw new IllegalStateException();
+  }
+
+  @Override
+  public ImmutableList<byte[]> getOfflineLicenseKeySetIds() {
+    return ImmutableList.of();
+  }
+
+  @Override
   @Nullable
   public PersistableBundle getMetrics() {
     return null;
@@ -156,16 +167,5 @@ public final class DummyExoMediaDrm implements ExoMediaDrm {
   @Override
   public @C.CryptoType int getCryptoType() {
     return C.CRYPTO_TYPE_UNSUPPORTED;
-  }
-
-  @Override
-  public void removeOfflineLicense(byte[] keySetId) {
-    // Should not be invoked. No session should exist.
-    throw new IllegalStateException();
-  }
-
-  @Override
-  public List<byte[]> getOfflineLicenseKeySetIds() {
-    return Collections.emptyList();
   }
 }
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
@@ -24,6 +24,7 @@ import androidx.media3.common.DrmInitData;
 import androidx.media3.common.util.UnstableApi;
 import androidx.media3.common.util.Util;
 import androidx.media3.decoder.CryptoConfig;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -155,5 +156,16 @@ public final class DummyExoMediaDrm implements ExoMediaDrm {
   @Override
   public @C.CryptoType int getCryptoType() {
     return C.CRYPTO_TYPE_UNSUPPORTED;
+  }
+
+  @Override
+  public void removeOfflineLicense(byte[] keySetId) {
+    // Should not be invoked. No session should exist.
+    throw new IllegalStateException();
+  }
+
+  @Override
+  public List<byte[]> getOfflineLicenseKeySetIds() {
+    return Collections.emptyList();
   }
 }
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
@@ -24,7 +24,6 @@ import androidx.media3.common.DrmInitData;
 import androidx.media3.common.util.UnstableApi;
 import androidx.media3.common.util.Util;
 import androidx.media3.decoder.CryptoConfig;
-import com.google.common.collect.ImmutableList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/ExoMediaDrm.java
```diff
@@ -583,4 +583,14 @@ byte[] provideKeyResponse(byte[] scope, byte[] response)
    */
   @C.CryptoType
   int getCryptoType();
+
+  /**
+   * See {@link MediaDrm#removeOfflineLicense(byte[])}
```
icbaker, line 588: It's OK to `@see` or `See @link ...` to the `MediaDrm` docs in a second line, but let's make the first line of this javadoc stand alone.

I also think it's worth flagging the caveat that this method is generally not needed here too.

So something like: 
> Removes an offline license.
>
> This method is generally not needed, and  should only be used if the preferred approach of passing {@link KEY_TYPE_RELEASE} to {@link #getKeyRequest} is not possible.
>
> See {@link MediaDrm.removeOfflineLicense} for more details.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/FrameworkMediaDrm.java
```diff
@@ -380,6 +382,21 @@ public FrameworkCryptoConfig createCryptoConfig(byte[] sessionId) throws MediaCr
     return C.CRYPTO_TYPE_FRAMEWORK;
   }
 
+  @Override
+  public void removeOfflineLicense(byte[] keySetId) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
icbaker, line 387: I don't think it's really acceptable to make this method a no-op on pre-Q devices without making that much clearer in the interface.

See for example the docs on `setOnKeyStatusChangeListener` and the implementation that throws an exception on unsupported API levels.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
```diff
@@ -156,4 +156,14 @@ public CryptoConfig createCryptoConfig(byte[] sessionId) {
...
   }
+
+  @Override
+  public void removeOfflineLicense(byte[] keySetId) {
+    // Do nothing
+  }
+
+  @Override
+  public List<byte[]> getOfflineLicenseKeySetIds() {
+    return null;
```
icbaker, line 167: This method is not annotated `@Nullable`, so it's implicitly non-null. Probably best to return an empty list here.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
```diff
@@ -156,4 +156,14 @@ public CryptoConfig createCryptoConfig(byte[] sessionId) {
   public @C.CryptoType int getCryptoType() {
     return C.CRYPTO_TYPE_UNSUPPORTED;
   }
+
+  @Override
+  public void removeOfflineLicense(byte[] keySetId) {
+    // Do nothing
```
icbaker, line 162: For consistency with methods above that take `keySetId` (like `getKeyRequest`) this should probably throw an exception

icbaker (changes requested):

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/ExoMediaDrm.java
shanujshekhar: Thanks for reviewing. Updated.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/FrameworkMediaDrm.java
shanujshekhar: Thanks for reviewing. Updated.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
shanujshekhar: Thanks for reviewing. Updated.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DummyExoMediaDrm.java
shanujshekhar: Thanks for reviewing. Updated.

icbaker (approved):

shanujshekhar:
Hey Ian, nothing urgent but I had a follow-up question I posted in [the issue](https://github.com/androidx/media/issues/659#issuecomment-1734032140), not sure if you had a chance to take a look but it would be really great if you could help in that regard. Thank you!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/FrameworkMediaDrm.java
```diff
@@ -327,6 +329,24 @@ public void restoreKeys(byte[] sessionId, byte[] keySetId) {
...
+      throw new UnsupportedOperationException();
+    }
+    mediaDrm.removeOfflineLicense(keySetId);
+  }
+
+  @Override
+  @RequiresApi(Build.VERSION_CODES.Q)
+  public List<byte[]> getOfflineLicenseKeySetIds() {
+    if (Util.SDK_INT < Build.VERSION_CODES.Q) {
+      return ImmutableList.of();
```
shanujshekhar, line 345: Thinking if we should also throw `UnsupportedOperationException()` exception incase of unsupported API level to match the pattern

icbaker: Yes you're right, that is easier to reason about and more consistent - done.

shanujshekhar: Got it. Thank you!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/ExoMediaDrm.java
```diff
@@ -523,6 +523,32 @@ byte[] provideKeyResponse(byte[] scope, byte[] response)
...
   void restoreKeys(byte[] sessionId, byte[] keySetId);
 
+  /**
+   * Removes an offline license.
+   *
+   * <p>This method is generally not needed, and should only be used if the preferred approach of
+   * generating a license release request by passing {@link #KEY_TYPE_RELEASE} to {@link
+   * #getKeyRequest} is not possible.
+   *
+   * <p>This is an optional method, and some implementations may only support it on certain Android
```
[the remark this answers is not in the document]
icbaker: Good point, done on both the new methods.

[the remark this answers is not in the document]
shanujshekhar: Thank you so much!

Issue #702, opened 2023-10-05
Support synchronized live playback for HLS


stevemayhew:
### Use case description

Multiple Android TV streamers in a bar or other hospitality situation require video synchronized to a single live offset.   This feature is documented in https://developer.android.com/guide/topics/media/exoplayer/live-streaming

The problem is this unlikely to work for HLS, as there is no spec mandated synchronization of time in the client to common time (NTP or the origin server).  The spec does recommend msec accurate wall-clock time in the playlist:

> #EXT-X-PROGRAM-DATE-TIME:<date-time-msec>
>
>   where date-time-msec is an ISO/IEC 8601:2004 [[ISO_8601](https://datatracker.ietf.org/doc/html/rfc8216#ref-ISO_8601)] date/time
>   representation, such as YYYY-MM-DDThh:mm:ss.SSSZ.  It SHOULD indicate
>   a time zone and fractional parts of seconds, to millisecond accuracy.

So, all the pieces except for time synchronization are present.

Note, this is also requested in https://github.com/google/ExoPlayer/issues/10990

### Proposed solution

Simple, synchronize `androidx.media3.exoplayer.hls.HlsMediaSource.Factory#elapsedRealTimeOffsetMs

### Alternatives considered
n/a

PR #697, opened 2023-10-04
SntpClient periodically re-syncs the offset to NTP server

stevemayhew:
Many of the Android TV platforms do not report accurate `SystemClock.elapsedRealtime()` In addition this drift is not consistant accross multiple boxes, so the time offset between any pair of boxes will vary over time.

This variance will lead to a drift in the Live Offset calculation.

Fix is simple, inValidate the NTP server query after a period of time (10 minutes) and re-issue the call

PR #703, opened 2023-10-05
Allows exact live sync for HLS

stevemayhew:
This pull request satisfies the enhancement request #702.   For playback longer than a few hours the fix for SntpClient is also required, see pull request #697


libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/HlsMediaSource.java
```diff
@@ -557,7 +559,7 @@ private SinglePeriodTimeline createTimelineForLive(
     return new SinglePeriodTimeline(
         presentationStartTimeMs,
         windowStartTimeMs,
-        /* elapsedRealtimeEpochOffsetMs= */ C.TIME_UNSET,
+        SntpClient.getElapsedRealtimeOffsetMs(),
         periodDurationUs,
```
stevemayhew, line 563: If the SntpClient is not initialized this change done nothing, so you could consider making the changes to `DefaultHlsPlaylistTracker` surrounded by a feature flag from the `HlsPlaylistTracker.Factory` (if it was not private), or simply a PSFB flag.

libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/playlist/DefaultHlsPlaylistTracker.java
```diff
@@ -141,6 +146,25 @@ public void start(
...
+          }
+        };
+    SntpClient.initialize(initialPlaylistLoader, callback);
+  }
+
+  private void requestMasterPlaylist(
+      EventDispatcher eventDispatcher, ParsingLoadable<HlsPlaylist> multivariantPlaylistLoadable) {
     long elapsedRealtime =
         initialPlaylistLoader.startLoading(
             multivariantPlaylistLoadable,
```
stevemayhew, line 170: The `requestMasterPlaylist` was simply a refactoring (extract method), then call the method in the `SntpClient` callback.  Most times this is a no-op as the `SntpClient` offset is up to date.

libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/playlist/DefaultHlsPlaylistTracker.java
```diff
@@ -695,6 +719,27 @@ private void loadPlaylistInternal(Uri playlistRequestUri) {
...
+              }
+            };
+        SntpClient.initialize(mediaPlaylistLoader, callback);
+      } else {
+        loadAfterTimeSync(playlistRequestUri);
+      }
+    }
+
+    private void loadAfterTimeSync(Uri playlistRequestUri) {
       ParsingLoadable.Parser<HlsPlaylist> mediaPlaylistParser =
```
stevemayhew, line 743: The same refactor lift method to create `loadAfterTimeSync()`.  Here only the primary playlist load needs to possibly refresh `SntpClient` as the master playlist is only loaded once

stevemayhew:
Thanks @BogdanLivadariu no one is giving it any attention, we've already shipped it ;-)

stevemayhew:
Closing this mess and re-opening with base against main

PR #709, opened 2023-10-07
Eat SysEx message

loki666:
Eat/ignore sysEx message instead of choking on them


loki666:
will reopen on main branch

PR #710, opened 2023-10-07
MIDI: Ignore SysEx messages

loki666:
ignore SysEx messages instead of choking

### DIFF
--- a/libraries/decoder_midi/src/main/java/androidx/media3/decoder/midi/TrackEvent.java
+++ b/libraries/decoder_midi/src/main/java/androidx/media3/decoder/midi/TrackEvent.java
@@ -84,9 +84,40 @@ import androidx.media3.common.util.UnstableApi;
     int firstByte = parsableTrackEventBytes.readUnsignedByte();
     eventDecoderSizeBytes = 1;
 
-    if ((firstByte & 0xF0) != 0xF0) {
-      // Most significant nibble is not 0xF, this is a MIDI channel event.
+    if (firstByte == SYSEX_BEGIN_STATUS) {
+      // TODO(b/228838584): Handle this gracefully.
+      statusByte = firstByte;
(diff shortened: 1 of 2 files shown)


libraries/decoder_midi/src/main/java/androidx/media3/decoder/midi/TrackEvent.java
```diff
@@ -139,10 +141,17 @@ public boolean populateFrom(ParsableByteArray parsableTrackEventBytes, int previ
           default: // Ignore all other Meta events.
             parsableTrackEventBytes.skipBytes(eventLength);
         }
-      } else {
+      } else if (firstByte == SYSEX_BEGIN_STATUS) {
```
christosts, line 144: Thanks for the contribution.

Since you're here, can you change the if-else structure to be

```
if ((firstByte & 0xF0) != 0xF0) {
  ...
} else if (firstByte == META_EVENT_STATUS) {
  ...
} else if (firstByte == SYSEX_BEGIN_STATUS) { 
  ...
} else {
throw ParserException.createForUnsupportedContainerFeature(
            "Unknown track event: " + firstByte);
}

File: libraries/decoder_midi/src/main/java/androidx/media3/decoder/midi/TrackEvent.java [hunk 2]
```diff
@@ -139,10 +141,17 @@ public boolean populateFrom(ParsableByteArray parsableTrackEventBytes, int previ
...
         // TODO(b/228838584): Handle this gracefully.
+        statusByte = firstByte;
+
+        int currentByte;
+        do { // eat SysEx Message
```
christosts, line 149: let's do this "Consume SysEx message."

christosts:

PR #784, opened 2023-11-01
RFC SQLiteDatabase -> SupportSQLiteDatabase

dabluck:
Would like feedback on approach before polishing. A few questions:

- I assume we are not okay with with common dep on sqlite library. I will make a DatabaseUtil in database lib and hopefully can remove that dependency?
- Is the StandaloneDatabaseProvider relying on androidx.sqlite:sqlite-framework an okay approach?
- Any advice for more tests to write / how to test migration scenarios?

Thanks

### DIFF
--- a/constants.gradle
+++ b/constants.gradle
@@ -51,6 +51,7 @@ project.ext {
     androidxMultidexVersion = '2.0.1'
     androidxRecyclerViewVersion = '1.3.0'
     androidxMaterialVersion = '1.8.0'
+    androidxSqliteVersion = '2.3.1'
     androidxTestCoreVersion = '1.5.0'
     androidxTestEspressoVersion = '3.5.1'
     androidxTestJUnitVersion = '1.1.5'
--- a/libraries/common/build.gradle
+++ b/libraries/common/build.gradle
@@ -65,7 +65,6 @@ dependencies {
     }
     api 'androidx.annotation:annotation-experimental:' + androidxAnnotationExperimentalVersion
     implementation 'androidx.annotation:annotation:' + androidxAnnotationVersion
-    implementation 'androidx.sqlite:sqlite:' + androidxSqliteVersion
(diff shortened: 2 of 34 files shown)


tonihei:
Thanks for the draft, that's super useful to see! 

> I assume we are not okay with with common dep on sqlite library.

I think it's logical and okay to depend on the sqlite lib from the database module, but the indirect dependencies for all datasource and exoplayer module users are not definitely not ideal. We'll discuss this a bit internally to see what the best solution would look like and then report that back here. 

> Is the StandaloneDatabaseProvider relying on androidx.sqlite:sqlite-framework an okay approach?

Looks right in general. The only changes we should be mindful about are breaking changes for existing users of such classes where switching the type would cause compilation problems. A common approach would be to have both paths for a while, deprecating the old one and remove it later. 

> Any advice for more tests to write / how to test migration scenarios?

To some extent, the existing tests already cover all cases because the you don't change any real functionality. What's worth testing is how much or little these changes break existing app integrations (e.g. the ExoPlayer demo app) as this gives a good signal of what we need to be careful about.

tonihei:
> We'll discuss this a bit internally to see what the best solution would look like and then report that back here.

Based on our discussion, the best final state is that all cache and offline classes move to their own module (instead of being part of the DataSource and ExoPlayer modules). This allows to add new dependencies to the Database modules without affecting all ExoPlayer users.

Unfortunately, I don't really see a way to integrate the proposed changes here without doing this module restructure first. And it's also something we need to do internally ourselves as PRs can't fully capture all the setup needed for a module in our internal code tree. 

We probably won't do it immediately, but will consider this change for 1.3.0 (depending on the actual effort we may end up postponing it again, so I can't promise it yet)

tonihei:
I'll also close this PR for now because we won't be able to merge its current version anyway. Once the module restructure has been done, someone can open a new PR against the new structure with the same changes I think.

dabluck:
Thanks @tonihei hopefully this can be prioritized

PR #837, opened 2023-11-24
Apply ClearKey DRM invalid URL workaround for Android 13 and newer

superjohan:
At least some Android 14 devices still have the same invalid URL issue when using ClearKey DRM, so might as well apply the check to newer versions of Android as well. We have seen this in the wild on Samsung S23 series devices at least.

Original fix: https://github.com/androidx/media/commit/715c948004eeabb6f7e2d45b8b8856ee4a015391

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -53,6 +53,9 @@
 *   Text:
 *   Metadata:
 *   DRM:
+    *   Extend workaround for spurious ClearKey `https://default.url` license
+        URL to API 33+ (previously the workaround only applied on API 33
+        exactly) ([#837](https://github.com/androidx/media/pull/837)).
 *   Effect:
 *   Muxers:
 *   IMA extension:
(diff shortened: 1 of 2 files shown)


icbaker:
Thanks for flagging this, I've confirmed I see the same behaviour on a Samsung S23+ running Android 14 (`samsung/dm2qxxx/dm2q:14/UP1A.230623.007/S916BXXU3ZWH8:user/release-keys`). I've queried with the DRM team whether there's another way we can determine if a device has [this fix](https://cs.android.com/android/_/android/platform/frameworks/av/+/3c57f9101cb06aab74556172fdac78c002ec2064) present or not (and therefore the workaround shouldn't be needed). In the meantime, the workaround seems very safe to apply on all devices >= 33, since it will only 'break' a playback if the **real** license URL is exactly `https://default.url` (no path).

I'll get this change merged.

PR #867, opened 2023-12-06
ffmpeg: add support of specifying Android ABI version when building FFmpeg

equeim:


tonihei:
Thanks for updating the script! I will start the merge process.

PR #914, opened 2023-12-21
Fix forward and rewind button texts cut off with material theme applied

### LINKED ISSUE #547: Text in forward button inside player controls not readable when using Material 3 theme
### Version

Media3 1.1.0

### More version details

_No response_

Should affect all devices, but tested with:
- Fold-Out Emulator Android 11
- Pixel 7 Android 13

Some more users reported this on Zapp: https://github.com/mediathekview/zapp/issues/366

_No response_

Yes

### Reproduction steps

- Replace the parent of `PlayerTheme` in the demo app with `Theme.Material3.Dark` or some other material 3 theme
- Play back any video
- Tap to show the control view buttons

### Expected result

The prev button should display the default **15 seconds forward increment** in a readable way.

### Actual result

The prev button seemingly displays a **5 seconds forward increment**: 
grafik

However, according to the layout inspector the text is set correctly. It does break in an unfortunate way, so only the 5 figure is visible to the user.

### Media

Clear DASH > HD

But this issue occurs with every video file.

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to dev.exoplayer@gmail.com after filing this issue.

cemrich:
This style change does not affect the `Theme.AppCompat.NoActionBar` theme applied to the demo app.

You may test this patch by applying the `Theme.Material3.Dark` as parent theme to `PlayerTheme` inside `demos/main/src/main/res/values/styles.xml`. The default 15 second forward text should not be cut off any longer.

Tested on api levels 34 and 26.
 
Fixes #547

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -116,6 +116,9 @@
     *   Fix issue where forward and rewind buttons are not visible when used
         with Material Design in a BottomSheetDialogFragment
         ([#511](https://github.com/androidx/media/issues/511)).
+    *   Fix issue where the numbers in the fast forward button of the
+        `PlayerControlView` were misaligned
+        ([#547](https://github.com/androidx/media/issues/547)).
 *   Downloads:
 *   OkHttp Extension:
 *   Cronet Extension:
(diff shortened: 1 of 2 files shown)


marcbaechinger:
I'm going to send this for internal review now. Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks!

Issue #927, opened 2023-12-28
What is media3 MediaMetadata.displayTitle role?


yt8492:
Before media3, Now Playing Card media title refers METADATA_KEY_DISPLAY_TITLE like this.

```kotlin
val mediaMetadataCompatBuilder = MediaMetadataCompat.Builder().apply {
      putString(MediaMetadataCompat.METADATA_KEY_DISPLAY_TITLE, playable.title)
      putString(MediaMetadataCompat.METADATA_KEY_DISPLAY_SUBTITLE, playable.subtitle)
      putString(MediaMetadataCompat.METADATA_KEY_ART_URI, playable.artUri)
}
```

But in media3, when I set displayTitle to MediaMetadata, it seems not working. here is reproducible example.
https://github.com/yt8492/media/commit/9c959e21bccdfedb8925844d071d8a8bc7cb360d
<image src="https://github.com/androidx/media/assets/39693306/2516fa21-04e2-450d-8159-f7c724c76d65" width="300" />

Is this behavior correct? If true, what is displayTitle role in media3?

marcbaechinger:
Thanks for your report.

Media3 is using `MediaMetadata.title` and converts from and to legacy keys [here](https://github.com/androidx/media/blob/main/libraries/session/src/main/java/androidx/media3/session/LegacyConversions.java#L441-L445) and [here](https://github.com/androidx/media/blob/main/libraries/session/src/main/java/androidx/media3/session/LegacyConversions.java#L575-L578).

So as a Media3 user you can just use `MediaMetadata.title`. This remove some ambiguity for Media3 users and you are still backwards compatible for users of the legacy or platform controller API that need to check both variants.

This behavior seems to be sensible to me. Are you using different values for `displayTitle` and `title` in your app and if you do, what is the reason to do so?

yt8492:
@marcbaechinger 
Thank you for the answer.

> Are you using different values for displayTitle and title in your app and if you do, what is the reason to do so?

In my case, I only set a value for displayTitle and not for title. There is no particular reason, I simply replaced from ExoPlayer2 to Media3.

This worked fine most of the cases, but some devices (Xperia, OPPO, etc.) were crashing because they could not get the title from the MediaMetadata.

```
17:19:20.643  E  java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object android.os.Bundle.get(java.lang.String)' on a null object reference
--------- beginning of crash
[...]

PR #928, opened 2023-12-28
Show MediaMetadata.displayTitle as a Now Playing Card media title

yt8492:


rohitjoins:
Dependent on resolution of https://github.com/androidx/media/issues/927

PR #941, opened 2024-01-02
Cc 708 parser fix

datdoantelus:
ANSI CTA-708-E S-2023 specification indicates that row and column lock values should be ignored in DefineWindow command and decoders should assume rows and columns are locked.

8.4.7 Window Row and Column Locking
Prior to CEA-708-D, parameters row lock (rl) and column lock (cl) in the DefineWindow (DFx) command were used for control of window row and column locking, providing alternatives in growing and shrinking caption windows. This standard specifies a single method for window growing and shrinking, and in Section 8.10.5.2 these parameters are reserved. Future versions of this standard may allow use of these parameters, but decoders compliant with the present version of CEA-708 shall ignore the values, and assume rows and columns are locked. Future versions of this standard may allow the optional decoder implementation of windows with unlocked rows and columns, and may use these syntax elements.

The change is to make Cea708Parser follow the specification.


datdoantelus:
wrong branch

PR #972, opened 2024-01-14
feat: show video track selectors on PlayerControlView

xeinebiu:
This PR adds functionality on PlayerControlView to display an extra setting on SettingsPopup (Window) where user is able to select another VideoTrackSelection similar to AudioTrackSelection.

image

image

### DIFF
--- a/libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
+++ b/libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
@@ -265,6 +265,8 @@ public class PlayerControlView extends FrameLayout {
   private static final int SETTINGS_PLAYBACK_SPEED_POSITION = 0;
   private static final int SETTINGS_AUDIO_TRACK_SELECTION_POSITION = 1;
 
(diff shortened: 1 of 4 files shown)


microkatz:

libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
```diff
@@ -193,7 +193,8 @@
 public class PlayerControlView extends FrameLayout {
 
   static {
-    MediaLibraryInfo.registerModule("media3.ui");
+    MediaLibraryInfo.registerModule("media3.ui"); // copybara:media3-only
+    // copybara:exo-only MediaLibraryInfo.registerModule("goog.exo.ui");
```
xeinebiu, line 197: Nice catch.

I don't remember removing these comments.

microkatz: Its okay. It was not you actually. These comments I should not have readded here. But thank you for noticing!

PR #973, opened 2024-01-14
Extract audio type from TS descriptors

jan-varecka-signageos-io:
The audio type of the stream, as defined by ISO/IEC 13818-1 in section 2.6.18, can be mapped to RoleFlags. Then this information can be used by the user for better selection of audio stream.

Real UC: source contains two audio streams, both are in the same language. One of them is intended for visually impaired viewer. With information about audio type, the user can better decide which stream he wants to hear.
Screenshot_20240115_005411

As audio type can be mapped to RoleFlags, I have decided not to extends Format class.


rohitjoins:
Have moved the `AudioType` interface from `C` to `TsUtil` as this information is only related to TS files. 

PR #1006, opened 2024-01-19
Fix PGS subtitles rendered with border and no fill color 

SimonHung:
bug: https://github.com/moneytoo/Player/issues/413
sample file: https://drive.google.com/file/d/1DLc9gVfyOUh65tWNOIWi2RdhNP6FMi_h

because '0' means it's an index to the palette table, not an RGBA value = 0


tonihei:
Thanks for the PR! Could you check the whether you signed the CLA (see robo comment above)? Otherwise we can't look into this pull request I'm afraid.

SimonHung:
OK, I have signed the CLA

icbaker:
Can you please check again? The CLA check is still failing, suggesting that it can't match up the email address on your commit with the CLA submission you made.

icbaker:
Please also update this to merge into the `main` branch, rather than `release`.

PR #1011, opened 2024-01-22
Fix HLS DTS Express bit-rate unknown issue.

cedricxperi:
For HLS-fMP4 and HLS-TS, the bit-rate of the DTS Express stream is not in the manifest file. The bit-rate has no value at the point of AudioTrack Initialisation. This will lead to a computed direct passthrough playback buffer size of 2250000 bytes. Some TVs (E.g. Xiaomi A Pro and Q2) do not support such a big buffer for direct passthrough playback. It will crash at the point of AudioTrack Initialization. Here we set the unknown bit-rate to a known peak bit-rate for DTS Express streams so that a smaller buffer size will be computed based on it.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java
@@ -76,7 +76,6 @@ import java.lang.annotation.Target;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.ArrayDeque;
-import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 import org.checkerframework.checker.nullness.qual.RequiresNonNull;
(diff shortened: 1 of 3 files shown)


tonihei:
Thanks for the contribution! The general idea makes sense, but given the symptoms I think it should be fixed in `DefaultAudioTrackBufferSizeProvider.getMaximumEncodedRateBytesPerSecond` instead.

cedricxperi:
@tonihei Thanks for the tip. Will take a look early next week.

cedricxperi:
@tonihei  The problem with doing this at getMaximumEncodedRateBytesPerSecond(@C.Encoding int encoding) is that the method only excepts encoding as a parameter. Due to Direct passthrough playback requirements, both DTS Express and DTS Digital Surround could be re-mapped to ENCODING_DTS. There is no way to distinguish the 2 DTS formats when inside getMaximumEncodedRateBytesPerSecond(). We only want to adjust the maximum bitrate for DTS Express and not DTS Digital Surround. What do you think?  Thanks.

cedricxperi:
@tonihei Thanks for the suggestions. :) This would need to work for HLS-TS as well, so I took your suggestion of adding a special case for DefaultAudioSink rather than do this at the AtomParsers.  Please see my latest commit.
Thanks.

tonihei:
Thanks, I'll work on merging this change.

PR #1015, opened 2024-01-24
Update PsshAtomUtil to parse version 1 and Expose inner class

kamaroyl:
This change addresses some of the todos in the PsshAtomUtil, exposing the inner class as well as add parsing and marshaling of v1 PSSH atoms.

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/PsshAtomUtil.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/PsshAtomUtil.java
@@ -72,10 +72,11 @@ public final class PsshAtomUtil {
     if (data != null && data.length != 0) {
       psshBox.putInt(data.length);
       psshBox.put(data);
-    } // Else the last 4 bytes are a 0 DataSize.
+    } else {
+      psshBox.putInt(0);
+    }
     return psshBox.array();
   }
-
   /**
    * Returns whether the data is a valid PSSH atom.
    *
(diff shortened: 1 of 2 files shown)


hichemcesar24 (approved):

tonihei:
Thanks for the contribution! 

Could you update your change to:
 - Only contain the minimal updates needed to add the additional `keyIds`. The pull request also renames some variables and moves code around and it would be good to keep it to its functional important minimum.
 - Rebase the change to target the `main` branch (not `release`).

In addition, could you explain how the data is used? The new code parses the `keyIds` but it would be helpful to see how this is useful in the end (e.g. do you expect to send more PRs to use the data elsewhere?)

kamaroyl:
Hi @tonihei,

Thanks for the feedback, I've made the changes. 

I'm currently using the v1 parsing to filter out tracks that have unused keys in a separate fork. I have it working with DASH but not HLS yet. I'm not sure how you all planned to deal with missing keys in multi-key content, but thought this might be a useful piece for you all.

tonihei:
>  I'm not sure how you all planned to deal with missing keys in multi-key content, but thought this might be a useful piece for you all.

I think we usually expect to get all keys from the manifest/playlist files. @icbaker can potentially comment further in case they are aware of an existing feature request or gap to discuss this further.  @kamaroyl If you think there is a piece missing that should be in the library, please also feel free to file a new feature request or send pull requests as needed.

Outside of that, I think the pull request looks fine and I'll work on merging it.

tonihei:
Note that re-added the scheme data parsing even for version 1 (because the spec allows it in its technical box definition) and I also made the constructor package-private so that only the parsed fields are visible. I hope that still fits you desired use case.

kamaroyl:
@tonihei there are several tickets in the previous repo with asks to check the status of the keys before playback (e.g. https://github.com/google/ExoPlayer/issues/4825). It seems like the official workaround is to try and restrict the streams by other properties, but that can be cumbersome depending on the varieties of videos.

kamaroyl:
> Note that re-added the scheme data parsing even for version 1 (because the spec allows it in its technical box definition) and I also made the constructor package-private so that only the parsed fields are visible. I hope that still fits you desired use case.

That works great, and fits the rest of the project better. I waffled on whether or not to include parsing of schema data since https://www.w3.org/TR/eme-initdata-cenc/ stated "the data size shall be zero," but it seems like it wouldn't hurt anything either way.

PR #1024, opened 2024-01-25
Fix support link in `ExoPlayerImpl#verifyApplicationThread()`

MGaetan89:
`ExoPlayerImpl#verifyApplicationThread()` throws an `IllegalStateException` and log a message if the player is accessed from the wrong thread. The message contains a link to the documentation to help developers understand the issue. However, the link is no longer valid. This PR updates it to point to the right location.


icbaker:
Sorry for the disruption. The link was accidentally broken on 2024-01-05 (internal cl/596016107) and fixed again on 2024-01-18 (internal cl/599521447) when we noticed the problem. So the existing/old message can remain I think, and I'm going to close this PR - but thanks for taking the time to fix it!

PR #1025, opened 2024-01-25
Player view to show post-processed resolution instead of the input one

dsparano:
_Problem_
If a video effect is applied to the video, or an "enhancement" such as LCEVC, is applied, the on screen player view debug text still shows the "input" resolution and pixel aspect ratio and not the one after post-processing. This can be easily reproduced with effects such as `Crop` or `ScaleAndRotateTransformation` which change the output resolution.

_Root cause_
The resolution and pixel aspect ratio shown are taken from the input Format object, which is a property of the stream and not of the video decoding and post processing applied.

_Proposed solution_
Take with, height and pixel aspect ratio from the player's VideoSize, which does get updated in case of post-processing.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/util/DebugTextViewHelper.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/util/DebugTextViewHelper.java
@@ -22,6 +22,7 @@ import androidx.annotation.Nullable;
 import androidx.media3.common.ColorInfo;
 import androidx.media3.common.Format;
 import androidx.media3.common.Player;
+import androidx.media3.common.VideoSize;
 import androidx.media3.common.util.Assertions;
 import androidx.media3.common.util.UnstableApi;
 import androidx.media3.exoplayer.DecoderCounters;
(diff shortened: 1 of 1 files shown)


dsparano:
Found `TODO b/292111083 - Remove the field and trigger the callback on every video size change.` in VideoFrameRenderControl and done it as it's needed to update the reported VideoSize when switching among video tracks.

claincly:
We recently changed the way of size notification and decided to temporarily disable the size notification to the app, in https://github.com/androidx/media/commit/dcae49a561d3dd0a67a44a913c063b0232bbc445

So unfortunately this fix won't work before we bring back the size notification.

claincly:
To clarify on this PR -

- I think this PR makes sense on its own and can be merged
- I wonder what is your intended use case though? Do you want to get the video size after being processed by `VideoFrameProcessor`? If that's the case, you won't be getting the correct size (I think) even after merging this PR with the newest code in `main`

dsparano:
@claincly I think this PR would be useful for at least two use cases:
- video effects (see above)
- use of decoder_lcevc, in which case the output (enhanced) resolution would be correctly set in `DebugTextViewHelper` as opposed to the base codec resolution as it is currently from `main`

claincly:
Sounds good, one thing related to pushing the PR:

dsparano:
@claincly no prob, I've given you write permission so you can commit into our originating branch

PR #1031, opened 2024-01-26
Seek at a chunk level when seeking to chunk start positions

garethfenn:
This avoids issues that can arise due to the slight discrepancies between chunk start times (obtained from the manifest of segment index) and the timestamps of the samples contained within those chunks.

This PR adds the equivalent logic of google/ExoPlayer@a4114f59b6897a28dd10cff2f11311dc04fc9363, for HLS TS streams. I have copied the description from that commit.

It is apparent when SeekParameters are used, added for HLS in google/ExoPlayer#9536, that the video sample chosen is not the first keyframe in the chunk, rather the last keyframe of the preceding chunk.

It is ultimately down to using "findSampleBefore()":
https://github.com/androidx/media/blob/f6fe90f30ba022c7e04e14a3dc5c28c568a4d1e2/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/SampleQueue.java#L978

It can be reproduced using the Exoplayer Demo "HLS Apple multivariant playlist advanced (TS)" stream. The first video keyframe PTS in the chunks are offset back by 16.666ms, hence not snapped to. The previous keyframe is used, 
some 1983.334ms out of sync with the chunk PTS and that of the audio PTS.

As far as I can tell, it is only visually apparent for tunneled playback where HW AV sync is used, resulting in a jarring acceleration of video decode to catch up with audio.


tonihei:
Thanks for the contribution! The general idea sounds great, we probably need to do some testing to verify it works as expected, but otherwise I think we can merge this change. 

Can I ask you to do two changes though?
 - The PR needs to target the `main` branch (not `release`)
 - `SampleQueue.seekTo(index)` starts supplying samples from this specific index without checking whether it is a keyframe. This works for DASH because chunks are by spec required to start with a keyframe. HLS chunks have no such requirement though unless they are specifically marked as independent. So we should check whether `mediaPlaylist.hasIndependentSegments` is true and only then take the go-to-first-sample-of-chunk shortcut.

garethfenn:
Thanks @tonihei. Apologies, I'll correct the target branch on Monday.

Regarding HLS independent chunks, I'd first like to highlight that there is already a check here:

https://github.com/androidx/media/blob/f6fe90f30ba022c7e04e14a3dc5c28c568a4d1e2/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/HlsChunkSource.java#L293

The result is SeekParameters are effectively disabled for content that doesn't declare the independent segments HLS tag. getAdjustedSeekPositionUS() does not adjust the seek position to the start of the chunk, and so this added code is not used. That being said, let me know if a secondary check is still required.

tonihei:
> The result is SeekParameters are effectively disabled for content that doesn't declare the independent segments HLS tag. getAdjustedSeekPositionUS() does not adjust the seek position to the start of the chunk, and so this added code is not used. That being said, let me know if a secondary check is still required.

Someone could still manually seek to exactly the start of a chunk even without adjustments via seek parameters. Given chunk durations are often 'round' numbers, this isn't too unlikely actually.

garethfenn:
> Someone could still manually seek to exactly the start of a chunk even without adjustments via seek parameters. Given chunk durations are often 'round' numbers, this isn't too unlikely actually.
Sure. I've added a conditional on the chunkSource having independent segments. This required adding a getter to HlsChunkSource. Let me know if this was a reasonable thing to do.

tonihei:
That should work, thanks! I'm going to test this a bit to see what happens.

tonihei:
Also added a test case for this. Thanks again, we'll merge this change soon!

Issue #1029, opened 2024-01-26
If instances from different sequences are overlapping how are they mixed? Which one get's to be on top of the other? How to manipulate z-order?


AradiPatrik:
In Composition.java, there's a comment indicating that instances of MediaItem from different sequences will be mixed in the output when they overlap in time. However, based on my tests involving 5 overlapping media items with transparency, I've noticed that the z-order appears to be undefined. The stacking order of the videos seems to change randomly with each execution. Here's the relevant comment from the code:

```
/**
 * The {@link EditedMediaItemSequence} instances to compose.
 *
 * <p>{@link MediaItem} instances from different sequences that are overlapping in time will be
 * mixed in the output.
 */
 ```
Given this behavior, I'm seeking a way to control the z-order of these videos. Specifically, is there a method or technique to set or manipulate the z-order of `MediaItem` instances within `EditedMediaItemSequence` ore entire `EditedMediaItemSequences` to ensure a consistent stacking order in the output?

dway123:
The default implementation of our VideoCompositor is DefaultVideoCompositor. z-order should be [defined](https://github.com/androidx/media/blob/b930b40a16c06318e43c81771fa2b1024bdb3f29/libraries/effect/src/main/java/androidx/media3/effect/DefaultVideoCompositor.java#L56-L68) by the order of the EditedMediaItemSequence inserted into the Composition's [`builder`](https://github.com/androidx/media/blob/b930b40a16c06318e43c81771fa2b1024bdb3f29/libraries/transformer/src/main/java/androidx/media3/transformer/Composition.java#L63-L76). 

Unfortunately, we don't have a way to customize the order, other than using the order of EditedMediaItemSequences in the Composition.

> The stacking order of the videos seems to change randomly with each execution.

This is surprising to me. I would expect the stacking order to be consistent in each execution. I guess you are getting this while using Transformer with a Composition with 5 overlapping mediaItems? If so, I wonder if Transformer is loading EditedMediaItemSequences in a nondeterministic order (which is a bug on our side)

AradiPatrik:
Hi @dway123, @droid-girl,

Thank you for your insights on the issue. As per our discussion, I've created a sample project that consistently reproduces the z-order inconsistency issue we've been observing with the Composition and DefaultVideoCompositor.

Repository Link:
[Media3Issue Sample Project](https://github.com/AradiPatrik/Media3Issue)

Brief Description of the Repository:
[...]

PR #1055, opened 2024-01-31
Fix indeterminate z-order of EditedMediaItemSequences

AradiPatrik:
## Root Cause:
The indeterminate z-order issue arises due to the non-deterministic loading of assets. As each asset finishes loading, DefaultVideoCompositor.registerInput is invoked, linking the z-order to the unpredictable asset loading sequence. Consequently, the final z-order in the output is uncertain, as it's contingent on the load completion sequence of the assets.

## Proposed Fix:
To resolve this, I've explored two options:

- *Option A*: Modify TransformerInternal.java to aggregate all loaded assets first, then sequentially invoke registerInput in the intended order.
- *Option B*: Introduce a sequenceIndex parameter to registerInput, decoupling the z-order from the invocation sequence and aligning it with this index instead.

## Implemented Solution:
I opted for Option B due to its straightforward implementation. Post-fix, the z-order issue was no longer reproducible, indicating the effectiveness of this solution.

## Links
- [Original issue](https://github.com/androidx/media/issues/1029)
- [Simple Sample Project for reliable reproduction](https://github.com/AradiPatrik/Media3Issue)

I'm open to feedback and willing to assist further to ensure a robust resolution to this z-order inconsistency.

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/VideoGraph.java
+++ b/libraries/common/src/main/java/androidx/media3/common/VideoGraph.java
@@ -76,7 +76,6 @@ public interface VideoGraph {
    * <p>If the method throws, the caller must call {@link #release}.
    *
    * @param sequenceIndex The sequence index of the input which can aid ordering of the inputs.
-   *
    * @return The id of the registered input, which can be used to get the underlying {@link
    *     VideoFrameProcessor} via {@link #getProcessor(int)}.
    */
--- a/libraries/common/src/main/java/androidx/media3/common/VideoGraph.java
+++ b/libraries/common/src/main/java/androidx/media3/common/VideoGraph.java
@@ -76,10 +76,8 @@ public interface VideoGraph {
    * <p>If the method throws, the caller must call {@link #release}.
    *
    * @param sequenceIndex The sequence index of the input which can aid ordering of the inputs.
-   * @return The id of the registered input, which can be used to get the underlying {@link
-   *     VideoFrameProcessor} via {@link #getProcessor(int)}.
    */
-  int registerInput(int sequenceIndex) throws VideoFrameProcessingException;
+  void registerInput(int sequenceIndex) throws VideoFrameProcessingException;
 
   /**
    * Returns the {@link VideoFrameProcessor} that handles the processing for an input registered via
--- a/libraries/common/src/main/java/androidx/media3/common/VideoGraph.java
+++ b/libraries/common/src/main/java/androidx/media3/common/VideoGraph.java
@@ -16,6 +16,7 @@
 
 package androidx.media3.common;
 
+import androidx.annotation.IntRange;
 import androidx.annotation.Nullable;
 import androidx.media3.common.util.UnstableApi;
 
@@ -73,20 +74,23 @@ public interface VideoGraph {
    * <p>A underlying processing {@link VideoFrameProcessor} is created every time this method is
    * called.
    *
+   * <p>All inputs must be registered before rendering frames to the underlying
+   * {@link #getProcessor(int) VideoFrameProcessor}.
+   *
    * <p>If the method throws, the caller must call {@link #release}.
    *
-   * @param sequenceIndex The sequence index of the input which can aid ordering of the inputs. The
-   *     index must start from 0.
+   * @param inputIndex The index of the input which could be used to order the inputs.
+   *     The index must start from 0.
    */
-  void registerInput(int sequenceIndex) throws VideoFrameProcessingException;
+  void registerInput(@IntRange(from = 0) int inputIndex) throws VideoFrameProcessingException;
 
   /**
    * Returns the {@link VideoFrameProcessor} that handles the processing for an input registered via
-   * {@link #registerInput(int)}. If the {@code sequenceIndex} is not {@linkplain
+   * {@link #registerInput(int)}. If the {@code inputIndex} is not {@linkplain
    * #registerInput(int) registered} before, this method will throw an {@link
    * IllegalStateException}.
    */
-  VideoFrameProcessor getProcessor(int sequenceIndex);
+  VideoFrameProcessor getProcessor(int inputIndex);
 
   /**
    * Sets the output surface and supporting information.
--- a/libraries/common/src/main/java/androidx/media3/common/VideoGraph.java
+++ b/libraries/common/src/main/java/androidx/media3/common/VideoGraph.java
@@ -75,7 +75,8 @@ public interface VideoGraph {
    *
    * <p>If the method throws, the caller must call {@link #release}.
    *
-   * @param sequenceIndex The sequence index of the input which can aid ordering of the inputs.
+   * @param sequenceIndex The sequence index of the input which can aid ordering of the inputs. The
+   *     index must start from 0.
    */
   void registerInput(int sequenceIndex) throws VideoFrameProcessingException;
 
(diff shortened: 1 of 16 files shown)


dway123:
Thank you for the quick PR! I'll try to take a look in the next few days. Also, cc @claincly who is pretty familiar with these files too

dway123:
I have less familiarity than @claincly here so I'll defer to him, but just noting that without as much context, I think the general option B selected in this PR sounds good to me (haven't looked to make sure the implementation style/architecture makes sense though). Option A would also require adding additional state and/or blocking, so I'd prefer B over A.

libraries/effect/src/main/java/androidx/media3/effect/VideoCompositor.java
```diff
@@ -48,8 +48,11 @@ interface Listener {
   /**
    * Registers a new input source, and returns a unique {@code inputId} corresponding to this
    * source, to be used in {@link #queueInputTexture}.
+   *
+   * @param sequenceIndex The sequence index of the input source, which is can be used to determine
+   *     the order of the input sources.
    */
-  int registerInputSource();
+  int registerInputSource(int sequenceIndex);
```
claincly, line 55: a high level comment: it's strange that this method takes in an ID and returns another ID, where those IDs can potentially be the same. In the actual impl(in `DefaultVideoCompositor`) those two indices match and that's what fixes the z-order problem. 

I feel the approach could be a bit top down, that this method just takes a `sequenceIndex` but doesn't return any index. The passed-in index ultimately defines the z-order, wdyt? @dway123 please also take a look and see if you agree.

AradiPatrik: @claincly This approach sounds good to me!

Let me change the code according to this design.

AradiPatrik: @dway123 @claincly Implemented the top down approach for id passing as discussed.

libraries/effect/src/main/java/androidx/media3/effect/MultipleInputVideoGraph.java
```diff
@@ -134,7 +134,7 @@ protected MultipleInputVideoGraph(
   @Override
   public void initialize() throws VideoFrameProcessingException {
     checkState(
-        preProcessors.isEmpty()
+        preProcessors.size() == 0
```
dway123, line 137: isn't isEmpty equivalent to size() == 0?

libraries/transformer/src/main/java/androidx/media3/transformer/AudioSampleExporter.java
```diff
@@ -94,7 +94,7 @@ public AudioSampleExporter(
   }
 
   @Override
-  public AudioGraphInput getInput(EditedMediaItem editedMediaItem, Format format)
+  public AudioGraphInput getInput(EditedMediaItem editedMediaItem, Format format, int sequenceIndex)
```
dway123, line 97: @claincly  not sure but does Audio also need this information?

dway123 (commented):
(some small comments at a glance :P )

libraries/effect/src/main/java/androidx/media3/effect/VideoCompositor.java
dway123: @claincly ptal?

libraries/effect/src/main/java/androidx/media3/effect/MultipleInputVideoGraph.java
```diff
@@ -252,16 +251,15 @@ public void onError(VideoFrameProcessingException exception) {
...
                 });
-    preProcessors.add(preProcessor);
-    return videoCompositorInputId;
+    preProcessors.put(sequenceIndex, preProcessor);
   }
 
   @Override
   public VideoFrameProcessor getProcessor(int inputId) {
-    checkState(inputId < preProcessors.size());
+    checkState(preProcessors.indexOfKey(inputId) >= 0);
```
claincly, line 262: nit: Can use `Util.contains(preProcessors, inputId)` here.

libraries/common/src/main/java/androidx/media3/common/VideoGraph.java
```diff
@@ -75,14 +75,13 @@ interface Listener {
    *
    * <p>If the method throws, the caller must call {@link #release}.
    *
-   * @return The id of the registered input, which can be used to get the underlying {@link
-   *     VideoFrameProcessor} via {@link #getProcessor(int)}.
+   * @param sequenceIndex The sequence index of the input which can aid ordering of the inputs.
```
claincly, line 78: It seems it's easier if we require sequence indices start from 0, so please add something like "The index must start from zero"

libraries/effect/src/main/java/androidx/media3/effect/SingleInputVideoGraph.java
```diff
@@ -38,7 +38,7 @@
 @UnstableApi
 public abstract class SingleInputVideoGraph implements VideoGraph {
 
-  /** The ID {@link #registerInput()} returns. */
+  /** The ID {@link #registerInput(int)} returns. */
```
claincly, line 41: nit: `The index of the only {@linkplain #registerInput(int) registered} input.`

libraries/effect/src/main/java/androidx/media3/effect/VideoCompositor.java
```diff
@@ -48,8 +48,11 @@ interface Listener {
   /**
    * Registers a new input source, and returns a unique {@code inputId} corresponding to this
    * source, to be used in {@link #queueInputTexture}.
+   *
+   * @param sequenceIndex The sequence index of the input source, which is can be used to determine
```
claincly, line 52: Please reword: `which is used to determine the order of the input sources`

libraries/transformer/src/main/java/androidx/media3/transformer/AudioSampleExporter.java
claincly: I'm not sure either, but I think it's OK since it's from the `SampleExporter` API change.

libraries/transformer/src/main/java/androidx/media3/transformer/SampleExporter.java
```diff
@@ -64,10 +64,11 @@ public SampleExporter(Format firstInputFormat, MuxerWrapper muxerWrapper) {
    *
    * @param editedMediaItem The initial {@link EditedMediaItem} of the input.
    * @param format The initial {@link Format} of the input.
+   * @param sequenceIndex The sequence index of the input.
```
claincly, line 67: supernit: `The index of the input sequence`?

claincly (changes requested):
Look good overall!

AradiPatrik:
@claincly I implemented the changes you requested. Please take a look and resolve it if it's OK. Let me know if there's anything else we need to change before this PR can be merged.

claincly:

AradiPatrik:
@claincly Any updates on this? Can I assist in any way?

claincly:
Hey Aradi, we are discussing it internally as it's an API change, will keep you in the loop!

yuriikonovalov:
@claincly could you provide some information (if it's possible) when it's potentially can be merged and released so that this fix is available to use in production?

PR #1117, opened 2024-02-22
Fix issue where padding was not skipped when reading odd-sized chunks from WAV files

loliball:
If a chunk body has an odd number of bytes, it must be followed by a padding byte with value 0. In other words, a chunk must always occupy an even number of bytes in the file. The padding byte should not be counted in the chunk header’s size field. For example, if a chunk body is 17 bytes in size, the header’s size field should be set to 17, even though the chunk body occupies 18 bytes (17 bytes of data followed by the padding byte).

https://wavefilegem.com/how_wave_files_work.html


loliball:
[Here is test wav file test.zip](https://github.com/androidx/media/files/14380811/test.zip)

Issue #1001, opened 2024-01-18
Analytics onLoad events for HttpMediaDrmCallback requests


stevemayhew:
### Use case description
An often significant delay for first frame with channel tunes occurs because of delays from the DRM License Server key or provisioning requests, it would be nice to have analytics events for these.

### Proposed solution
The standard `MediaSourceEventListener` report details for `DATA_TYPE_DRM` for fetches from `EXT-X-KEY` with an actual key URL, However the same level of details is missing from the `DrmSessionEventListener` events.

We could:
1. Add load statistics to the existing `drmKeysLoaded` event 
2. Add two new events (`drmKeyRequestStarted`, `drmKeyRequestCompleted`)

### Alternatives considered
Adding logging to the `HttpMediaDrmCallback`,  this is simple but not as useful as a hook that allowed using `AnalyticsListener` to post the data to operational logging.

stevemayhew:
@icbaker I'm curious why `DefaultDrmSession.RequestHandler` was even implemented rather than using the existing `com.google.android.exoplayer2.upstream.Loader`. which does largely the same thing with a richer set of API calls to observe and manage the load.

icbaker:
> ### Alternatives considered
>
> Adding logging to the `HttpMediaDrmCallback`, this is simple but not as useful as a hook that allowed using `AnalyticsListener` to post the data to operational logging.

Not sure if this is what you meant, but to be explicit: you can add a `TransferListener` to the `DataSource.Factory` you pass to `HttpMediaDrmCallback`, and you should get some callbacks including URLs etc.

----

> @icbaker I'm curious why `DefaultDrmSession.RequestHandler` was even implemented rather than using the existing `com.google.android.exoplayer2.upstream.Loader`.

Good question - I'm afraid I don't know the reasoning at the time, but I suspect it might have been felt that the `Loader` machinery was a bit heavyweight for "just" making an HTTP request that you expect to complete fairly quickly (vs e.g. keeping a long-running HTTP request open while you load a whole MP4 movie).

---

> We could:
> 
> 1. Add load statistics to the existing `drmKeysLoaded` event
> 2. Add two new events (`drmKeyRequestStarted`, `drmKeyRequestCompleted`)

I like (1) more than (2), because we could also include additional info like the type of key load (offline, renewal, etc).

[...]

PR #1134, opened 2024-02-26
Adds load info to `DrmSessionEventListeneronDrmKeysLoaded()`

stevemayhew:
This is a first pass at what we talked about in issue https://github.com/androidx/media/issues/1001

* added the wrapper class `KeyLoadInfo` to embody the requests required  to fetch the key
* added `KeyLoadInfo` to the `DrmSessionEventListener.drmKeysLoaded()` method
* Deprecated `DrmSessionEventListener.onDrmKeysLoaded(int, MediaPeriodId)`
* added new `onDrmKeysLoaded()` with the `KeyLoadInfo`, setup so both method are called

Next steps are to add it to `AnalyticsListener` and update any broken test cases possibly add some new ones

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
@@ -21,6 +21,7 @@ import static java.lang.Math.min;
 
 import android.annotation.SuppressLint;
 import android.media.NotProvisionedException;
+import android.net.Uri;
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.os.Looper;
@@ -36,6 +37,7 @@ import androidx.media3.common.util.Consumer;
 import androidx.media3.common.util.CopyOnWriteMultiset;
 import androidx.media3.common.util.Log;
 import androidx.media3.common.util.Util;
+import androidx.media3.datasource.DataSpec;
 import androidx.media3.decoder.CryptoConfig;
 import androidx.media3.exoplayer.analytics.PlayerId;
 import androidx.media3.exoplayer.drm.ExoMediaDrm.KeyRequest;
@@ -149,6 +151,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
   private byte @MonotonicNonNull [] offlineLicenseKeySetId;
 
   @Nullable private KeyRequest currentKeyRequest;
+  @Nullable private KeyLoadInfo currentKeyLoadInfo;
   @Nullable private ProvisionRequest currentProvisionRequest;
 
   /**
@@ -489,6 +492,9 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
 
   private void postKeyRequest(byte[] scope, int type, boolean allowRetry) {
     try {
+      if (type == ExoMediaDrm.KEY_TYPE_STREAMING) {
+        currentKeyLoadInfo = new KeyLoadInfo();
+      }
       currentKeyRequest = mediaDrm.getKeyRequest(scope, schemeDatas, type, keyRequestParameters);
       Util.castNonNull(requestHandler)
           .post(MSG_KEYS, Assertions.checkNotNull(currentKeyRequest), allowRetry);
@@ -524,7 +530,8 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
           offlineLicenseKeySetId = keySetId;
         }
         state = STATE_OPENED_WITH_KEYS;
-        dispatchEvent(DrmSessionEventListener.EventDispatcher::drmKeysLoaded);
+        dispatchEvent(eventDispatcher -> eventDispatcher.drmKeysLoaded(currentKeyLoadInfo));
+        currentKeyLoadInfo = null;
       }
     } catch (Exception e) {
       onKeysError(e, /* thrownByExoMediaDrm= */ true);
@@ -644,6 +651,19 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
             break;
           case MSG_KEYS:
             response = callback.executeKeyRequest(uuid, (KeyRequest) requestTask.request);
+            if (currentKeyLoadInfo != null) {
+              String requestUrl = ((KeyRequest) requestTask.request).getLicenseServerUrl();
+              LoadEventInfo loadEventInfo =
+                  new LoadEventInfo(
+                      requestTask.taskId,
+                      new DataSpec.Builder().setUri(requestUrl).build(),
+                      Uri.parse(requestUrl),
+                      Collections.emptyMap(),
+                      SystemClock.elapsedRealtime(),
+                      /* loadDurationMs= */ SystemClock.elapsedRealtime() - requestTask.startTimeMs,
+                      ((byte []) response).length);
+              currentKeyLoadInfo.setMainLoadRequest(loadEventInfo);
+            }
             break;
           default:
             throw new RuntimeException();
@@ -699,6 +719,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
         // The error is fatal.
         return false;
       }
+      currentKeyLoadInfo.addRetryLoadRequest(loadEventInfo);
       synchronized (this) {
         if (!isReleased) {
           sendMessageDelayed(Message.obtain(originalMsg), retryDelayMs);
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DrmSessionEventListener.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DrmSessionEventListener.java
@@ -47,12 +47,20 @@ public interface DrmSessionEventListener {
       int windowIndex, @Nullable MediaPeriodId mediaPeriodId, @DrmSession.State int state) {}
 
   /**
+   * @deprecated Implement {@link #onDrmKeysLoaded(int, MediaPeriodId, KeyLoadInfo)} instead
+   */
+  @Deprecated
+  default void onDrmKeysLoaded(int windowIndex, @Nullable MediaPeriodId mediaPeriodId) {}
+
+  /**
    * Called each time keys are loaded.
    *
    * @param windowIndex The window index in the timeline this media period belongs to.
    * @param mediaPeriodId The {@link MediaPeriodId} associated with the drm session.
+   * @param keyLoadInfo The {@link KeyLoadInfo} with load info for the drm server request[s]
    */
-  default void onDrmKeysLoaded(int windowIndex, @Nullable MediaPeriodId mediaPeriodId) {}
+  default void onDrmKeysLoaded(
+      int windowIndex, @Nullable MediaPeriodId mediaPeriodId, @Nullable KeyLoadInfo keyLoadInfo) {}
 
   /**
    * Called when a drm error occurs.
@@ -177,12 +185,17 @@ public interface DrmSessionEventListener {
       }
     }
 
-    /** Dispatches {@link #onDrmKeysLoaded(int, MediaPeriodId)}. */
-    public void drmKeysLoaded() {
+    /** Dispatches {@link #onDrmKeysLoaded(int, MediaPeriodId)}.
+     * and {@link #onDrmKeysLoaded(int, MediaPeriodId, KeyLoadInfo)}*/
+    public void drmKeysLoaded(@Nullable KeyLoadInfo keyLoadInfo) {
       for (ListenerAndHandler listenerAndHandler : listenerAndHandlers) {
         DrmSessionEventListener listener = listenerAndHandler.listener;
         postOrRun(
-            listenerAndHandler.handler, () -> listener.onDrmKeysLoaded(windowIndex, mediaPeriodId));
+            listenerAndHandler.handler,
+            () -> {
+              listener.onDrmKeysLoaded(windowIndex, mediaPeriodId);
+              listener.onDrmKeysLoaded(windowIndex, mediaPeriodId, keyLoadInfo);
+            });
       }
     }
 
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyLoadInfo.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyLoadInfo.java
@@ -1,0 +1,26 @@
+package androidx.media3.exoplayer.drm;
+
+import androidx.media3.exoplayer.source.LoadEventInfo;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Encapsulates info for the sequence of load requests ({@link LoadEventInfo}, which were required
+ * to complete loading a DRM key
+ */
+public class KeyLoadInfo {
+  public LoadEventInfo loadEventInfo;
+  public final List<LoadEventInfo> retriedLoadRequests;
+
+  public KeyLoadInfo() {
+    retriedLoadRequests = new ArrayList<>();
+  }
+
+  void setMainLoadRequest(LoadEventInfo loadEventInfo) {
+    this.loadEventInfo = loadEventInfo;
+  }
+
+  void addRetryLoadRequest(LoadEventInfo loadEventInfo) {
+    retriedLoadRequests.add(loadEventInfo);
+  }
+}
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/OfflineLicenseHelper.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/OfflineLicenseHelper.java
@@ -144,7 +144,8 @@ public final class OfflineLicenseHelper {
     DrmSessionEventListener eventListener =
         new DrmSessionEventListener() {
           @Override
-          public void onDrmKeysLoaded(int windowIndex, @Nullable MediaPeriodId mediaPeriodId) {
+          public void onDrmKeysLoaded(int windowIndex, @Nullable MediaPeriodId mediaPeriodId,
+              @Nullable KeyLoadInfo keyLoadInfo) {
             drmListenerConditionVariable.open();
           }
 
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/CompositeMediaSource.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/CompositeMediaSource.java
@@ -26,6 +26,7 @@ import androidx.media3.common.util.Util;
 import androidx.media3.datasource.TransferListener;
 import androidx.media3.exoplayer.drm.DrmSession;
 import androidx.media3.exoplayer.drm.DrmSessionEventListener;
+import androidx.media3.exoplayer.drm.KeyLoadInfo;
 import java.io.IOException;
 import java.util.HashMap;
 
@@ -316,9 +317,10 @@ public abstract class CompositeMediaSource<T> extends BaseMediaSource {
     }
 
     @Override
-    public void onDrmKeysLoaded(int windowIndex, @Nullable MediaPeriodId mediaPeriodId) {
+    public void onDrmKeysLoaded(
+        int windowIndex, @Nullable MediaPeriodId mediaPeriodId, @Nullable KeyLoadInfo keyLoadInfo) {
       if (maybeUpdateEventDispatcher(windowIndex, mediaPeriodId)) {
-        drmEventDispatcher.drmKeysLoaded();
+        drmEventDispatcher.drmKeysLoaded(keyLoadInfo);
       }
     }
 
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/ServerSideAdInsertionMediaSource.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/ServerSideAdInsertionMediaSource.java
@@ -46,6 +46,7 @@ import androidx.media3.exoplayer.LoadingInfo;
 import androidx.media3.exoplayer.SeekParameters;
 import androidx.media3.exoplayer.drm.DrmSession;
 import androidx.media3.exoplayer.drm.DrmSessionEventListener;
+import androidx.media3.exoplayer.drm.KeyLoadInfo;
 import androidx.media3.exoplayer.source.BaseMediaSource;
 import androidx.media3.exoplayer.source.EmptySampleStream;
 import androidx.media3.exoplayer.source.ForwardingTimeline;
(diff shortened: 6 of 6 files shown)


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
```diff
@@ -644,6 +651,19 @@ public void handleMessage(Message msg) {
             break;
           case MSG_KEYS:
             response = callback.executeKeyRequest(uuid, (KeyRequest) requestTask.request);
+            if (currentKeyLoadInfo != null) {
+              String requestUrl = ((KeyRequest) requestTask.request).getLicenseServerUrl();
```
stevemayhew, line 655: This needs some work, the knowledge of URL's and such is really all private to the `MediaDrmCallback` so either we need to:

1. Ignore this and provide vacuous URL's (which, unless the `KeyRequest` has the data this is already true)
2. add methods to `MediaDrmInfo` to fill in the `LoadEventInfo`

Our use case does not care about the URL, simply the load times and retries.  So either way is good.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
```diff
@@ -699,6 +719,7 @@ private boolean maybeRetryRequest(Message originalMsg, MediaDrmCallbackException
         // The error is fatal.
         return false;
       }
+      currentKeyLoadInfo.addRetryLoadRequest(loadEventInfo);
       synchronized (this) {
```
stevemayhew, line 723: This might be useful if you want the trails of `LoadEventInfo`'s including redirects in the path.  The main `LoadEventInfo` should include all the time.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DrmSessionEventListener.java
```diff
@@ -46,13 +46,21 @@ default void onDrmSessionAcquired(int windowIndex, @Nullable MediaPeriodId media
...
   /**
    * Called each time keys are loaded.
    *
    * @param windowIndex The window index in the timeline this media period belongs to.
    * @param mediaPeriodId The {@link MediaPeriodId} associated with the drm session.
+   * @param keyLoadInfo The {@link KeyLoadInfo} with load info for the drm server request[s]
    */
-  default void onDrmKeysLoaded(int windowIndex, @Nullable MediaPeriodId mediaPeriodId) {}
+  default void onDrmKeysLoaded(
+      int windowIndex, @Nullable MediaPeriodId mediaPeriodId, @Nullable KeyLoadInfo keyLoadInfo) {}
```
stevemayhew, line 64: followed the pattern to deprecate then call the deprecated method.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
```diff
@@ -489,6 +492,9 @@ private long getLicenseDurationRemainingSec() {
 
   private void postKeyRequest(byte[] scope, int type, boolean allowRetry) {
     try {
+      if (type == ExoMediaDrm.KEY_TYPE_STREAMING) {
+        currentKeyLoadInfo = new KeyLoadInfo();
+      }
```
stevemayhew, line 497: This should probably just check for `mode != DefaultDrmSessionManager.MODE_RELEASE`.   That way we cover download requests too.

icbaker (changes requested):

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
stevemayhew: @icbaker, thanks all great ideas...  How about we just let the `MediaDrmCallback` built the object, most of the info is already in the `HttpMediaDrmCallback`, the other implementations `MediaDrmCallback ` don't have to change as the default (null return) makes sense for them.

```java

  /**
   * Get the {@link LoadEventInfo} for the last executed request.
   * This can be null if no load is performed to execute the request
   *
   * @return the {@link LoadEventInfo} or null if no load was performed
   */
  @Nullable default LoadEventInfo getLastLoadEventInfo() { return null; }
  
 ````

With this, the `LoadEventInfo` is easy to produce so it is completely internally consistent (redirect included).

BTW, a few WV License Proxies we work with use delayed re-directs to manage load (Apache Traffic Server's "thundering herd" for example)

File: libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java (hunk 2 above)
Comment (stevemayhew) [reply libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java:723]: see last comment, with that redirects will be reported correctly.

File: libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java (hunk 4 above)
Comment (stevemayhew) [reply libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java:497]: Yes, that sounds like a good idea.  We can start a list for what to add to `KeyLoadInfo` here.

File: libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyLoadInfo.java [hunk 5]
```diff
@@ -0,0 +1,26 @@
...
+
+import androidx.media3.exoplayer.source.LoadEventInfo;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Encapsulates info for the sequence of load requests ({@link LoadEventInfo}, which were required
+ * to complete loading a DRM key
+ */
+public class KeyLoadInfo {
```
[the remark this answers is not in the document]
stevemayhew: Absolutely, a pretty simple refactor and a good practice.   The `DefaultDrmSession` can hold on to the builder, as we don't need the actual object until the final event callback.  Final is a good idea too, I'll make these two changes.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
stevemayhew: See latest commit, adds this method.   With the `LoadEventInfo` produced in the object responsible for the actual load request it is completely consistant with what `Loader` would produce.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
stevemayhew: This is done in the latest commit, except for adding the `KeyLoadInfo` to the `drmKeysRemoved()` method.   I'm fine with doing this too, however then we should rename `KeyLoadInfo` to `KeyRequestInfo`.  I'm all for the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) thinking too ;-)   

But, seems most of the work will be rebasing all the files open for this change so it is probably now or never.

I'll follow your lead on it.

stevemayhew:
Unless we want to go for renaming `KeyLoadInfo` to `KeyRequestInfo` and moving forward with the `drmKeysRemoved()` update too... I would squash the commits and run the code beautifier on what we have.

I'm good either way.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/HttpMediaDrmCallback.java
```diff
@@ -182,7 +192,17 @@ private static byte[] executePost(
...
+          lastLoadEventInfo =
+              new LoadEventInfo(
+                  -1,   // note this is replaced with the actual taskId from the request
+                  originalDataSpec,
+                  dataSource.getLastOpenedUri(),
+                  dataSource.getLastResponseHeaders(),
+                  SystemClock.elapsedRealtime(),
+                  /* loadDurationMs= */ SystemClock.elapsedRealtime() - startTimeMs,
+                  ((byte []) response).length);
+            return response;
```
stevemayhew, line 205: Creating the `LoadEventInfo` in the actual `HttpMediaDrmCallback` addresses all the todo's with its correctness issues.   Also note now the `loadDurationMs` does not include any queuing time in the `RequestHandler` (so it's pure network and proxy overhead measured).

Lastly, `DataSource.getLastOpenedUri()` should get the redirected URI (if any), at least this is the contract for `DataSource.getUri()`

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyLoadInfo.java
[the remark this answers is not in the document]
stevemayhew: resolving, last commit converts to `Builder`.  Great idea, makes it much easier to add stuff to this object.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/HttpMediaDrmCallback.java
```diff
@@ -182,7 +192,17 @@ private static byte[] executePost(
       while (true) {
         DataSourceInputStream inputStream = new DataSourceInputStream(dataSource, dataSpec);
         try {
-          return Util.toByteArray(inputStream);
+          byte[] response = Util.toByteArray(inputStream);
+          lastLoadEventInfo =
+              new LoadEventInfo(
+                  -1,   // note this is replaced with the actual taskId from the request
```
icbaker, line 198: What do you think about plumbing an (optional) `taskId` into `ExoMediaDrm.KeyRequest` and `ExoMediaDrm.ProvisionRequest`? Then it would be available here and you wouldn't need to use this placeholder value.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/HttpMediaDrmCallback.java
icbaker: Looks generally good, thanks

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyLoadInfo.java
```diff
@@ -0,0 +1,57 @@
...
+   */
+  public final LoadEventInfo loadEventInfo;
+
+  /** If the load required multiple retries, the {@link LoadEventInfo} for each retry
+   */
+  public final List<LoadEventInfo> retriedLoadRequests;
+
+  /** The DRM {@link SchemeData} that identifes the loaded key
+   */
+  public final List<SchemeData> schemeDatas;
```
icbaker, line 50: please make these two `ImmutableList` and take defensive copies in the constructor below

You can use `ImmutableList.Builder` for `retriedLoadRequests` in the builder, and then `ImmutableList.copyOf` in the constructor will be a no-op because it does an `instanceof ImmutableList` check first.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyLoadInfo.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public KeyLoadInfo(Builder builder) {
```
icbaker, line 52: stylistic/consistency nit: Make this private, and add an explicit `public Builder buildUpon()` method if needed (though I don't think you do need it?)

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/MediaDrmCallback.java
```diff
@@ -44,4 +46,15 @@ byte[] executeProvisionRequest(UUID uuid, ProvisionRequest request)
    * @throws MediaDrmCallbackException If an error occurred executing the request.
    */
   byte[] executeKeyRequest(UUID uuid, KeyRequest request) throws MediaDrmCallbackException;
+
+  /**
+   * Get the {@link LoadEventInfo} for the last executed request.
+   * <p></p>
+   * Valid after a call to {@link #executeKeyRequest(UUID, KeyRequest)} or
+   * {@link #executeProvisionRequest(UUID, ProvisionRequest)}, either contains
+   * the load event info for that request or null if no load is performed to execute the request
```
icbaker, line 55: nit - extend this wording to something like "or null if no load was performed or this implementation does not support reporting load info"

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
icbaker: > I'm fine with doing this too, however then we should rename `KeyLoadInfo` to `KeyRequestInfo`

Yeah, I think we should do this rename and include key releasing - I agree it's going to be easier to do this now than later.

icbaker (changes requested):
Sorry this took me so long to get back to, I'll try harder not to leave it so long on the next round of comments!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/HttpMediaDrmCallback.java
stevemayhew: Yes, it is a bit ugly this way...  I stopped short because changing the `MediaDrmCallback.executeProvisionRequest()` signature, as it has lots of implementations (unless we add new methods, which is also kind of clumsy feeling.   What if we added the value to `ProvisionRequest` and `KeyRequest`

stevemayhew:
I'm on vacation next week @icbaker so it will take a bit to get to these changes.  Only one I need your input on is best path for the `taskId`

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyLoadInfo.java
stevemayhew: Yes, that is a good practice, what about simply encapsulate access to the fields with methods that return `ImmutableList`?

Feel free to make a branch with the change you want and push a commit, you should have write access to this fork.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/HttpMediaDrmCallback.java
stevemayhew: @icbaker  this may fall into the category of perfect is the enemy of done category, but yes... It looks distatesful.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyLoadInfo.java
stevemayhew: Again, made it private in the renamed class.   Sorry for this, should have fixed it first then renamed rather than rename then fix.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyLoadInfo.java
```diff
@@ -9,18 +10,48 @@
...
-  }
+  public static class Builder {
+    private LoadEventInfo loadEventInfo;
+    private final List<LoadEventInfo> retriedLoadRequests;
+    private final List<SchemeData> schemeDatas;
+
+    public Builder(List<SchemeData> schemeDatas) {
+      this.schemeDatas = schemeDatas;
+      retriedLoadRequests = new ArrayList<>();
+      loadEventInfo = null;
```
stevemayhew: Also, in the renamed version handled the case where `schemeDatas` can be null (the caller has a Nullable member here).

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyRequestInfo.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.List;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+
+/**
+ * Encapsulates info for the sequence of load requests ({@link LoadEventInfo}, which were required
+ * to complete loading a DRM key
+ */
+public class KeyRequestInfo {
```
stevemayhew, line 16: Refactor renamed, in prep for other license requests beside Key load.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/KeyRequestInfo.java
```diff
@@ -0,0 +1,66 @@
...
+
+  public static class Builder {
+    @MonotonicNonNull private LoadEventInfo loadEventInfo;
+    private final List<LoadEventInfo> retriedLoadRequests;
+    @Nullable private final List<SchemeData> schemeDatas;
+
```
stevemayhew, line 22: Clean up the lint, `schemeDatas` is nullable for offline keys (we may want to add an additional identifying member for this case).

`loadEventInfo` is non-null after the main request is indemnified (which must be before `build()`

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/HttpMediaDrmCallback.java
icbaker: > What if we added the value to `ProvisionRequest` and `KeyRequest`

This is what I originally intended, but I just had a play around with this and it would require adding `taskId` to `ExoMediaDrm.getKeyRequest`, so it also has interface evolution problems (and this would also be the first piece of `KeyRequest` that isn't derived directly (or slightly indirectly) from the framework `MediaDrm.getKeyRequest`.

----

I agree that your current approach is probably the least bad option here.

However unfortunately while thinking about this I spotted a different/bigger problem which I'll raise in a new comment.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
```diff
@@ -666,6 +672,12 @@ public void handleMessage(Message msg) {
             break;
           case MSG_KEYS:
             response = callback.executeKeyRequest(uuid, (KeyRequest) requestTask.request);
+            if (currentKeyRequestInfo != null) {
+              LoadEventInfo loadEventInfo = callback.getLastLoadEventInfo();
```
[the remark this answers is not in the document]
stevemayhew: > Unfortunately because of the threading model used here, I don't think we can **always** guarantee that `getLastLoadEventInfo()` here corresponds to the request made above on L674.
> 
> This can happen because each `DefaultDrmSession` instance has its own request thread, but two instances can easily share the same `MediaDrmCallback` instance.

Ah, great catch...  I feel pretty stupid, given that  I just solved this same problem with my custom `LoadErrorHandlingPolicy` I wound up keeping the state object key'd by the load `taskId`

When I get back from vacation next week I'll move my original commits to a branch and download your changes (the force push would make a mess if I just pulled it) so I can see what is going on and have a closer look at your suggestions.   Thanks!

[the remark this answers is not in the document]
icbaker: > the force push would make a mess if I just pulled it

Ah sorry, the force push was part of me planning to send this PR for review internally before I realised this larger issue. I didn't push any 'meaningful' changes though, just formatted the changes in a commit, then rebased on top of the latest `main` commit. You should be able to force the branch pointer back to the 'old' version of https://github.com/androidx/media/commit/ebec29367295da0ba7ba8d20c27dc1e1a1383950 which I think is https://github.com/androidx/media/commit/9d27d45a65, then do your own force push to 'take back control' of the PR.

Probably something like this (though I'm no git expert, so there may be a better way to achieve this!):

```shell
$ git branch --force p-add-loadinfo-to-ondrmkeysloaded 9d27d45a65
$ git push --force
```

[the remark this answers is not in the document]
stevemayhew: > Unfortunately because of the threading model used here, I don't think we can **always** guarantee that `getLastLoadEventInfo()` here corresponds to the request made above on L674.
> 
> This can happen because each `DefaultDrmSession` instance has its own request thread, but two instances can easily share the same `MediaDrmCallback` instance.

Yes, stashing the `LoadEventInfo` anywhere other than the `RequestTask`, which is new'd for each request will violate the statelessness of whatever it is stuffed into.   There is no clean way out of just doing this.

> To resolve this I think its best to tightly couple each request to the reported `LoadEventInfo`, so we can keep `MediaDrmCallback` impls 'stateless' so it doesn't matter if they're invoked on multiple threads at the same time.
> 
> Which I think leads us to either:
> 
> 1. Pass `TransferListener` into `MediaDrmCallback.executeXXXRequest` (my previous suggestion in [Adds load info to `DrmSessionEventListeneronDrmKeysLoaded()` #1134 (comment)](https://github.com/androidx/media/pull/1134#discussion_r1503874316))
> 2. Enrich the return type of `MediaDrmCallback.executeKeyRequest` to `KeyResponse`, which has `byte[] data` field and optional `@Nullable LoadEventInfo` too.
>    

also, we could

3. Add to or lift a common base class from `ProvisionRequest` and `KeyRequest` that includes the `LoadEventInfo` as a member.

Number 3 avoids a signature change for `executeXXXRequest`,  but it seems uglier than your number 2, which is pretty strait forward.

So I would vote for your number 2.    I can do it or feel free to branch the current code and make a pull request onto my repo with your idea, I can build and test it.   Branches and pull request are best way to collaborate (once we agree the basic approach),  but no worries about the force push and thanks for re-basing to keep up with head of `main` (this is a big set of changes to keep on the side for long)

Thanks again for catching this.

stevemayhew:
This latest force-push was to rebase to latest main branch, dropped changes to `HttpMediaDrmCallback` to "resolve" the conflicts.  Will add code back that was moved from  `HttpMediaDrmCallback` to `DrmUtil.executePost()` in the next commit / push.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java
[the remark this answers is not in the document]
stevemayhew: @icbaker  finished the rebase to handle the move of modified code `DrmUtil.executePost()`.   In the process I addressed the thread safety of the `LoadEventInfo` object.  This is now entirely encapsulated in the `MediaDrmCallback.executeXXXRequest` methods, they now return an object with the response data and the `LoadEventInfo`

All of the test cases have been modified to create some manner of `LoadEventInfo` as well.

If this looks good we can stop here (this code is difficult to rebase) and leave the tech-debt to harden the test cases a bit to actually check for the `LoadEventInfo` (perhaps at least the `DefaultAnalyticsCollectorTest`, that would be a pretty quick add)

Let me know.

dgillenxperi:
Can this code be merged? A review was requested almost a year ago and we would benefit from having this change.

stevemayhew:
Hey @dgillenxperi need to rebase it to latest mainline.  

I've neglected it as it seemed that the Google teams interest in merging it had abated.  @andrewlewis and @icbaker is there any interest in taking this change?  If so I can rebase and prepare any other review requested changes

icbaker:
I think the general shape of this looks good - can you rebase it onto the tip of `main`?

I suggest we aim to merge it without adding the release -> `onKeysRemoved` logic in this change (we can do that in a follow-up). Same for `sessionId` I think.

stevemayhew:
Here you go, rebased to `main` and all lib-exoplayer unit tests pass.

icbaker:
I've pushed a couple of clean-up commits.

I'm going to rebase this onto the tip of `main`, add a release note and send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks!

icbaker:
There are a few implementations of `MediaDrmCallback` internally in google's monorepo, which mean they all have to be migrated from returning `byte[]` to `KeyResponse` in the same commit that the signature changes. That's quite awkward to do at the same time as merging a PR from GitHub, so I'm going to introduce the `KeyResponse` type in a separate change (just wrapping `byte[]`), get it reviewed and submitted, then rebase this on top (where we will add the `LoadEventInfo` field - but updated to be nullable so it doesn't have to be set).

icbaker:
There is still some un-guarded multi-threaded access to `DefaultDrmSession.currentKeyRequestInfo`. It is assigned on the calling thread of `acquire()` (playback thread), but accessed inside `DefaultDrmSession.RequestHandler.handleMessage` which is running on the [DRM request thread](https://github.com/androidx/media/blob/17956831696361097f9459632b57b5c8d0e04781/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSession.java#L317-L319).

I will add some locking to guard this access.

PR #1138, opened 2024-02-28
Fixed removal of user info for URLs that contain encoded @ characters

Lavamancer:
This bug occurs when using email addresses encoded as the user (i.e: `rtsp://user%40name:pass@foo.bar/foo.mkv`)

The solution is simply to use the `getEncodedAuthority` method instead of `getAuthority`. This resolves the issue because getAuthority decodes the encoded "@" symbols before splitting, resulting in a malformed URL.

Malformed URL example: `rtsp://name:pass/foo.mkv`

After the fix, it properly returns: `rtsp://foo.bar/foo.mkv`

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -44,6 +44,8 @@
 *   DASH Extension:
 *   Smooth Streaming Extension:
 *   RTSP Extension:
+    *   Fix user info removal for URLs that contain encoded @ characters
+        ([#1138](https://github.com/androidx/media/pull/1138)).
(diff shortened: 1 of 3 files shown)


Lavamancer:
The `removeUserInfo` method is used with the confidence that it will only remove that part since some URLs not only contain host and port but also path, query parameters, etc. The current issue of certain `%40` being converted to `@` has been resolved, allowing emails to be used as usernames in the user information.

microkatz:
@Lavamancer 

Issue #1150, opened 2024-03-05
HLS Iframe only track can't be played since 1.2.0.alpha01


zgzong:
### Version

Media3 1.2.0

### More version details

The HLS IFrame only track can be played on r1.1.1, same track can't be played since media3 r1.2.0-alpha01. 

### Devices that reproduce the issue

On android mobile, tablet and STB. 

Tested on mobile: pixel 7, pixel tablet. amazon fireTV 4k stb. 

### Reproducible in the demo app?

Yes

### Reproduction steps

Add Iframe only track stream into demo media.exolist.json, and play the demo app to play the iframe only track. 
Any version before (include) r1.1.1, can play IFrame only track. 
But since r1.2.0-alpha01, the stream can't be played, and no exception to throw. 

### Expected result

The IFrame only track can be played.

### Actual result

Segment downloaded, but nothing rendered.

### Media

[iframe_only_xmap.zip](https://github.com/androidx/media/files/14497931/iframe_only_xmap.zip)
[iframe_only.zip](https://github.com/androidx/media/files/14497959/iframe_only.zip)
Here are two Iframe only tracks.

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

icbaker:
CC @dsparano (author of https://github.com/androidx/media/pull/419)

The specific comment thread where we discussed HLS in the context of that PR is here: https://github.com/androidx/media/pull/419#discussion_r1217890240

dsparano:
Let me take a look at this.

dsparano:
Sorry for the delay. I think I've got a solution for H264 which would include general cases of byterange and I-Frame only HLS streams, which I had not personally tested when on https://github.com/androidx/media/pull/419.
However, I think I should test with the other readers (H262, H263 and H265) as well and for them I should have some streams available soon (if anyone has some valid streams pls share), after which I should be able to open a PR.

If an interim solution is required asap, then I suggest to disable "Send a synthesised empty pusi" if mode==MODE_HLS, @icbaker please let me know.

dsparano:
@icbaker possible fix here https://github.com/v-novaltd/androidx-media/tree/dsparano-exo245 shall I open a PR?

tonihei:
@dsparano Given the update seems to fix all reported issues, please go ahead and open the PR. We can continue any further detailed discussion in there.

dsparano:
[...]

Issue #1126, opened 2024-02-22
ERROR_CODE_IO_UNSPECIFIED error on HLS stream


needz:
### Version

Media3 1.2.1

### More version details

Started happening since version 1.2.0-alpha01
Issue is not happening in version 1.1.1 of ExoPlayer Demo App

### Devices that reproduce the issue

ADT-3
Sony Android TV
nVidia Shield Android TV

### Reproducible in the demo app?

Yes
Normal operation can be tested in version 1.1.1

### Reproduction steps

Play given HLS stream in ExoPlayer Demo App

### Expected result

Media plays without playback error

### Actual result

Media stops after 10 seconds (end of chunk) with ERROR_CODE_IO_UNSPECIFIED

Found out this portion of TsExtractor.java messes things up (line 328 - 334):

`      // Send a synthesised empty pusi to allow for packetFinished to be triggered on the last unit.
      for (int i = 0; i < tsPayloadReaders.size(); i++) {
        TsPayloadReader payloadReader = tsPayloadReaders.valueAt(i);
        if (payloadReader instanceof PesReader) {
          payloadReader.consume(new ParsableByteArray(), FLAG_PAYLOAD_UNIT_START_INDICATOR);
        }
      }`

Event Logs:

`EventLogger              E  playerFailed [eventTime=9.61, mediaPos=8.74, window=0, period=0, errorCode=ERROR_CODE_IO_UNSPECIFIED
                                                       androidx.media3.exoplayer.ExoPlaybackException: Source error
                                                           at androidx.media3.exoplayer.ExoPlayerImplInternal.handleIoException(ExoPlayerImplInternal.java:704)
                                                           at androidx.media3.exoplayer.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:677)
                                                           at android.os.Handler.dispatchMessage(Handler.java:102)
                                                           at android.os.Looper.loop(Looper.java:223)
                                                           at android.os.HandlerThread.run(HandlerThread.java:67)
                                                       Caused by: androidx.media3.exoplayer.upstream.Loader$UnexpectedLoaderException: Unexpected IllegalArgumentException: null
                                                           at androidx.media3.exoplayer.upstream.Loader$LoadTask.run(Loader.java:443)
                                                           at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
[...]

PR #1189, opened 2024-03-15
Fix missing last frame of MPEG-TS content

dsparano:
Also includes fixes for https://github.com/androidx/media/issues/1126 and https://github.com/androidx/media/issues/1150

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/ts/H265Reader.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/ts/H265Reader.java
@@ -394,7 +394,7 @@ public final class H265Reader implements ElementaryStreamReader {
 
     public void end(long position) {
       // Output a sample with the NAL units since the current nalUnitPosition
-      outputSample(/* offset= */ (int)(position - nalUnitPosition));
+      outputSample(/* offset= */ (int) (position - nalUnitPosition));
       // Output a final sample with the remaining NAL units up to the passed position
       samplePosition = nalUnitPosition;
       nalUnitPosition = position;
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/ts/H265Reader.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/ts/H265Reader.java
@@ -175,7 +175,6 @@ public final class H265Reader implements ElementaryStreamReader {
   public void packetFinished(boolean isEndOfInput) {
     assertTracksCreated();
     if (isEndOfInput) {
-      sampleReader.getSampleIsKeyframe();
       sampleReader.end(totalBytesWritten);
     }
   }
@@ -378,9 +377,15 @@ public final class H265Reader implements ElementaryStreamReader {
       }
     }
 
-    public boolean getSampleIsKeyframe() {
+    public void end(long position) {
       sampleIsKeyframe = nalUnitHasKeyframeData;
-      return sampleIsKeyframe;
+      // Output a sample with the NAL units since the current nalUnitPosition
+      outputSample(/* offset= */ (int) (position - nalUnitPosition));
+      // Output a final sample with the remaining NAL units up to the passed position
+      samplePosition = nalUnitPosition;
+      nalUnitPosition = position;
+      outputSample(/* offset= */ 0);
+      readingSample = false;
     }
 
     private void outputSample(int offset) {
@@ -390,16 +395,6 @@ public final class H265Reader implements ElementaryStreamReader {
       @C.BufferFlags int flags = sampleIsKeyframe ? C.BUFFER_FLAG_KEY_FRAME : 0;
       int size = (int) (nalUnitPosition - samplePosition);
       output.sampleMetadata(sampleTimeUs, flags, size, offset, null);
-    }
-
-    public void end(long position) {
-      // Output a sample with the NAL units since the current nalUnitPosition
-      outputSample(/* offset= */ (int) (position - nalUnitPosition));
-      // Output a final sample with the remaining NAL units up to the passed position
-      samplePosition = nalUnitPosition;
-      nalUnitPosition = position;
-      outputSample(/* offset= */ 0);
-      readingSample = false;
     }
 
     /** Returns whether a NAL unit type is one that occurs before any VCL NAL units in a sample. */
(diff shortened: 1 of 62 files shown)


rohitjoins:
Hi @dsparano,

Considering we had the [discussion](https://github.com/androidx/media/pull/864#issuecomment-1850114202) already, could you please grant me write access so I can merge the PR?"

rohitjoins:
Hi @dsparano,

There are lots of tests failing in [extractor.ts](https://github.com/androidx/media/tree/release/libraries/extractor/src/test/java/androidx/media3/extractor/ts) package with this change. 

You can fix this by temporarily setting [DumpFileAsserts.DUMP_FILE_ACTION](https://github.com/androidx/media/blob/release/libraries/test_utils/src/main/java/androidx/media3/test/utils/DumpFileAsserts.java#L75) to WRITE_TO_LOCAL.

Can you please run the tests to see if the changes in the dump files are correct?

rohitjoins:
Hi @dsparano,

> could you please grant me write access so I can merge the PR?

Thank you for sending an invite to collaborate on the `v-novaltd/androidx-media` repository. When I accepted the invitation, it said I have `View` access and I'm unable to push changes to the repository now as well.

Is there something you can do from your end to fix this?

dsparano:
@rohitjoins sorry, you have now write access.

Re tests I'll look into them asap.

rohitjoins:
@dsparano, I have added the updated dump files to the PR.

dsparano:
Thanks @rohitjoins

libraries/extractor/src/main/java/androidx/media3/extractor/ts/PesReader.java
```diff
@@ -165,6 +165,16 @@ private void setState(int state) {
     bytesRead = 0;
   }
 
+  public boolean canConsumeDummyEndOfInput() {
+    // Pusi only payload to trigger end of sample data is only applicable if
+    // pes does not have a length field and body is being read, another exclusion
```
rohitjoins, line 170: In the comment you say: `pes does not have a length field and body is being read`, but the condition check `payloadSize != C.LENGTH_UNSET` which means length field is set. 

The code is doing opposite to what the comment says, please clarify.

dsparano: Sorry you're right, the comment is right and the code in my last commit (when I added the function canConsumeDummyEndOfInput() is wrong, as from the previous commit the correct check is `!pesReader.hasPacketLength()` which means `payloadSize == C.LENGTH_UNSET`. Thanks for spotting it.

rohitjoins: I have updated it. Thanks for confirming.

rohitjoins:
There are test failures with this change in `TransformerWithInAppMuxerEndToEndTest`.

rohitjoins:
@dsparano We're temporarily rolling back the PR (https://github.com/androidx/media/pull/419) due to a bug fix release scheduled for next week. We'll reintroduce the changes once we've addressed the outstanding issues with this PR.

dsparano:
@rohitjoins I've cherry picked the #419 merge, re-applied the previous commits for fixing 1126 and 1150, and added a couple more to fix the H265 HLS cases which were not entirely correct and the `TransformerWithInAppMuxerEndToEndTest`. I think we should be good to go if you could please take a look.

rohitjoins:
Thank you for including the changes from #419 and fixing the tests. I'll take a look. Thanks :)

libraries/extractor/src/main/java/androidx/media3/extractor/ts/H265Reader.java
```diff
@@ -374,6 +377,17 @@ public void endNalUnit(long position, int offset, boolean hasOutputFormat) {
...
     }
 
+    public void end(long position) {
+      sampleIsKeyframe = nalUnitHasKeyframeData;
+      // Output a sample with the NAL units since the current nalUnitPosition
+      outputSample(/* offset= */ (int) (position - nalUnitPosition));
+      // Output a final sample with the remaining NAL units up to the passed position
+      samplePosition = nalUnitPosition;
+      nalUnitPosition = position;
+      outputSample(/* offset= */ 0);
```
rohitjoins, line 387: @dsparano, why are we outputting two different samples with the same timestamp for H.265 and not other readers? Please see the H.265 specific dump files in this PR. 

This may cause some problem with the renderers as it takes into consideration the last output time and only reads more samples when its timestamp is greater than the previous one.

dsparano: I did it like that because with HEVC at the boundary of HLS segments the sample taken up to the `position` was giving a merged Access Unit (AU), if compared with the main branch, of the last AU of the previous segment and the first one of the next segment. With the code you see now it returns the same AUs compared with main code across boundaries but also the very last AU of the last segment of a VOD HLS, which main fails to provide.

The challenge here is that the end of input at segment boundaries is not so much different than the end of input at end of file or VOD HLS, so I couldn't find a better way to handle them both without re-designing the entire logic more heavily.

If you ask why this is not needed for H264 and the other readers I think it is because of the different rules in determining the boundaries of AUs between the various codecs.

Any design tweak you can suggest I'm happy to give it a try.

rohitjoins: will you be able to provide an example stream which causes problems without this splitting?

dsparano:
@rohitjoins sure this is it: https://venus.v-nova.com/daniele/hls/byterange/h265/bbb_720p25_h265_keyint15_60sec.m3u8
All the HLS streams I created for this ticket are under here: https://venus.v-nova.com/daniele/hls/

libraries/extractor/src/main/java/androidx/media3/extractor/ts/PesReader.java
```diff
@@ -107,7 +107,8 @@ public final void consume(ParsableByteArray data, @Flags int flags) throws Parse
             Log.w(TAG, "Unexpected start indicator: expected " + payloadSize + " more bytes");
           }
           // Either way, notify the reader that it has now finished.
-          reader.packetFinished();
+          boolean isEndOfInput = (data.limit() == 0);
```
rohitjoins, line 110: @dsparano, I believe we will trigger `packetFinished` with `isEndOfInput` `true`here when we run out of data if network gets too slow, or if live stream generation runs behind. This may result in unwanted playback experience?

rohitjoins: oh yes, please ignore the comment.

PR #1214, opened 2024-03-22
Remove unused `package` property from `AndroidManifest.xml` files

MGaetan89:
Since Android Gradle Plugin 8.0, it is [recommended](https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes#namespace-dsl) to set the namespace in the module's `build.gradle` file, rather than in its `AndroidManifest.xml` file.
The `namespace`s were already set in the `build.gradle` files, but the `package` equivalent were still present in the `AndroidManifest.xml` files.

### DIFF
--- a/demos/cast/src/main/AndroidManifest.xml
+++ b/demos/cast/src/main/AndroidManifest.xml
@@ -13,8 +13,7 @@
      See the License for the specific language governing permissions and
      limitations under the License.
 -->
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="androidx.media3.demo.cast">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
   <uses-permission android:name="android.permission.INTERNET"/>
   <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
--- a/demos/gl/src/main/AndroidManifest.xml
+++ b/demos/gl/src/main/AndroidManifest.xml
@@ -13,8 +13,7 @@
      See the License for the specific language governing permissions and
      limitations under the License.
 -->
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="androidx.media3.demo.gl">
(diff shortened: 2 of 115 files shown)


oceanjules:
I have tried removing the package name a while ago during our internal clean-up week, but ran into the following error:

`Manifest merger failed: manifest:package attribute is not declared`

There might be some setup that prevents us from doing this for now. @icbaker?

icbaker:
We use Blaze internally (which is what Bazel is based on, though it's different) - and yeah these `package` properties seem to still be needed there.

I think it would be possible to craft some [copybara](https://github.com/google/copybara) in-line comments that would keep the `package` properties in the internal version of the code while removing them from this version, but unless there's an issue with them being present I'm afraid it doesn't seem worth the time to make this change in ~100 files.

----

As an aside, unfortunately merging PRs is a somewhat heavyweight process for us, because we have to run them backwards through Copybara and manually re-insert anything internal-only that was stripped from the touched files. For small changes/clean-ups like https://github.com/androidx/media/pull/1213, https://github.com/androidx/media/pull/1215, and https://github.com/androidx/media/pull/1216 it's actually much quicker if you just file an issue to let us know the problem and we can make the change directly on the internal version of the code. I understand this means you're not as clearly credited in the commit history, so if you would like us to merge these PRs then we can do so, but if you don't mind too much then I can just close them and make equivalent changes directly internally (I've already sent the Guava one for review, but for the others).

MGaetan89:
Sure, if it's easier to do all the changes internally, go ahead, that's fine for me 🙂

icbaker:
I'm going to close this, because it's too fiddly to merge this while keeping our internal build system working, and there doesn't seem to be a concrete benefit to doing so.

oceanjules:
@MGaetan89, we managed to make Gradle happy with our internal tools, so https://github.com/androidx/media/commit/70d7639ecb0c742448a46bcbdb194de064689824 should bring down the log spam significantly

PR #1215, opened 2024-03-22
Remove outdated TODO about some module's `group`

MGaetan89:
The `group` is already set to `androidx.media3` in the root `build.gradle` file:
https://github.com/androidx/media/blob/d13a0f4ec62ca092b79746a5725b62a3244cc5b4/build.gradle#L42

These two TODOs are probably leftovers from when it was `com.google.android.exoplayer`. No other module override this property.


icbaker:
Merged the equivalent change in https://github.com/androidx/media/commit/fd268eed46baefa59901035470246d07e6048062

PR #1225, opened 2024-03-26
Add support for parsing RTP packets with header extension used in RTSP

Kekelic:
RTP packets that contain header extension should be properly handled. Something about that is available in RFC3550, Section 5.3.1.

This fix only reads the header extension from the RTP packet if exists. It is done before parsing the remaining payload data, so that payload data can be handled properly.

Without this fix, Player that runs RTSP video will fail under the error: RTP H264 packetization mode [0] not supported. This is thrown in RtpH264Reader which consumes payload and chooses rtpH264PacketMode depending on the first byte of payload. So if payload contains the header extension, reader will fail.


microkatz:
Apologies for the delay. Thank you for creating a pull request to the project!

May I ask if you create a unit test in RtpPacketTest for your change? Also I think your code reads part of the extension header but does not actually read the extension. Your code I think would have issue with an header extension with a non-zero sized payload.

Kekelic:
> @Kekelic
> 
> Apologies for the delay. Thank you for creating a pull request to the project!
> 
> May I ask if you create a unit test in RtpPacketTest for your change? Also I think your code reads part of the extension header but does not actually read the extension. Your code I think would have issue with an header extension with a non-zero sized payload.

@microkatz 
I'm glad you answered.

Probably I could create it if needed.
Yes, that's what i was trying to emphasize. Code only read that header from RTP packet and don't do anything with it anymore. But if it doesn't read,  video will fail.
Not sure why u think that. I have a system with video stream which produces RTP packets with extension and only this fix make it to work. Also i tested this fix with a video stream that produces RTP packets without extension and it worked same as before this fix.

microkatz:
@Kekelic 

Definitely reach out if you are struggling!

You can look at the earlier code in that `parse(ParsableByteArray packetBuffer)` and the diagram in [RFC 3550 Section 5.1 - RTP Fixed Header Fields](https://datatracker.ietf.org/doc/html/rfc3550#section-5.1) for examples on how to parse the data field.

The extension looks like the following:
```
 0                   1                   2                   3
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |      defined by profile       |           length              |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |                        header extension                       |
   |                             ....                              |
```

Basically you either want to read the first Int(4 bytes) and mask against 0xFF to get the length. Or just read the first short(two bytes) then read another short(2 bytes) as the length. The second way would be as follows(also this is rough, with comments for you, extraneously descriptive parameter names, etc.):

```
//Extension.
 if (hasExtension){
   int headerExtensionProfileData = packetBuffer.readShort();
   int headerExtensionPayloadLength = packetBuffer.readShort(); // Number of 32 bit words in extension payload
   packetBuffer.readInt(); // extension header
   if (headerExtensionPayloadLength != 0) {
       byte[] extensionPayload = new byte[headerExtensionPayloadLength * 4];
       packetBuffer.readBytes(extensionPayload, 0, headerExtensionPayloadLength*4);
   }
 }
```

What would be very helpful for your unit test is if you took the example RTP from the H264 stream that fails at the moment and use that. Then you could debug in the parsing code and make sure the length and payload data matches as you expect.

Kekelic:
@microkatz 

Great explained and I like your descriptive parameter names.

Both solutions worked, but I made it work with a first solution with small changes. Take a look.

Also, I made 2 unit tests with RTP data that failed without this fix as you told me, but it didn't require me to make RtpH264ReaderTest file. So I'm not sure if that's only what you are looking for. But I covered checking for payload data matches as expected and to build packet matches packet data which required me to edit a RtpPacket little bit more.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtpPacket.java
```diff
@@ -233,6 +271,24 @@ public static RtpPacket parse(ParsableByteArray packetBuffer) {
       csrc = EMPTY;
     }
 
+    //Extension.
+    byte[] headerExtension;
+    byte[] extensionPayload;
+    if (hasExtension) {
+      headerExtension = new byte[HEADER_EXTENSION_SIZE];
+      packetBuffer.readBytes(headerExtension, 0, HEADER_EXTENSION_SIZE);
+      int extensionPayloadLength = (headerExtension[2] & 0xFF) << 8 | (headerExtension[3] & 0xFF);
```
microkatz, line 280: Is the & with 0xFF necessary? I think that would just return headerExtension[N]?

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtpPacket.java
```diff
@@ -143,6 +173,7 @@ public RtpPacket build() {
   public static final int MIN_SEQUENCE_NUMBER = 0;
   public static final int MAX_SEQUENCE_NUMBER = 0xFFFF;
   public static final int CSRC_SIZE = 4;
+  public static final int HEADER_EXTENSION_SIZE = 4;
```
microkatz, line 176: I think the name of this constant field is misleading? Its not the size of the header extension but the size of the first bit of it that contains the "profile related value" and the extension length. At least that is what you are using it for.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtpPacket.java
```diff
@@ -233,6 +271,24 @@ public static RtpPacket parse(ParsableByteArray packetBuffer) {
...
+    byte[] headerExtension;
+    byte[] extensionPayload;
+    if (hasExtension) {
+      headerExtension = new byte[HEADER_EXTENSION_SIZE];
+      packetBuffer.readBytes(headerExtension, 0, HEADER_EXTENSION_SIZE);
+      int extensionPayloadLength = (headerExtension[2] & 0xFF) << 8 | (headerExtension[3] & 0xFF);
+      if (extensionPayloadLength != 0) {
+        extensionPayload = new byte[extensionPayloadLength * 4];
+        packetBuffer.readBytes(extensionPayload, 0, extensionPayloadLength * 4);
+      }else {
```
microkatz, line 284: nit: Add space between  '}' and 'else'

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtpPacket.java
```diff
@@ -246,6 +302,9 @@ public static RtpPacket parse(ParsableByteArray packetBuffer) {
         .setTimestamp(timestamp)
         .setSsrc(ssrc)
         .setCsrc(csrc)
+        .setExtension(hasExtension)
```
microkatz, line 305: I don't think its necessary to save the header extension data. None of the RTSP readers are utilizing this data at the moment and its just adding cruft to the class. I think its enough to just parse the extension if it exists as to not have the readers choke on the payload.

Kekelic: @microkatz 

True, good suggestions.
So I can also remove saving "`headerExtension`" (thats's also misleading) and `extensionPayload` data? Which means I can also delete second Unit test I added `buildRtpPacketWithHeaderExtension_matchesPacketData`. And If I don't save hasExtension I will not check it in first Unit test I added `parseRtpPacketWithHeaderExtension` ?

microkatz: The reason why I mentioned the RtpH264ReaderTest was to have a unit test parsing an RTP packet with both a header extension and a payload. In this case, without your changes, the Rtp parsing logic would incorrectly group the headerextension into the general payload. If you can create a unit test in RtpPacketTest that exemplifies this then that is fine as well.

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtpPacket.java
microkatz: Ah. I see you may have used the 0xFF mapping due to java sign handling. You could use some guava method? Like [https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/primitives/Ints.html#fromBytes(byte,byte,byte,byte)](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/primitives/Ints.html#fromBytes%28byte,byte,byte,byte%29).
`Ints.fromBytes(0, 0, headerExtension[2], headerExtension[3])`

Kekelic: Actually that I was planning to do firstly, but didn't know could I include `Ints` in file.

Kekelic:
@microkatz Hi, sorry for the big delay.

I made a changes we discussed. Now, data about header extension is not saved in `RtpPacket` and `RtpPacket `now only reads headers extension. `RtpPacketTest` has one new unit test `parseRtpPacketWithHeaderExtension` that tests use case when RTP packet has header extension. 

What I think now can be confusing that `RtpPacket ` has hardcoded extension value to false (as I removed saving data, I set it to be as before) and all unit tests in `RtpPacketTest ` check extension to be false including this new unit test which I added. I just want to check that this is fine?

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/RtpPacketTest.java
```diff
@@ -68,6 +68,29 @@ public final class RtpPacketTest {
       Arrays.copyOfRange(
           rtpDataWithLargeTimestamp, RtpPacket.MIN_HEADER_SIZE, rtpDataWithLargeTimestamp.length);
 
+  /*
+   10.. .... = Version: RFC 1889 Version (2)
+   ..0. .... = Padding: False
+   ...0 .... = Extension: True
```
microkatz, line 74: If the extension is true then this bit should be a 1.

microkatz:
@Kekelic

>What I think now can be confusing that RtpPacket  has hardcoded extension value to false (as I removed saving data, I set it to be as before) and all unit tests in RtpPacketTest  check extension to be false including this new unit test which I added. I just want to check that this is fine?

I understand what you are asking. I think that you could put a comment above the check in your unit test to note that the created RtpPacket object will parse but not save the extension data. You can also change the title of your unit test to note the specific behavior you are intending to test. For example, you can change the name to "parseRtpPacketWithHeaderExtension_createsRtpPacketWithoutHeaderExtension".

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/RtpPacketTest.java
Kekelic: True, it was an error in copying the code. Also, I changed 1 to 0 at Marker bit.

Kekelic:
@microkatz 
It seems fine to me now. If there is something else I need to add or change, please let me know.
Also do you know when this PR will be merged/included in release?

microkatz:

Issue #1224, opened 2024-03-26
Application using the library will crash if 1000 DownloadHelper instances are created with a non-null RenderersFactory


retiredreplicant:
### Version

Media3 1.3.0

### More version details

d13a0f4ec62ca092b79746a5725b62a3244cc5b4

### Devices that reproduce the issue

All devices

### Devices that do not reproduce the issue

No devices excluded

### Reproducible in the demo app?

No

### Reproduction steps

1. Create a player application that can create and release 1000 DownloadHelper instances with a non-null RenderersFactory (or modify the demo app to do so.)
2. Create and release 1000 DownloadHelper instances with a non-null RenderersFactory 

### Expected result

An application using the player should be able to create an unlimited number of DownloadHelper instances with a non-null RenderersFactory.

### Actual result

The application will crash with a "Too many receivers, total of 1000, registered for pid" exception.

**Source of Issue**
The source of the issue is that the DownloadHelper uses the passed RenderersFactory to create renderers, and to get capabilities for those renderers, but never releases the renderers that are created. One of the renderers created is a MediaCodecAudioRenderer. That renderer is created with a DefaultAudioSink. When the DownloadHelper processes a DOWNLOAD_HELPER_CALLBACK_MESSAGE_PREPARED message, the DownloadHelper calls runTrackSelection and this results in the DefaultAudioSink creating an instance of AudioCapabilitiesReceiver and registering that receiver. The release() method for the DefaultAudioSink unregisters that listener, but since the DownloadHelper does not call release() on the renderers it creates, release is never called on the DefaultAudioSink and the unregister never occurs.  If 1000 DownloadHelpers are created with a non-null RenderersFactory, the application will crash with a "Too many receivers, total of 1000, registered for pid" exception. 

**Reproduction with Demo App**
Not reproducible with the demo app only because there are not 1000 items to download.  However, the registration of a AudioCapabilitiesReceiver for each download without a corresponding unregister can be seen with the demo app.  Set breakpoints at lines 1520 `audioCapabilitiesReceived.unregister();` and 1749 `audioCapabilities = audioCapabilitiesReceiver.register();` in DefaultAudioSink and initiated a download with the demo app.  You will hit the breakpoint at 1749 (register) for each download, but the breakpoint at 1520 (unregister) is never hit.  This demonstrates receivers being registered but never unregistered.  Note that the player **_does_** properly release the renderers that it creates.  It is only the DownloadHelper that does not. 

**Recommend Change**
[...]

PR #1230, opened 2024-03-28
Fix too many receivers app crash when initiating multiple Download Helpers

sahilgarg90:
This PR will fix below linked issue,

https://github.com/androidx/media/issues/1224


icbaker:
I think this is no longer needed after https://github.com/androidx/media/commit/c5e894e2d63aae9d0014e01fff99b34cd5ba3ec5

PR #1235, opened 2024-04-02
Add Dolby Vision Transcoding and Editing Support

ybai001:
This code contribution adds Dolby Vision transcoding and editing support to transformer.


adarsh-dhakad (approved):

SheenaChhabra:
@ybai001 I just wanted to update that due to other high priority work, it might take a few weeks to review this change. I appreciate your patience.

ybai001:
> @ybai001 I just wanted to update that due to other high priority work, it might take a few weeks to review this change. I appreciate your patience.

Sure, understood.

libraries/transformer/src/main/java/androidx/media3/transformer/DefaultDecoderFactory.java
```diff
@@ -135,6 +135,9 @@ public DefaultCodec createForVideoDecoding(
     if (SDK_INT >= 31 && requestSdrToneMapping) {
       mediaFormat.setInteger(
           MediaFormat.KEY_COLOR_TRANSFER_REQUEST, MediaFormat.COLOR_TRANSFER_SDR_VIDEO);
+    } else if (SDK_INT >= 31 && MimeTypes.VIDEO_DOLBY_VISION.equals(format.sampleMimeType)) {
```
SheenaChhabra, line 138: We generally write the condition in reversed way `format.sampleMimeType.equals(MimeTypes.VIDEO_DOLBY_VISION)`

SheenaChhabra, line 138: Can you share some documentation which explains this code?

icbaker: There's lots of occurrences of the ['yoda condition'](https://en.wikipedia.org/wiki/Yoda_conditions) in our codebase, but I agree with @SheenaChhabra it is now discouraged in new code. In this case, because `Format.sampleMimeType` is `@Nullable` you will need to use `Objects.equals(format.sampleMimeType, MimeTypes.VIDEO_DOLBY_VISION)`.

ybai001: Dolby Vision is not only a HDR type but also a media type. But it is based on H.264, H.265 or AV1. That is to say, the Dolby Vision stream is base encoded data + Dolby Vision metadata. Dolby Vision decoder will do two things in playback use case: (1) Call base codec, e.g. H.265 decoder, to decode the base encoded data; (2) Apply Dolby Vision metadata to decoded base data (output of step 1) to get Dolby Vision experience.

In editing use case,  Dolby Vision decoder does step 1 only. We don't apply the Dolby Vision metadata because the result of this process is related to the screen characteristic and environment brightness. For editing use case, the Dolby Vision metadata will be generated by Dolby Vision encoder. For transcoding use case, this is a trade off between better quality and consistent result on different devices.

By default, Dolby Vision decoder will do both above step 1 & 2. The purpose of
`mediaFormat.setInteger(MediaFormat.KEY_COLOR_TRANSFER_REQUEST, MediaFormat.COLOR_TRANSFER_HLG);`
is to notify Dolby Vision decoder to do step 1 only.

There is no open specification to explain above logic. But I can share some documentations about Dolby Vision.
1. Dolby Vision official site: https://www.dolby.com/technologies/dolby-vision/
2. Dolby Vision specification and white paper: https://professionalsupport.dolby.com/s/specifications-and-white-papers?language=en_US

libraries/transformer/src/main/java/androidx/media3/transformer/FrameworkMuxer.java
```diff
@@ -108,6 +111,11 @@ public int addTrack(Format format) throws MuxerException {
     if (isVideo) {
       mediaFormat = MediaFormat.createVideoFormat(sampleMimeType, format.width, format.height);
       MediaFormatUtil.maybeSetColorInfo(mediaFormat, format.colorInfo);
+      if (sampleMimeType.equals(MimeTypes.VIDEO_DOLBY_VISION) &&
+          android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.TIRAMISU) {
+        mediaFormat.setInteger(MediaFormat.KEY_PROFILE, getDvProfile(format));
```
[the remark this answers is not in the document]
ybai001: As far as Dolby Vision and Profile, the spec is here: https://professionalsupport.dolby.com/s/article/What-is-Dolby-Vision-Profile?language=en_US

And I'd like to add some explanation why I did this code change:

In Media3 library code FrameworkerMuxer.java, addTrack(Format format) method, it calls system MediaMuxer.addTrack(mediaFormat) method without PROFILE/LEVEL settings.

This is OK for other video formats but results in issue for Dolby Vision. The key point is in Android AOSP utils.cpp file, convertMessageToMetadata() method, KEY_PROFILE is checked. If there is no this parameter, BAD_VALUE will be returned.

see: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/libstagefright/Utils.cpp;drc=216a97bdad3c0f492094250b33d82abf27098932;l=2115

ybai001:
> We need to have some tests cases around these changes.

What can I do for this request? At least, I can provide test stream. If you mean I need to implement some test cases, could you guide me what kinds of test cases I should add. Previously, I added some DD+JOC and AC-4 test cases for my media3 contributions so that I am familiar with that process. But for transformer, this is my first contribution.

libraries/transformer/src/main/java/androidx/media3/transformer/DefaultDecoderFactory.java
ybai001: Done.

SheenaChhabra:
> > We need to have some tests cases around these changes.
> 
> What can I do for this request? At least, I can provide test stream. If you mean I need to implement some test cases, could you guide me what kinds of test cases I should add. Previously, I added some DD+JOC and AC-4 test cases for my media3 contributions so that I am familiar with that process. But for transformer, this is my first contribution.

We could include following tests:
1. Transmuxing a dolby vision input file should preserve the dolby vision data. This could be a robolectric test. For this, you can refer tests in MediaItemExportTest.java
2. Transcoding a normal H.264/H.265 to dolby vision. For this your can refer tests in TransformerEndToEndTest.java
3. Editing a dolby vision input. For this your can refer tests in TransformerEndToEndTest.java

I would also suggest to run these scenario using Transformer demo app. This is to ensure that it works as expected.

ybai001:
> > > We need to have some tests cases around these changes.
> > 
> > 
> > What can I do for this request? At least, I can provide test stream. If you mean I need to implement some test cases, could you guide me what kinds of test cases I should add. Previously, I added some DD+JOC and AC-4 test cases for my media3 contributions so that I am familiar with that process. But for transformer, this is my first contribution.
> 
> We could include following tests:
> 
> 1. Transmuxing a dolby vision input file should preserve the dolby vision data. This could be a robolectric test. For this, you can refer tests in MediaItemExportTest.java
> 2. Transcoding a normal H.264/H.265 to dolby vision. For this your can refer tests in TransformerEndToEndTest.java
> 3. Editing a dolby vision input. For this your can refer tests in TransformerEndToEndTest.java
> 
> I would also suggest to run these scenario using Transformer demo app. This is to ensure that it works as expected.

Hi, @SheenaChhabra,
Thanks for your suggestion. As far as the three tests you mentioned,
1. I checked the file MediaItemExportTest.java and there are several test cases for transmux. But all of them are general test cases. None of them is for specific video format, e.g. AV1. Could you help to clarify why we need to add test case for Dolby Vision but not for other video format? And which test case I should refer to?
2. I have added the case as you mentioned. I created it based on this one (for AV1) https://github.com/androidx/media/blob/main/libraries/transformer/src/androidTest/java/androidx/media3/transformer/TransformerEndToEndTest.java#L1517
3. I checked the file TransformerEndToEndTest.java. But all of them is related to HDR editing. Instead, I find there is a Dolby Vision editing case in https://github.com/androidx/media/blob/main/libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java#L191. Is it enough for your requirement?

libraries/transformer/src/main/java/androidx/media3/transformer/DefaultDecoderFactory.java
SheenaChhabra: > For transcoding use case, this is a trade off between better quality and consistent result on different devices.

Which trade off you are talking about?

ybai001: > > For transcoding use case, this is a trade off between better quality and consistent result on different devices.
> 
> Which trade off you are talking about?

For transcoding use case, ideally, we should do above step 1 & 2 for decoding then encoding to another video format. This is "better quality". In practice (in this contribution), we do step 1 only for decoding. The reason is the result of step 2 is related to the screen characteristic and environment brightness. That means user will get different results on different devices. The transcoding result is not "consistent". So "ignoring step 2" is a trade off.

ybai001:
I have uploaded the test cases code.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/TransformerEndToEndTest.java
```diff
@@ -1280,6 +1280,39 @@ public void transcode_withOutputVideoMimeTypeAv1_completesSuccessfully() throws
     assertThat(exportResult.videoMimeType).isEqualTo(MimeTypes.VIDEO_AV1);
   }
 
+  @Test
+  public void transcode_withOutputVideoMimeTypeDolbyVision_completesSuccessfully() throws Exception {
```
SheenaChhabra, line 1284: On which device you have tested this which can produce dolby vision metadata?

libraries/transformer/src/androidTest/java/androidx/media3/transformer/TransformerEndToEndTest.java
```diff
@@ -1280,6 +1280,39 @@ public void transcode_withOutputVideoMimeTypeAv1_completesSuccessfully() throws
...
+    Transformer transformer =
+        new Transformer.Builder(context).setVideoMimeType(MimeTypes.VIDEO_DOLBY_VISION).build();
+
+    ExportTestResult exportTestResult =
+        new TransformerAndroidTestRunner.Builder(context, transformer)
+            .build()
+            .run(testId, editedMediaItem);
+    ExportResult exportResult = exportTestResult.exportResult;
+
+    assert exportTestResult.filePath != null;
```
SheenaChhabra, line 1307: I think its not needed.

libraries/transformer/src/main/java/androidx/media3/transformer/FrameworkMuxer.java
```diff
@@ -276,9 +284,64 @@ private static ImmutableList<String> getSupportedVideoSampleMimeTypes() {
...
+      if (pps <= 22118400) {
+        level = MediaCodecInfo.CodecProfileLevel.DolbyVisionLevelHd24;  // Level 01
+      } else { // pps <= 27648000
+        level = MediaCodecInfo.CodecProfileLevel.DolbyVisionLevelHd30;  // Level 02
+      }
+    } else if (maxWidthHeight <= 1920 && pps <= 49766400) {
+      level = MediaCodecInfo.CodecProfileLevel.DolbyVisionLevelFhd24;  // Level 03
+    } else if (maxWidthHeight <= 2560 && pps <= 62208000) {
+      level = MediaCodecInfo.CodecProfileLevel.DolbyVisionLevelFhd30;  // Level 04
+    } else if (maxWidthHeight <= 3840) {
```
SheenaChhabra, line 323: Can we use this existing method?

https://source.corp.google.com/piper///depot/google3/third_party/java_src/android_libs/media/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecUtil.java;l=307?q=MediaCodecUti#:~:text=306-,307,-308

libraries/transformer/src/main/java/androidx/media3/transformer/EncoderUtil.java
```diff
@@ -150,6 +150,13 @@ public static ImmutableList<Integer> getCodecProfilesForHdrFormat(
           return ImmutableList.of(MediaCodecInfo.CodecProfileLevel.AV1ProfileMain10HDR10);
         }
         break;
+      case MimeTypes.VIDEO_DOLBY_VISION:
+        if (colorTransfer == C.COLOR_TRANSFER_HLG) {
+          return ImmutableList.of(
+              MediaCodecInfo.CodecProfileLevel.DolbyVisionProfileDvheSt);
```
SheenaChhabra, line 156: Out of all profile, why have we choose this one?

libraries/transformer/src/main/java/androidx/media3/transformer/DefaultDecoderFactory.java
SheenaChhabra: I am still not sure of this change. if the decoder givens us dynamic metadata then I think the pipeline will simple ignore it.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -139,6 +139,41 @@ public void export_transmuxHlg10File() throws Exception {
...
+    if (Util.SDK_INT < 24) {
+      // TODO: b/285543404 - Remove suppression once we can transmux H.265/HEVC before API 24.
+      recordTestSkipped(context, testId, /* reason= */ "Can't transmux H.265/HEVC before API 24");
+      return;
+    }
+
+    if (AndroidTestUtil.skipAndLogIfFormatsUnsupported(
+        context,
+        testId,
+        /* inputFormat= */ MP4_ASSET_DOLBY_VISION_HDR_FORMAT,
```
SheenaChhabra, line 155: Why do we need this when we are not transcoding?

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -139,6 +139,41 @@ public void export_transmuxHlg10File() throws Exception {
     assertThat(actualColorTransfer).isEqualTo(C.COLOR_TRANSFER_HLG);
   }
 
+  @Test
+  public void export_transmuxDolbyVisionFile() throws Exception {
+    Context context = ApplicationProvider.getApplicationContext();
+
+    if (Util.SDK_INT < 24) {
+      // TODO: b/285543404 - Remove suppression once we can transmux H.265/HEVC before API 24.
+      recordTestSkipped(context, testId, /* reason= */ "Can't transmux H.265/HEVC before API 24");
```
SheenaChhabra, line 148: The below check for format support should check whether muxer supports this format or not so this is not needed.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -139,6 +139,41 @@ public void export_transmuxHlg10File() throws Exception {
...
+    }
+
+    Transformer transformer = new Transformer.Builder(context).build();
+    MediaItem mediaItem = MediaItem.fromUri(Uri.parse(MP4_ASSET_DOLBY_VISION_HDR));
+
+    ExportTestResult exportTestResult =
+        new TransformerAndroidTestRunner.Builder(context, transformer)
+            .build()
+            .run(testId, mediaItem);
+    assert exportTestResult.filePath != null;
```
SheenaChhabra, line 167: We generally use `assertThat(xyz).isNotNull();`

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -139,6 +139,41 @@ public void export_transmuxHlg10File() throws Exception {
...
+    @C.ColorTransfer
+    int actualColorTransfer =
+        Objects.requireNonNull(
```
SheenaChhabra, line 170: I think we can remove it.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -139,6 +139,41 @@ public void export_transmuxHlg10File() throws Exception {
...
+            retrieveTrackFormat(context, exportTestResult.filePath, C.TRACK_TYPE_VIDEO)
+                .colorInfo)
+            .colorTransfer;
+    assertThat(actualColorTransfer).isEqualTo(C.COLOR_TRANSFER_HLG);
```
SheenaChhabra, line 174: Can we check the mime type as well?

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -224,10 +259,12 @@ public void exportAndTranscode_dolbyVisionFile_whenHdrEditingIsSupported() throw
         new TransformerAndroidTestRunner.Builder(context, transformer)
             .build()
             .run(testId, editedMediaItem);
+    assert exportTestResult.filePath != null;
     @C.ColorTransfer
     int actualColorTransfer =
-        retrieveTrackFormat(context, exportTestResult.filePath, C.TRACK_TYPE_VIDEO)
-            .colorInfo
+        Objects.requireNonNull(
```
SheenaChhabra, line 265: Why do we need these changes?

libraries/transformer/src/androidTest/java/androidx/media3/transformer/TransformerEndToEndTest.java
ybai001: I added this assert sentence because Android Studio had a warning message.  I can remove it if you are fine.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: OK, I'll change it.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: OK

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: Just because there is a warning message from latest Android Studio. I can revert this change if you are fine.

SheenaChhabra: Just had a discussion with my team about these warnings and we have decided to ignore these. This is because we have other internal tools to check these nullness issue (although we have disabled them for test to avoid null check overhead).
TLDR; you can remove these and other null checks from the test code.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -360,6 +397,74 @@ public void onFallbackApplied(
...
 
+  @Test
+  public void exportAndTranscode_dolbyVisionFile_whenHdrEditingUnsupported_toneMapsOrThrows()
+      throws Exception {
+    Context context = ApplicationProvider.getApplicationContext();
+    Format format = MP4_ASSET_DOLBY_VISION_HDR_FORMAT;
+    assert format.colorInfo != null;
+    if (deviceSupportsHdrEditing(VIDEO_H265, format.colorInfo)) {
+      recordTestSkipped(context, testId, /* reason= */ "Device supports Dolby Vision editing.");
+      return;
```
SheenaChhabra, line 408: We can possibly use assumeDeviceDoesNotSupportHdrEditing() ?

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -360,6 +397,74 @@ public void onFallbackApplied(
...
+    }
+
+    if (AndroidTestUtil.skipAndLogIfFormatsUnsupported(
+        context, testId, /* inputFormat= */ format, /* outputFormat= */ null)) {
```
SheenaChhabra, line 412: This method has been renamed to assumeFormatsSupported(). Sorry for the conflicts.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -360,6 +397,74 @@ public void onFallbackApplied(
...
+    }
+
+    AtomicBoolean isFallbackListenerInvoked = new AtomicBoolean();
+    AtomicBoolean isToneMappingFallbackApplied = new AtomicBoolean();
+    Transformer transformer =
+        new Transformer.Builder(context)
+            .addListener(
+                new Transformer.Listener() {
+                  @Override
+                  public void onFallbackApplied(
```
SheenaChhabra, line 423: This method is deprecated, use the alternative.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -360,6 +397,74 @@ public void onFallbackApplied(
...
+      assertThat(isToneMappingFallbackApplied.get()).isTrue();
+      assert exportTestResult.filePath != null;
+      @C.ColorTransfer
+      int actualColorTransfer =
+          Objects.requireNonNull(
+              retrieveTrackFormat(context, exportTestResult.filePath, C.TRACK_TYPE_VIDEO)
+                  .colorInfo)
+              .colorTransfer;
+      assertThat(actualColorTransfer).isEqualTo(C.COLOR_TRANSFER_SDR);
+    } catch (ExportException exception) {
```
SheenaChhabra, line 453: I am not sure why this catch block exists in the existing tests but transformer propagates the `ExportException` via listener and also sets it on `ExportResult` object.
So we should fetch this exception from the `ExportResult` and do the validation.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: sure.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: I'll remove this checking as you latest comment.

libraries/transformer/src/main/java/androidx/media3/transformer/EncoderUtil.java
ybai001: Good question. All exist devices which support Dolby Vision encoding on market will encode the input stream to this profile only. In future, maybe we will extend this list. But Dolby has not finalized which profiles we should support for encoding.

libraries/transformer/src/main/java/androidx/media3/transformer/FrameworkMuxer.java
```diff
@@ -108,6 +111,11 @@ public int addTrack(Format format) throws MuxerException {
     if (isVideo) {
       mediaFormat = MediaFormat.createVideoFormat(sampleMimeType, format.width, format.height);
       MediaFormatUtil.maybeSetColorInfo(mediaFormat, format.colorInfo);
+      if (sampleMimeType.equals(MimeTypes.VIDEO_DOLBY_VISION) &&
+          android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.TIRAMISU) {
```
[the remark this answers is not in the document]
ybai001: sure.

libraries/transformer/src/main/java/androidx/media3/transformer/FrameworkMuxer.java
ybai001: It's for different purpose. The method in your link is to PARSE the Dolby Vision profile and level from codec string. The use case is to play a streaming content. You can get codec string from a MPEG-DASH manifest file. Then paring this codec string, you will get profile/level info. This info can be used to select which Dolby Vision decoder to handle this streaming content (H.264 based Dolby Vision decoder or H.265 based Dolby Vision decoder). In this case, Format.codecs is a valid codec string.

As far as this code contribution case, it is used to decide which Dolby Vision profile/level should be used based on incoming content width/height/framerate. In this case. Format.codecs is NULL.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/TransformerEndToEndTest.java
ybai001: I'll remove it based on your latest comment.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: Based on your latest comment, I'll remove this sentence.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -139,6 +139,41 @@ public void export_transmuxHlg10File() throws Exception {
     assertThat(actualColorTransfer).isEqualTo(C.COLOR_TRANSFER_HLG);
   }
 
+  @Test
+  public void export_transmuxDolbyVisionFile() throws Exception {
```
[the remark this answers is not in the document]
ybai001: Could you give me more hint about it?
1. I added the Dolby Vision test case here because there are export_transmuxHdr10File() and export_transmuxHlg10File() test cases in this file. I did the similar thing in my test case implementation.
2. In MediaItemExportTest, I found some transmux test cases but I didn't find corresponding test cases for HLG and HDR10. Do I need to add test case for Dolby Vision in this file?

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: I just did the same thing as export_transmuxHdr10File() and export_transmuxHlg10File(). Shall we remove above code in these three methods together?

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: Sorry, I'm not familiar with resolving confliction on GitHub. Let's complete other issues then handle the confliction issue. (I fork the main branch at the beginning of this code contribution. At that time, there is no this method.)

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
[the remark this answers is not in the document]
SheenaChhabra: I think these tests needs to run on real device because in `Robolectric` tests, `FrameworkMuxer` does not write to the actual output file and we are doing validations on the actual output file. So we should keep the test case here only. Thanks!

libraries/transformer/src/main/java/androidx/media3/transformer/FrameworkMuxer.java
```diff
@@ -276,9 +283,64 @@ private static ImmutableList<String> getSupportedVideoSampleMimeTypes() {
...
+    if (maxWidthHeight <= 1280) {
+      if (pps <= 22118400) {
+        level = MediaCodecInfo.CodecProfileLevel.DolbyVisionLevelHd24;  // Level 01
+      } else { // pps <= 27648000
+        level = MediaCodecInfo.CodecProfileLevel.DolbyVisionLevelHd30;  // Level 02
+      }
+    } else if (maxWidthHeight <= 1920 && pps <= 49766400) {
+      level = MediaCodecInfo.CodecProfileLevel.DolbyVisionLevelFhd24;  // Level 03
+    } else if (maxWidthHeight <= 2560 && pps <= 62208000) {
+      level = MediaCodecInfo.CodecProfileLevel.DolbyVisionLevelFhd30;  // Level 04
```
SheenaChhabra, line 321: Do you know why the value of these constants is not same as mentioned level? For example DolbyVisionLevelFhd30 = 8

libraries/transformer/src/main/java/androidx/media3/transformer/FrameworkMuxer.java
```diff
@@ -276,9 +283,64 @@ private static ImmutableList<String> getSupportedVideoSampleMimeTypes() {
...
+  // Refer to https://professionalsupport.dolby.com/s/article/What-is-Dolby-Vision-Profile
+  @RequiresApi(33)
+  private static int getDvProfile(Format format) {
+    // Currently, only profile 8 is supported for encoding
+    // TODO: set profile ID based on format.
+    return MediaCodecInfo.CodecProfileLevel.DolbyVisionProfileDvheSt;
+  }
+
+  // Get Dolby Vision level
+  // Refer to https://professionalsupport.dolby.com/s/article/What-is-Dolby-Vision-Profile
```
SheenaChhabra, line 305: Please write it as a java doc comment. Same comment for other method.

SheenaChhabra:
@ychaparov Can you please help with another review of this PR?

libraries/transformer/src/androidTest/java/androidx/media3/transformer/TransformerEndToEndTest.java
```diff
@@ -1280,6 +1280,38 @@ public void transcode_withOutputVideoMimeTypeAv1_completesSuccessfully() throws
...
+            .build()
+            .run(testId, editedMediaItem);
+    ExportResult exportResult = exportTestResult.exportResult;
+
+    String actualMimeType =
+        retrieveTrackFormat(context, exportTestResult.filePath, C.TRACK_TYPE_VIDEO).sampleMimeType;
+    assertThat(actualMimeType).isEqualTo(MimeTypes.VIDEO_DOLBY_VISION);
+    assertThat(exportResult.exportException).isNull();
+    assertThat(exportResult.durationMs).isGreaterThan(0);
+    assertThat(exportResult.videoMimeType).isEqualTo(MimeTypes.VIDEO_DOLBY_VISION);
```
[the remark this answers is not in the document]
ybai001: I just followed Line 1280 and 1333 in transcode_withOutputVideoMimeTypeAv1_completesSuccessfully() and transcode_withOutputAudioMimeTypeAac_completesSuccessfully() respectively.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/TransformerEndToEndTest.java
```diff
@@ -1280,6 +1280,38 @@ public void transcode_withOutputVideoMimeTypeAv1_completesSuccessfully() throws
...
+    ExportTestResult exportTestResult =
+        new TransformerAndroidTestRunner.Builder(context, transformer)
+            .build()
+            .run(testId, editedMediaItem);
+    ExportResult exportResult = exportTestResult.exportResult;
+
+    String actualMimeType =
+        retrieveTrackFormat(context, exportTestResult.filePath, C.TRACK_TYPE_VIDEO).sampleMimeType;
+    assertThat(actualMimeType).isEqualTo(MimeTypes.VIDEO_DOLBY_VISION);
+    assertThat(exportResult.exportException).isNull();
```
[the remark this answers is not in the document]
ybai001: I just followed Line 1278 and 1331 in transcode_withOutputVideoMimeTypeAv1_completesSuccessfully() and transcode_withOutputAudioMimeTypeAac_completesSuccessfully() respectively.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -139,6 +140,43 @@ public void export_transmuxHlg10File() throws Exception {
...
+    MediaItem mediaItem = MediaItem.fromUri(Uri.parse(MP4_ASSET_DOLBY_VISION_HDR));
+
+    ExportTestResult exportTestResult =
+        new TransformerAndroidTestRunner.Builder(context, transformer)
+            .build()
+            .run(testId, mediaItem);
+    @C.ColorTransfer
+    int actualColorTransfer =
+        retrieveTrackFormat(context, exportTestResult.filePath, C.TRACK_TYPE_VIDEO)
+            .colorInfo
```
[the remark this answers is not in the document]
ybai001: OK, I'll change it.

libraries/transformer/src/main/java/androidx/media3/transformer/FrameworkMuxer.java
```diff
@@ -276,9 +283,64 @@ private static ImmutableList<String> getSupportedVideoSampleMimeTypes() {
...
     }
     return supportedMimeTypes.build();
   }
+
+  // Get Dolby Vision profile
+  // Refer to https://professionalsupport.dolby.com/s/article/What-is-Dolby-Vision-Profile
+  @RequiresApi(33)
+  private static int getDvProfile(Format format) {
+    // Currently, only profile 8 is supported for encoding
+    // TODO: set profile ID based on format.
```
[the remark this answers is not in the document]
ybai001: Currently, only profile 8 (H.265 based) is supported for encoding. In future, maybe Dolby will support other Dolby Vision profile (e.g. profile 9/10). It's not finalized. If that is true, we need to add logic here to decide which profile should be used based on the parameter "format".

libraries/transformer/src/main/java/androidx/media3/transformer/FrameworkMuxer.java
ybai001: Because the codec profile/level constants should be a bitmask, which is defined by Android AOSP. Dolby defined our profile/level to follow this rule. See https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/media/java/android/media/MediaCodecInfo.java;drc=6a4bef0f90e822e19866e53a98b85029bff04ea0;l=4301

I think Google defined it in this way due to efficiency. For example, to judge whether one profile is supported by current device, you just need to do one time comparison:
if (appointedLevel & (DolbyVisionLevelFhd24 | DolbyVisionLevelFhd24) {
    // Current device support this level
} else {
    // Current device doesn't support this level
}

libraries/transformer/src/main/java/androidx/media3/transformer/FrameworkMuxer.java
ybai001: OK.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
```diff
@@ -139,6 +140,43 @@ public void export_transmuxHlg10File() throws Exception {
     assertThat(actualColorTransfer).isEqualTo(C.COLOR_TRANSFER_HLG);
   }
 
+  @Test
+  public void export_transmuxDolbyVisionFile() throws Exception {
+    Context context = ApplicationProvider.getApplicationContext();
+
+    if (Util.SDK_INT < 24) {
```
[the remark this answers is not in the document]
ybai001: OK, I'll change the code.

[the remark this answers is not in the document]
icbaker: You can also use `@SdkSuppress(minSdkVersion = 24)` if your only condition is SDK level.

https://developer.android.com/reference/androidx/test/filters/SdkSuppress

This is preferable to `assumeTrue` imo because it skips the test "earlier", and just doesn't run it at all, instead of needing to mark it as 'skipped' or 'passed' (which has mixed support in different test running environments).

ychaparov:
Looks good to me. @SheenaChhabra please approve when you think your comments have been resolved

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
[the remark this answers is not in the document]
ybai001: OK, I think this method is better. I'll change the code again.

ybai001:
@SheenaChhabra, I have updated code as your comments except resolving conflicts. Let me know whether there is any other comment from your side.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/TransformerEndToEndTest.java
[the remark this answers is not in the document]
SheenaChhabra: I am not sure why existing tests has this assertion but we can remove it for the newly added tests.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
SheenaChhabra: I think this is not fully resolved as the decoder check (AndroidTestUtil.skipAndLogIfFormatsUnsupported) is still present which should not be required for transmuxing.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
SheenaChhabra: This is still not resolved.

libraries/transformer/src/main/java/androidx/media3/transformer/DefaultDecoderFactory.java
SheenaChhabra: This comment seems unresolved.

SheenaChhabra (changes requested):
Overall changes are good but some of the comments are still not resolved.
Once the comments are resolved/answered and the conflicts are resolved, we can take this forward.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/TransformerEndToEndTest.java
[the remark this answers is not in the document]
ybai001: OK, I'll remove it.

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: OK, I'll remove this part of checking.

    if (AndroidTestUtil.skipAndLogIfFormatsUnsupported(
        context,
        testId,
        /* inputFormat= */ MP4_ASSET_DOLBY_VISION_HDR_FORMAT,
        /* outputFormat= */ null)) {
      return;
    }

libraries/transformer/src/main/java/androidx/media3/transformer/DefaultDecoderFactory.java
ybai001: This is important for DV editing/transcoding use case. As explained in my first comment,

Dolby Vision is not only a HDR type but also a media type. But it is based on H.264, H.265 or AV1. That is to say, the Dolby Vision stream is base encoded data + Dolby Vision metadata. Dolby Vision decoder will do two things in playback use case: (1) Call base codec, e.g. H.265 decoder, to decode the base encoded data; (2) Apply Dolby Vision metadata to decoded base data (output of step 1) to get Dolby Vision experience.

In editing use case, Dolby Vision decoder does step 1 only. We don't apply the Dolby Vision metadata because the result of this process is related to the screen characteristic and environment brightness. For editing use case, the Dolby Vision metadata will be generated by Dolby Vision encoder. For transcoding use case, this is a trade off between better quality and consistent result on different devices.

By default, Dolby Vision decoder will do both above step 1 & 2. This is correct for playback use case.
But for editing/transcoding use case, Dolby Vision decoder to do step 1 only. (ignoring Dolby Vision metadata.) That is the purpose of
mediaFormat.setInteger(MediaFormat.KEY_COLOR_TRANSFER_REQUEST, MediaFormat.COLOR_TRANSFER_HLG);
which notifies Dolby Vision decoder to do step 1 only. 

Does this answer your question?

libraries/transformer/src/androidTest/java/androidx/media3/transformer/mh/HdrEditingTest.java
ybai001: OK, I'll change the code to use 

    default void onFallbackApplied(
        Composition composition,
        TransformationRequest originalTransformationRequest,
        TransformationRequest fallbackTransformationRequest) {
      MediaItem mediaItem = composition.sequences.get(0).editedMediaItems.get(0).mediaItem;
      onFallbackApplied(mediaItem, originalTransformationRequest, fallbackTransformationRequest);
    }

ybai001:
> Overall changes are good but some of the comments are still not resolved. Once the comments are resolved/answered and the conflicts are resolved, we can take this forward.

I have updated the code as you requested and resolved the conflicts. Let's me know whether there is any other issue.

ybai001:
After resolving the merge conflict, I find the code base is updated to the latest one and I have to update my test code to match the latest one. (Some definitions are changed totally, e.g. MP4_ASSET.
I'll refine my code and upload it.

ybai001:
@SheenaChhabra, I updated code based on latest code on main branch. I think I resolved all the conflicts. Let me know whether you have more questions. Sorry for tons of commits and comments today.

SheenaChhabra (approved):
Thanks for making this change.

ybai001:
> Thanks for making this change. I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks!

Thanks. BTW, just want to let you know:
Google can't push changes to this organization-owned forks because Dolby didn't grant you collaborator access. I have applied for it to our Dolby admin but she didn't agree with it. As result, you can merge this PR but it will result in an 'evil' merge. See details here: https://github.com/androidx/media/blob/release/CONTRIBUTING.md#push-access-to-pr-branches

ybai001:
> Thanks for making this change. I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks!

Excuse me, may I know the status of this PR? :)

SheenaChhabra:
Apologies for the delay. While the changes looks good, the test failed on a few devices which needs further investigation. The plan is to merge this by the end of this quarter. 
Thanks for your patience!!

ybai001:
> Apologies for the delay. While the changes looks good, the test failed on a few devices which needs further investigation. The plan is to merge this by the end of this quarter. Thanks for your patience!!

Thanks for info update.

PR #1255, opened 2024-04-09
Fix album title metadata in media item conversion

phcannesson:
Spotted a small issue in `MediaItem` conversion from cast `MediaQueueItem`


marcbaechinger:
Thanks!

PR #1265, opened 2024-04-11
Add AC-4 Level-4 ISO base media file format support

ybai001:
This pull request is same as https://github.com/androidx/media/pull/875. Since 875 is broken due to too many commit numbers, I re-created this new one.
 
========================

AC-4 Level-4 (Object Based Audio) is a Dolby official audio stream type for Atmos Music. This code update enables Android Media3 support local MP4 and MPEG-DASH AC-4 Level-4 playback.


libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java
```diff
@@ -113,13 +210,335 @@ private SyncFrameInfo(
...
+      }
+
+      ac4Presentations.put(presentationIndex, ac4Presentation);
+    }
+
+    int channelCount = -1;
+    // Using first presentation (default presentation) channel count
+    int presentationIndex = 0;
+    Ac4Presentation ac4Presentation =
+        Objects.requireNonNull(ac4Presentations.get(presentationIndex));
```
rohitjoins, line 520: @ybai001,

Why are we looping over all `presentationIndex` above and storing it in a `Map` when all we require is the `ac4Presentation` for the index `0`?

Is it to verify if the file doesn't have bad data? Even in that case storing it in a `Map` seems an overkill, when we don't intend to use the map for anything else but index `0`.

ybai001: @rohitjoins, 
This AC4 parser implementation is similar as the one in Android Framework: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/module/extractors/mp4/AC4Parser.cpp;drc=2e87c835abea90b0cf61b42dc9aa8a882f2ef5aa;l=292. It parses all the information for all presentations because there is API https://developer.android.com/reference/android/media/MediaExtractor#getAudioPresentations(int) to fetch the presentations info for appointed audio track.

Since AndroidX Media extractor doesn't support AudioPresentation, I know this loop is redundant now. 

But I know Android framework MPEGExtractor was deprecated and it will be replaced by AndroidX Media extractor. Although AndroidX Media doesn't support AudioPresentation now but I believe it will support it sooner or later. That's why I use a Map to store all audio presentation info and keep this logic in my implementation.

If one audio track includes multiple audio presentations, and if application doesn't set which presentation should be played, the Dolby AC-4 decoder will decode the default presentation (the first presentation).

rohitjoins: @ybai001,

I would suggest to remove any logic which will be used in the future as it can be added when required. For now we should only parse and use the `first presentation`.

ybai001: @rohitjoins , OK, I'll update the code.

ybai001: @rohitjoins, do you have any update on this pull request?

rohitjoins: Hi @ybai001,

I'll take a look and send this PR for internal review.

sqshet-dlb: Hi @rohitjoins ,

Do you have further updates on this CL. We have more support added for AC-4 presentations parsing for different containers that are dependent on this pull request. I am looking to create new pull requests once this pull request is reviewed and merged.

Thanks!

ybai001: Hi, @rohitjoins,

Any update on this PR for internal review? Thanks.

sqshet-dlb: Hi @rohitjoins,

Just wondering if this CL is good to go. Please let us know if you need any further clarification needed regarding the changes here.

We have few other dependent changes in the pipeline that needs to be pushed up for review once this CL gets merged.
 
Thanks

ybai001: > Hi @sqshet-dlb,
> 
> I’m currently making some updates to the PR, such as adding more documentation and refactoring some parts into new methods before sending it for internal review. I’ll keep you updated on the progress and aim to merge it soon.

Hi, @rohitjoins, as you mentioned on July 2nd,
_I'll take a look and send this PR for internal review._
It's nearly 3 months. Could you help to explain why this pull request has not been sent for internal review? Could you give me a deadline? This is very important for us since it blocks several other contributions as @sqshet-dlb said.

ybai001: Hi, @rohitjoins 
Could you help to provide us any update about this contribution? 
Thanks.

ybai001: @rohitjoins , there is a merge conflict with IAMF code. Do I need to resolve this conflict? Because you said you would send this for internal code review, I'm not sure whether I should modify code again. Thanks.

rohitjoins: Hey @ybai001,

No need to update this PR as I already do it for the internal version. If no blockers come up in review, this should be merged this week.

ybai001: @rohitjoins, thanks for the quick feedback. :)

rohitjoins:
Hi @ybai001,

The PR has been merged as https://github.com/androidx/media/commit/74611bbdc009b3c143a556a74f0e84f386fb5e19.

I refactored the code to remove elements from `AudioPresentations` that aren’t currently needed, as we’re not returning it yet. Please take a look at the merged commit and let me know if you spot any issues.

ybai001:
> Hi @ybai001,
> 
> The PR has been merged as [74611bb](https://github.com/androidx/media/commit/74611bbdc009b3c143a556a74f0e84f386fb5e19).
> 
> I refactored the code to remove elements from `AudioPresentations` that aren’t currently needed, as we’re not returning it yet. Please take a look at the merged commit and let me know if you spot any issues.

Hi, @rohitjoins

I truly appreciate your effort on this pull request. It's almost one year! :) I understood this was a big change in extractor and it was related to Dolby specific format that you were not familiar with. I have verified it on my test device based on the lastest main branch code. It can work. So thanks again!

PR #1299, opened 2024-04-20
Fix ENCODING_DTS_UHD_P2 Issue

cedricxperi:
During Direct Passthrough Audio Playback on platforms that supports AudioFormat.ENCODING_DTS_UHD_P2, we need to include C.ENCODING_DTS_UHD_P2 in DefaultAudioSink.getFramesPerEncodedSample() and DefaultAudioTrackBufferSizeProvider.getMaximumEncodedRateBytesPerSecond().


cedricxperi:
@microkatz Here is the new PR that merges into main branch. Thanks.

microkatz:
@cedricxperi 

Thanks for creating the new PR! One quick point, I think you may have missed a spot in the `DefaultAudioTrackBufferSizeProvider.getPassthroughBufferSizeInBytes` method? Would ENCODING_DTS_UHD_P2 need the same multiple for the buffersize computation for passthrough as ENCODING_DTS_HD?

cedricxperi:
> @cedricxperi
> 
> Thanks for creating the new PR! One quick point, I think you may have missed a spot in the `DefaultAudioTrackBufferSizeProvider.getPassthroughBufferSizeInBytes` method? Would ENCODING_DTS_UHD_P2 need the same multiple for the buffersize computation for passthrough as ENCODING_DTS_HD?

@microkatz Thanks for pointing that out. However, DTS:X Profile 2 has a smaller number of audio samples per channel per frame compared to DTS-HD (DTS Express). So it does not require a bigger multiple for buffersize computation.  :)

microkatz:

PR #1300, opened 2024-04-20
Fix DASH thumbnails cropping the incorrect tile for non-square images

hakonschia:
The calculation for finding the column a `tileIndex` corresponds to is currently `tileIndex % inputFormat.tileCountVertical`. This should instead be `tileIndex % inputFormat.tileCountHorizontal` (see [this](https://stackoverflow.com/a/11821923/7750841) SO answer) to work for non-square images. It works for square images since `tileCountVertical == tileCountHorizontal`, so the same values are produced.

Using `tileCountVertical` for non-square images results in either:
- When `tileCountVertical` is equal to 1 (sprites with one row of images) it always crops out the first tile
- When `tileCountVertical` is larger than 1 it will attempt to crop out a bitmap outside the bounds of the larger bitmap, resulting in `Bitmap.createBitmap()` throwing an exception.

<details>
<summary>Exception thrown</summary>

```
Playback error
  androidx.media3.exoplayer.ExoPlaybackException: Unexpected runtime error
   at androidx.media3.exoplayer.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:703)
   at android.os.Handler.dispatchMessage(Handler.java:103)
   at android.os.Looper.loopOnce(Looper.java:232)
   at android.os.Looper.loop(Looper.java:317)
   at android.os.HandlerThread.run(HandlerThread.java:85)
 Caused by: java.lang.IllegalArgumentException: x + width must be <= bitmap.width()
   at android.graphics.Bitmap.createBitmap(Bitmap.java:910)
   at android.graphics.Bitmap.createBitmap(Bitmap.java:869)
   at androidx.media3.exoplayer.image.ImageRenderer.cropTileFromImageGrid(ImageRenderer.java:575)
   at androidx.media3.exoplayer.image.ImageRenderer.drainOutput(ImageRenderer.java:343)
   at androidx.media3.exoplayer.image.ImageRenderer.render(ImageRenderer.java:182)
   at androidx.media3.exoplayer.ExoPlayerImplInternal.doSomeWork(ExoPlayerImplInternal.java:1114)
   at androidx.media3.exoplayer.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:546)
   at android.os.Handler.dispatchMessage(Handler.java:103) 
   at android.os.Looper.loopOnce(Looper.java:232) 
   at android.os.Looper.loop(Looper.java:317) 
   at android.os.HandlerThread.run(HandlerThread.java:85) 
```

</details>

I have updated `ImageRenderTest` to use a 2x3 image instead of 2x2. Reverting my change will make the updated tests fail, as expected, but please verify they still work as intended.

I have created a [separate branch](https://github.com/hakonschia/media/tree/fix-dash-thumbnails-cropping-demo/) that displays thumbnails in the demo app that can be used for testing. Two samples are added, one for the one row case and one for non-squares with more than one row. The videos below are taken from this branch.

<details>
<summary>Before</summary>

https://github.com/androidx/media/assets/33124564/3f5ee00c-8a97-49bf-9398-12c9960212f6

</details>

<details>
<summary>After</summary>

https://github.com/androidx/media/assets/33124564/63d7f909-acee-440d-b369-8358c798fe69

</details>

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -45,6 +45,8 @@
     *   Propagate ID3 `TCON` frame to `MediaMetadata.genre`
         ([#1305](https://github.com/androidx/media/issues/1305)).
 *   Image:
+    *   Add support for non-square DASH thumbnail grids
+        ([#1255](https://github.com/androidx/media/pull/1300)).
 *   DRM:
     *   Allow setting a `LoadErrorHandlingPolicy` on
         `DefaultDrmSessionManagerProvider`
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/image/ImageRenderer.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/image/ImageRenderer.java
@@ -571,7 +571,7 @@ public class ImageRenderer extends BaseRenderer {
     checkStateNotNull(outputBitmap);
     int tileWidth = outputBitmap.getWidth() / checkStateNotNull(inputFormat).tileCountHorizontal;
     int tileHeight = outputBitmap.getHeight() / checkStateNotNull(inputFormat).tileCountVertical;
-    int tileStartXCoordinate = tileWidth * (tileIndex % inputFormat.tileCountVertical);
+    int tileStartXCoordinate = tileWidth * (tileIndex % inputFormat.tileCountHorizontal);
     int tileStartYCoordinate = tileHeight * (tileIndex / inputFormat.tileCountHorizontal);
     return Bitmap.createBitmap(
         outputBitmap, tileStartXCoordinate, tileStartYCoordinate, tileWidth, tileHeight);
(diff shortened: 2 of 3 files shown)


hakonschia:
@microkatz I've created a new test now instead, but I'm not really if what I did with `overridenBitmap` is the best way of using a different bitmap in this test

Let me know if you'd prefer to have these two commits squashed to avoid `ImageRenderTest` having changes in the first commit that were just reverted again in the second commit

hakonschia:
@microkatz sure, that seems like a better solution to me, I've pushed that change now. I just copied `queuingImageOutput`, but it might make more sense for this to be a moved to a class variable at this point instead of duplicating it

microkatz:

RELEASENOTES.md
```diff
@@ -40,6 +40,8 @@
     *   Propagate ID3 `TCON` frame to `MediaMetadata.genre`
         ([#1305](https://github.com/androidx/media/issues/1305)).
 *   Image:
+    *   Add support for non-square DASH thumbnail grids
+        ([#1255](https://github.com/androidx/media/pull/1300)).
```
hakonschia, line 44: @microkatz The PR number here doesn't match the actual PR (and issue 1255 is unrelated to this PR)

PR #1315, opened 2024-04-25
CEA-708: Update current row value when new line is added

datdoantelus:
Update current row value when new line is added.  This is to prevent an incorrect newline to be added in setPenLocation.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -507,6 +507,9 @@ This release includes the following changes since the
         `rowLock` and `columnLock` should both be assumed to be true, regardless
         of the values present in the stream (`columnLock` support is not
         implemented, so it's effectively assumed to always be false).
+    *   CEA-708: Avoid duplicate newlines being added by ExoPlayer's naive
+        handling of the 'set pen location' command
+        ([#1315](https://github.com/androidx/media/pull/1315)).
 *   Image:
     *   Add support for DASH thumbnails. Grid images are cropped and individual
         thumbnails are provided to `ImageOutput` close to their presentation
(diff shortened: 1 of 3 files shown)


icbaker:
Ah it looks like we previously merged the `rowLock/columnLock` fix (from you) in https://github.com/androidx/media/pull/942, and then it got accidentally reverted (by me) in https://github.com/androidx/media/commit/dd2e4a59780dd85392b2023915e40fcca14e55df. I'm really sorry about that. As this part is more like a fix on top of https://github.com/androidx/media/commit/dd2e4a59780dd85392b2023915e40fcca14e55df, rather than an 'original' change, I'm going to make this fix internally rather than directly merge this PR - the end result should be the same, but it will be a bit easier for me to tell the story of how this change ended up getting lost (for future code archaeologists). Hope that's OK!

I think 'Fix 2' in this PR wasn't part of https://github.com/androidx/media/pull/942 - so maybe you could update this PR to only contain 'Fix 2'?

icbaker:
I've re-applied the `rowLock/columnLock` fix in https://github.com/androidx/media/commit/e2847b3b80112e7bce0e0706604bdee8126704e5

icbaker:
> This is to prevent an incorrect newline to be added in setPenLocation.

Would you be able to provide a bit more info about this? Even better if you can craft a test case that demonstrates the issue that is fixed by this change.

In particular I'm trying to understand how this interacts with the existing `this.row = row` assignment in `setPenLocation()` - I wonder if this assignment should be removed as part of this change?

icbaker:

PR #1339, opened 2024-05-03
Allow LoadControl.shouldContinueLoading accept playWhenReady as a parameter

colinkho:
Motivation: Load Control requires different min and max buffer values depending if the player is playing or paused.


colinkho:
Amended commit and force pushed to correct username to allow detection of CLA

adarsh-dhakad (approved):

tonihei:
Thanks, I'll continue the merge process internally and will probably force-push too with additional formatting updates etc.

tonihei:
This is merged now. You may have seen it already, but we updated the pull request to move the parameters class inside `LoadControl` itself and also used it for `shouldStartPlayback`.

PR #1367, opened 2024-05-11
Fix PGS subtitles rendered with border and no fill color

TheBeastLT:
Since #1006 has seen no activity for some time, I thought I'd apply the commit from the correct base, since this bug is quite annoying and has been there for quite a while


icbaker:

PR #1389, opened 2024-05-23
Add DV profile 10 handling in getAlternativeCodecMimeType() method

ybai001:
I found your team had added DV profile 10 support in https://github.com/androidx/media/blob/main/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecUtil.java#L1346 in latest main branch. But it seems that method getAlternativeCodecMimeType() should be updated too. So I submitted this contribution.


microkatz:
Hi @ybai001 

ybai001:
> Hi @ybai001
> 
> Would it be possible to open a new PR from an individual-owned fork? We can't push changes to organization-owned forks like this one unless we have collaborator access. If that's not possible then we can still merge this PR but it will result in an ['evil' merge](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefevilmergeaevilmerge). See more info here: https://github.com/androidx/media/blob/release/CONTRIBUTING.md#push-access-to-pr-branches

Hi, @microkatz , Yes, I was told about it in another pull request. Unfortunately, due to the company process limitation, I can't do it now. So please handle it as an evil merge. At the same time, we are trying to solve this situation in our company internally so that we can do it as you said in future. But it can't be done in short time.

PR #1416, opened 2024-05-31
Add a CustomData field to the Format class

khouzam:
Summary:
This change aims to add a generic `CustomData` field to the `Format` class.

The intent is to allow ExoPlayer customers to add extra data to the Format class without forcing specific data to be included, impacting customers that do not need it and would allow for the data to be changed without requiring changes to the `Media3` codebase.

The `CustomDataAllocator` allows for the field to be pre-populated with a default object the same way that the rest of the `format` fields are.


Tolriq:
If this is accepted can you see https://github.com/androidx/media/pull/594 so that you pass that custom data down to the audio sinks too as it would solve a few use cases.

tonihei:
> The CustomDataAllocator allows for the field to be pre-populated with a default object the same way that the rest of the format fields are.

Is this important for your use case? I think it's cleaner to add add this static init class and instead rely on the the code that creates Format objects to populate the right fields.

khouzam:
> > The CustomDataAllocator allows for the field to be pre-populated with a default object the same way that the rest of the format fields are.
> 
> Is this important for your use case? I think it's cleaner to add add this static init class and instead rely on the the code that creates Format objects to populate the right fields.

Thanks for bringing this up. The initial intent of the `allocator` was to ensure that the `customData` was always allocated to help keep the code clean whenever we would access the field. Since we would have to cast already every time we access a field, having to add a null check would have made the code much more cumbersome than needed.

But after some thoughts, having a method, to perform the null check, and return the proper default object if it hasn't been populated, lets us keep our code clean and doesn't require the `customAllocator` anymore.

tonihei:
Thanks for updating the PR, I'll merge it internally now and you'll see additional changes appearing here (mostly formatting).

PR #1437, opened 2024-06-09
Add support for setting the max seek to previous position on `ExoPlayer.Builder` and `CastPlayer`

### LINKED ISSUE #1425: Add the setter counter part of `Player.getMaxSeekToPreviousPosition()`
### Use case description

Currently, the `Player` interface provides a `getMaxSeekToPreviousPosition()` method, which returns a delay after which the player seeks to the previous item, instead of restarting the current item, when calling `Player.seekToPrevious()`. That delay is currently hardcoded to `C.DEFAULT_MAX_SEEK_TO_PREVIOUS_POSITION_MS` (3 seconds).
There is also a `Player.Listener.onMaxSeekToPreviousPositionChanged(long)` method, which is called when the delay changes. However, as far as we could see, there is currently no `Player.setMaxSeekToPreviousPosition(long)` method, allowing us to customise that delay.

### Proposed solution

Provide a `Player.setMaxSeekToPreviousPosition(long)` method se the delay can be freely customised by developers.
If this request seems reasonable to you, I'm open to work on a PR to help integrate it in Media3.

### Alternatives considered

Keep working with the current default delay.

MGaetan89:
This PR introduces the following changes:
- Add a new constructor to `CastPlayer` to provide a `maxSeekToPreviousPositionMs` argument.
- Add `ExoPlayer.Builder.setMaxSeekToPreviousPosition(long).
- Update `MediaControllerImplLegacy.getMaxSeekToPreviousPosition()` to return the value from `controllerInfo.playerInfo.maxSeekToPreviousPositionMs` instead of 0.
- Add `TestExoPlayerBuilder.setMaxSeekToPreviousPosition(long)`.
- Add `TestExoPlayerBuilder.getMaxSeekToPreviousPosition()`.

> [!NOTE]
> This new functionality only allows setting the value at initialisation (for `ExoPlayer` and `CastPlayer`). Let me know if I should also add a dedicated setter on these classes (or on `Player` directly).

Fixes #1425

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -6,6 +6,9 @@
 *   ExoPlayer:
     *   Add `ExoPlayer.isReleased()` to check whether `Exoplayer.release()` has
         been called.
+    *   Add `ExoPlayer.Builder.setMaxSeekToPreviousPositionMs` to configure the
+        maximum position for which `seekToPrevious()` seeks to the previous item
+        ([#1425](https://github.com/androidx/media/issues/1425)).
 *   Transformer:
     *   Remove `ExportResult.processedInputs` field. If you use this field for
         codec details, then use `DefaultDecoderFactory.listener` instead. In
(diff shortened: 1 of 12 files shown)


tonihei:
Thanks, I'll merge it internally and you may see some additional commit appearing here in the meantime (mostly formatting changes).

PR #1479, opened 2024-06-21
Fix potential side effects inside of the player.

dryganets:
ExoPlayer uses handlers extensively.
One of the hidden problem of handlers is removal of the messages with ID == 0.
There is a side effect of such a removal - it also removes all callbacks.

[postDelayed](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/os/Handler.java;l=498-500?q=Handler), creates a message through [Message.obtain](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/os/Message.java;l=196-201) call. What value is always 0 as a result.

So [removeMessages(what)](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/os/Handler.java;l=785-787?q=Handler) will drop the callbacks as well.

Here I'm fixing the cases what remove the 0 message in particular.
I expect some of the problems in the playback stack miraculously disappear as a result of the change.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
@@ -326,10 +326,10 @@ public final class Loader implements LoaderErrorThrower {
 
     private static final String TAG = "LoadTask";
 
-    private static final int MSG_START = 0;
-    private static final int MSG_FINISH = 1;
-    private static final int MSG_IO_EXCEPTION = 2;
-    private static final int MSG_FATAL_ERROR = 3;
+    private static final int MSG_START = 1;
+    private static final int MSG_FINISH = 2;
+    private static final int MSG_IO_EXCEPTION = 3;
+    private static final int MSG_FATAL_ERROR = 4;
 
     public final int defaultMinRetryCount;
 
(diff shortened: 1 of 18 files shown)


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
```diff
@@ -326,10 +326,10 @@ private final class LoadTask<T extends Loadable> extends Handler implements Runn
 
     private static final String TAG = "LoadTask";
 
-    private static final int MSG_START = 0;
-    private static final int MSG_FINISH = 1;
-    private static final int MSG_IO_EXCEPTION = 2;
-    private static final int MSG_FATAL_ERROR = 3;
+    private static final int MSG_START = 1;
```
dryganets, line 329: This one is likely to cause some hidden issues as the same looper/handler is used to post runnable in the code.

tonihei:
Thanks for bringing this up! We weren't aware of this rather unexpected side effect of using messages with `what==0`. I assume nothing is actually broken at the moment because none of the cases that use `removeMessages` also uses `Runnables` at the same time. We'll still update the message number as a precaution. I'll start with your pull request and then extend it further. You may see some updates appearing here while I do this and merge the change internally.

PR #1487, opened 2024-06-23
Avoid unnecessary getSystemService call on WifiLockManager and WakeLo…

colinkho:
…ckManager if not enabled

getSystemService is expensive during initialization if it is never used. This delays the need to make the getSystemService call to only instances when the lock managers are actually needed


icbaker:
Thanks for the PR!

tonihei:
The original ANR this change is trying to avoid is also fixed completely in Media3 1.6.0 now

Issue #1483, opened 2024-06-22
Potential for Multiple Non-Internal Discontinuity Events


skidder:
### Version

Media3 1.1.1 / ExoPlayer 2.19.1

### Devices that reproduce the issue

Observed on Android OS 11-14, could include more.

### Reproducible in the demo app?

Not tested

### Reproduction steps

1. Implement a playback scenario with multiple periods or ads, ensuring the playback queue has distinct playing and reading periods.
2. Induce a renderer error during playback, specifically when the playing period differs from the reading period.
3. Observe the handling of discontinuity events following the renderer error, particularly the reporting of non-internal discontinuity reasons (Player.DISCONTINUITY_REASON_AUTO_TRANSITION) in quick succession or across different parts of the playback queue.

It's also manually reproducible in unit tests by instantiating `PlaybackInfoUpdate` and invoking `setPositionDiscontinuity` multiple times with non-internal events, i.e.
```java
    // rapidly set position discontinuity
    playbackInfoUpdate.setPositionDiscontinuity(Player.DISCONTINUITY_REASON_SEEK);
    playbackInfoUpdate.setPositionDiscontinuity(Player.DISCONTINUITY_REASON_INTERNAL);
    playbackInfoUpdate.setPositionDiscontinuity(Player.DISCONTINUITY_REASON_SEEK);
```

Here's a draft PR showing the issue https://github.com/google/ExoPlayer/pull/11392

### Expected result

No assert or unit test failures.

### Actual result

In production we've seen app crashes with the following stack-trace:
```
java.lang.IllegalArgumentException: null
    at com.google.android.exoplayer2.util.Assertions.checkArgument
    at com.google.android.exoplayer2.ExoPlayerImplInternal$PlaybackInfoUpdate.setPositionDiscontinuity(SourceFile:121)
    at com.google.android.exoplayer2.ExoPlayerImplInternal.handlePositionDiscontinuity(SourceFile:2489)
    at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(SourceFile:626)
    at android.os.Handler.dispatchMessage(Handler.java:102)
    at android.os.Looper.loopOnce(Looper.java:210)
    at android.os.Looper.loop(Looper.java:299)
    at android.os.HandlerThread.run(HandlerThread.java:67)
```

[...]

PR #1491, opened 2024-06-23
[exoplayer] Fix to handle multiple concurrent discontinuity updates.

skidder:
## Summary
This pull request fixes issue #1483  where multiple concurrent discontinuity updates can lead to an `IllegalArgumentException` in ExoPlayer and potentially an app-crash. The issue arises when the player encounters renderer errors while the playing period differs from the reading period, leading to multiple non-internal discontinuity reasons being reported in quick succession. This regression was introduced with commit 79b688ef30d4a3306d4d321ddf4464b428074ea2, having added another path to report non-internal discontinuity.

### Modification to ExoPlayerImplInternal:
 * Adjusted the `setPositionDiscontinuity` method in `PlaybackInfoUpdate` to handle multiple concurrent discontinuity events more gracefully by allowing multiple non-internal discontinuity reasons, though only reporting the first seen. Implemented by removing `assert` statement that no longer holds true.

### Added Unit Test:
 * Introduced a unit test to ensure that multiple non-internal discontinuity reasons are handled without causing assertion failures.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/ExoPlayerImplInternal.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/ExoPlayerImplInternal.java
@@ -119,9 +119,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
     public void setPositionDiscontinuity(@DiscontinuityReason int discontinuityReason) {
       if (positionDiscontinuity
           && this.discontinuityReason != Player.DISCONTINUITY_REASON_INTERNAL) {
-        // We always prefer non-internal discontinuity reasons. We also assume that we won't report
-        // more than one non-internal discontinuity per message iteration.
-        Assertions.checkArgument(discontinuityReason == Player.DISCONTINUITY_REASON_INTERNAL);
+        // We always prefer non-internal discontinuity reasons. Multiple non-internal discontinuity
+        // reasons are possible, but rare. Report the first reason seen.
         return;
       }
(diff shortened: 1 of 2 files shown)


tonihei:
Thanks for the PR. I sent a slightly different fix for internal review already. We don't really want to overwrite the discontinuity reasons (unless they are internal) to not lose important state changes, that's also the reason for this assertion in the first place. The new fix sends out any pending updates first before setting a new discontinuity reason, which should avoid the problematic situation from happening. I'll close this PR, but thanks again for looking into this so quickly and proposing a fix!

skidder:
> Thanks for the PR. I sent a slightly different fix for internal review already.

Nice, good to hear the issue has been validated and a fix is on the way. 

Do you have an estimate of when the fix will be made public? Also, is it easily backported to Exoplayer 2.19.1? If not, would you recommend using an earlier Exoplayer release instead (i.e. 2.19.0)?

tonihei:
Backporting should be easy I think, it's a one-line change ;)  But if you can, please also consider upgrading to the latest version to not miss out on other bug fixes and improvement.

skidder:
> Backporting should be easy I think, it's a one-line change ;) 

🔥 

> But if you can, please also consider upgrading to the latest version to not miss out on other bug fixes and improvement.

Definitely, that would be my preference, though I'm currently using an older release of react-native-video that depends on Exoplayer 2. Appreciate the work y'all are doing on the newer Media3 releases!

PR #1549, opened 2024-07-15
Clean up code now that min SDK version is 21

MGaetan89:
This PR does some cleanup following the bump of the min SDK version from 19 to 21:
- Make sure to use `project.ext.minSdkVersion` wherever possible, instead of hardcoding the min SDK version.
- Remove usages of Multidex in every modules.
- Remove some legacy `SDK_INT` checks.


MGaetan89:
I'm checking if `MediaSessionImplBase`, `MediaSessionImplApi18`, and `MediaSessionImplApi19` can also be removed. They appeared to be unused, so I guess they are remains of when the min SDK version was bumped to 19.

MGaetan89:
> I'm checking if `MediaSessionImplBase`, `MediaSessionImplApi18`, and `MediaSessionImplApi19` can also be removed. They appeared to be unused, so I guess they are remains of when the min SDK version was bumped to 19.

Should be good for review now 👍🏻

icbaker:
Thanks for the extra fixes! Please could you split out the changes in `androidx.media3.session.legacy` so we can park those for now? A lot of this code is already unused, so it makes sense to remove larger swathes of it before removing just these API-guarded bits.

MGaetan89:
Sure, I'll do that 👍🏻 
Should I submit the changes to `androidx.media3.session.legacy` in a dedicated PR, or just remove them from this one?

icbaker:
I think just remove them from this one, I don't think it's worth splitting them to another PR because there will be so many merge conflicts after we remove the 'wider' unused code.

MGaetan89:
> I think just remove them from this one, I don't think it's worth splitting them to another PR because there will be so many merge conflicts after we remove the 'wider' unused code.

The changes have been reverted. I'll also rebase the PR on `main` to integrate the latest changes.

icbaker:

PR #1566, opened 2024-07-26
Refactor getMaxVideoSizeInViewport into TrackSelectionUtil so it can …

colinkho:
…be reused in external code


icbaker:

PR #1576, opened 2024-07-30
Allow custom ExecutorService to be supplied to the Loader

colinkho:


icbaker:

icbaker:
Internal review has suggested taking `Executor` instead of `ExecutorService`, and therefore avoiding shutting down an externally-provided `ExecutorService`. I've made this change.

PR #1577, opened 2024-07-30
Propagate SubtitleDecoderExceptions to be reported through a callback…

colinkho:
… from TextOutput

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/text/TextOutput.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/text/TextOutput.java
@@ -15,10 +15,12 @@
  */
 package androidx.media3.exoplayer.text;
 
+import androidx.media3.common.Format;
 import androidx.media3.common.text.Cue;
 import androidx.media3.common.text.CueGroup;
 import androidx.media3.common.util.UnstableApi;
 import java.util.List;
+import javax.annotation.Nullable;
 
 /** Receives text output. */
 @UnstableApi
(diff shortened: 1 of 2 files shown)


icbaker:
I realised after discussing this that, since `1.4.0-alpha02`, subtitle decoding/parsing happens during extraction by default - and `TextRenderer` is basically just parsing an internally-defined `Bundle` (which shouldn't fail).

This means that any parsing errors due to invalid subtitle input data happen during extraction, rather than rendering. At the moment these bubble out as fatal errors (which is a behaviour change compared to 1.3.1 and earlier). We probably want to make these errors non-fatal (to match the previous behaviour more closely), and at the same time we can work out how to plumb them out so an app can receive them. At the moment I'm looking at this being done via a new `SubtitleParser.EventListener` interface, which would be added to a `DefaultSubtitleParserFactory` instance and passed into individual `SubtitleParser` instances, which would be responsible for directing any parsing-related errors to the listener (instead of letting them bubble out).

One question I have based on the interface you've proposed in the PR: What is the purpose of the `Format` field? At the moment a `SubtitleParser` instance doesn't have a `Format` object, so we wouldn't easily be able to pass this value into the proposed `SubtitleParser.EventListener.onSubtitleParsingError(Exception)` callback.

colinkho:
Thanks Ian. Not having Format should be ok for us given that we will get an Exception with a trace that we can log. If there's information we are missing after integrating with this new callback, I can raise a PR

Not directly impacting this PR but I am more concerned about 

> At the moment these bubble out as fatal errors (which is a behaviour change compared to 1.3.1 and earlier). We probably want to make these errors non-fatal (to match the previous behaviour more closely)

Is this something that will be done in this release?

PR #1580, opened 2024-08-01
Add a callback on Loader to signal when a Load starts

colinkho:
This includes variables that captures the error and error count if it is starting a retry

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
@@ -91,6 +91,18 @@ public final class Loader implements LoaderErrorThrower {
   public interface Callback<T extends Loadable> {
 
     /**
+     * Called when a load has started for the first time or through a retry.
+     *
+     * @param loadable The loadable whose load has completed.
+     * @param elapsedRealtimeMs {@link SystemClock#elapsedRealtime} when the load ended.
+     * @param loadDurationMs The duration in milliseconds of the load since {@link #startLoading}
+     *     was called.
+     * @param error The load error of the previous load attempt if this is a retry, otherwise null
+     * @param errorCount The number of errors this load has encountered thus far.
+     */
+    default void onLoadStarted(T loadable, long elapsedRealtimeMs, long loadDurationMs, @Nullable IOException error, int errorCount) {}
+
+    /**
      * Called when a load has completed.
      *
      * <p>Note: There is guaranteed to be a memory barrier between {@link Loadable#load()} exiting
@@ -531,6 +543,9 @@ public final class Loader implements LoaderErrorThrower {
     }
 
     private void execute() {
+      long nowMs = SystemClock.elapsedRealtime();
+      long durationMs = nowMs - startTimeMs;
+      Assertions.checkNotNull(this.callback).onLoadStarted(loadable, nowMs, durationMs, currentError, errorCount);
       currentError = null;
       downloadExecutor.execute(Assertions.checkNotNull(currentTask));
     }
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
@@ -97,8 +97,11 @@ public final class Loader implements LoaderErrorThrower {
      * @param elapsedRealtimeMs {@link SystemClock#elapsedRealtime} when the load attempts to start.
      * @param loadDurationMs The duration in milliseconds of the load since {@link #startLoading}
      *     was called.
+     * @param retryCount The number of failed attempts since {@link #startLoading} was called (this
+     *     is zero for the first load attempt).
      */
-    default void onLoadStarted(T loadable, long elapsedRealtimeMs, long loadDurationMs) {}
+    default void onLoadStarted(
+        T loadable, long elapsedRealtimeMs, long loadDurationMs, int retryCount) {}
 
     /**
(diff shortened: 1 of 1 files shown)


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
```diff
@@ -90,6 +90,18 @@ public interface Loadable {
   /** A callback to be notified of {@link Loader} events. */
   public interface Callback<T extends Loadable> {
 
+    /**
+     * Called when a load has started for the first time or through a retry.
+     *
+     * @param loadable The loadable whose load has completed.
+     * @param elapsedRealtimeMs {@link SystemClock#elapsedRealtime} when the load ended.
```
icbaker, line 97: If this method is called to indicate the start of a load, why are we passing the time that it ended (surely that hasn't happened yet, since the end comes after the start)

Similar comment for `loadDurationMs` below - since wouldn't you expect this to be basically zero at this point?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
```diff
@@ -90,6 +90,18 @@ public interface Loadable {
...
+     * @param loadDurationMs The duration in milliseconds of the load since {@link #startLoading}
+     *     was called.
+     * @param error The load error of the previous load attempt if this is a retry, otherwise null
```
icbaker, line 100: `error` and `errorCount` are already passed to `onLoadError` - which I think means an implementation of this interface already has access to this information without needing to pass it here too?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
colinkho: Apologies I did not update the javadoc when I had copied it from onLoadCompleted. I will update it

Basically, we want this method for retries too. So the loadDurationMs is actually the time from when startLoading was first called to when a load was actually executed.

The elapsedRealtimeMs helps give us timestamps when retries occur as well since this can happen many times per load

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
colinkho: Ah good point, we can cache that value on the impl side. Ultimately, I wanted a way to signify that the onLoadStart attempt was a retry or first attempt but agreed it can be tracked internally

colinkho:
@icbaker I have incorporated the comments

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
icbaker: > The elapsedRealtimeMs helps give us timestamps when retries occur as well since this can happen many times per load

I think this information is available from the `elapsedRealtimeMs` value passed to the `onLoadError` call associated with the retry?

> So the loadDurationMs is actually the time from when startLoading was first called to when a load was actually executed.

It seems that this could also be tracked by an keeping track of the state of each load in your implementation?

colinkho: > I think this information is available from the elapsedRealtimeMs value passed to the onLoadError call associated with the retry?

Hmm, does `elapsedRealtimeMs` on `onLoadError` tells us the time when a retry is actually being executed? From the code, it seems indicate the time when an error was thrown and schedule `MSG_START` with a  possible `retryDelay`. If there is no retryDelay, I think we can use that `elapsedRealtimeMs`, however if there is a delay, then are you suggesting we add that delay manually onto whatever we get for `onLoadError`?

> It seems that this could also be tracked by an keeping track of the state of each load in your implementation?

We might but I think it involves passing the `startTimeMs` returned from `startLoading` to the callback after the `LoadTask` has started. It's a little bit messy but doable. Any reason to not include this as a param given that all other load methods feed the same parameter?

colinkho:
Apologies Ian, I accidentally updated my main branch and had newer commits rebased on top. Hope this doesn't complicate the internal review

icbaker:
No worries, i've force pushed my own changes back.

icbaker:
btw I think people often send PRs from **not** the main branch of their forks, possibly for this reason, but as an aside: using the main branch in your fork also slightly upsets GitHub's command line tool because we have to run:

```
$ gh pr checkout 1580 --branch pr-1580
```

Because the regular command will try to update **our** `main` branch directly, which we don't want (and often fails anyway):

```
$ gh pr checkout 1580
```

So for future PRs, it might be useful to create a new well-named branch in your fork for each one.

colinkho:
Thanks. I didn't realize that the branch of the fork continuously updates the PR. I had thought it took a snapshot or merely cherry-picked the commits.

Also good info on the `gh` command line. I've been doing things the old-fashioned `git` way -- I'll try this out. Thank you!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/Loader.java
icbaker: After some discussion on the internal review, we're leaning towards putting some sort of parameter back in here to (as you suggested) make it clear that the meaning of 'start' here is distinct from the meaning in `startLoading` (i.e. this will be called once per retry).

So I'm going to put back your `errorCount` parameter, but rename it `retryCount`.

We're not planning to include the `error` parameter, because as agreed that's available from the `onLoadError` method.

PR #1607, opened 2024-08-09
Make Extractor/Mp4 Atom Parsing code reusable by external custom mp4 …

colinkho:
…extractors

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/AtomParsers.java
@@ -62,7 +62,7 @@ import java.util.Objects;
 
 /** Utility methods for parsing MP4 format atom payloads according to ISO/IEC 14496-12. */
 @SuppressWarnings("ConstantField")
-/* package */ final class AtomParsers {
+public final class AtomParsers {
 
   private static final String TAG = "AtomParsers";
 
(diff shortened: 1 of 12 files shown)


icbaker:
Thanks for sending this. I'm going to submit some of this myself directly internally:
1. Move a couple of parsing methods from `Atom` to `AtomParsers`
2. Move `Atom` to `lib-container` and make it public (so it can in future be used by `muxer`)

Once these changes are submitted I'll let you know and you can rebase this change so it only has the `AtomParsers`, `Sniffer` and `TrackSampleTable` changes.

icbaker:
The changes have been submitted:
* https://github.com/androidx/media/commit/9dfd72b6c62a56c569fad7af88fad61a549d2e17
* https://github.com/androidx/media/commit/9d008da356bd6e26f4f18a9dba48ce9741caad6b
* https://github.com/androidx/media/commit/ee27334f06f7c74618049e9ae99b7f002918aeee

The final change renamed `Atom` to `Mp4Box` for two reasons:
1. The 'box' terminology more closely matches the ISO 14496-12 spec.
2. Other MP4-specific files in `androidx.media3.container` are prefixed with `Mp4` instead of being in an `mp4` sub-package.

I suggest we rename `AtomParsers` to `BoxParsers` and keep it in the `extractor.mp4` package. We don't need to also touch every call-site, because it makes the blast radius of the change too large.

I don't mind if you do that in this PR, or we can do it directly as an internal change.

colinkho:
@icbaker Rebased to capture your commits and renamed to BoxParsers. Thank you!

icbaker:

PR #1613, opened 2024-08-13
IndexOutOfBoundException at HlsChunkSource.getNextSegmentHolder #9479

VijayakumaraRevanna:
https://github.com/google/ExoPlayer/issues/9479


icbaker:
https://github.com/google/ExoPlayer/issues/9479 was closed as 'bad media'. I'm afraid we are unlikely to consider a PR that changes the library behaviour to support invalid media.

Please open a new issue on this tracker, with details of the issue you're seeing and an explanation of why the media involved is valid - then we can revisit this PR.

PR #1616, opened 2024-08-13
Differ AudioManager retrieval to whenever AudioFocusManagement is req…

colinkho:
…uired

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/AudioFocusManager.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/AudioFocusManager.java
@@ -162,10 +162,11 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
   private static final float VOLUME_MULTIPLIER_DUCK = 0.2f;
   private static final float VOLUME_MULTIPLIER_DEFAULT = 1.0f;
 
-  private final AudioFocusListener focusListener;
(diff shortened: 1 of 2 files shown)


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/AudioFocusManager.java
```diff
@@ -455,6 +453,13 @@ private void executePlayerCommand(@PlayerCommand int playerCommand) {
     }
   }
 
+  private AudioManager getAudioManager() {
+    if (audioManager == null) {
```
icbaker, line 457: NIt: you can use Guava's `Suppliers.memoize` for this, to avoid manually needing to check for null, so change `private final AudioManager audioManager;` above to `private final Supplier<AudioManager> audioManager;`, change the instantiation to:
```
this.audioManager =
        Suppliers.memoize(() -> checkNotNull(
            (AudioManager) context.getApplicationContext().getSystemService(Context.AUDIO_SERVICE)));
```

And then change the usage sites from `audioManager` to `audioManager.get()`

Maybe change the field name to `audioManagerSupplier` too - up to you.

colinkho: Good suggestion. Will change accordingly

icbaker:

PR #1618, opened 2024-08-13
Add a static rendererSupportsFormat method for MCVR

khouzam:
This refactors the `supportsFormat` method on the MediaCodecVideoRenderer to be a static method that can be called from external components to determine if a video format is supported by the device renderers. Since `context` is the only component that is part of the MCVR, it can easily be obtained externally and passed to the method.

This can help optimize format support decisions ahead of time of a player having been instantiated, such as removing unsupported representations from a dash manifest.

This could also be done for the MCAR but would also require passing in the AudioSink and isn't required at this time.


colinkho:
Hey @microkatz, possible you could take a look at this? Ty!

microkatz:

PR #1624, opened 2024-08-15
Add getter for ChunkIndex on InitializationChunk

colinkho:


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/chunk/InitializationChunk.java
```diff
@@ -109,4 +110,9 @@ public void load() throws IOException {
       DataSourceUtil.closeQuietly(dataSource);
     }
   }
+
+  @Nullable
+  public ChunkIndex getChunkIndex() {
+    return chunkExtractor.getChunkIndex();
```
icbaker, line 116: @tonihei Is it correct to call `chunkExtractor.getChunkIndex()` here, vs do it above on L107 just after we've finished extracting **only** this initialization chunk, and then cache it in a field?

It seems that the code as-is could theoretically return a `ChunkIndex` derived from chunks later than this `InitializationChunk`, which seems like it could be a bit confusing.

tonihei: Sounds like the safer and less error-prone option, yes. 

While you are still at it - do you mind adding Javadoc to the new public method? That will allow you to clearly state that this is the chunk index loaded from the initialization chunk.

icbaker:

PR #1638, opened 2024-08-20
Do not flush eac3(joc) decoder on reuse

mzchan-dolby:
eac3(joc) decoders do not need to be flushed to be reused for the next compatible track. This change allows for gapless playback on devices with Dolby decoders that require internal re-initialization on flush.


mzchan1:
Please abandon this PR as it was pushed via a non CLA account. Thanks

Issue #1632, opened 2024-08-20
Speaker info for WebVTT subtitles


ashiagr:
We're using `WebvttParser()` to parse WebVTT transcripts, which internally uses `WebvttCueParser()` to parse cues.
While `WebvttCueParser.parseCueText(...)` provides the payload with style formatting, it removes the speaker identifiers or the `TAG_VOICE`.

Can we get an in-built support for speaker info within the cues of WebVTT transcripts?

icbaker:
I guess you're referring to this? https://www.w3.org/TR/webvtt1/#webvtt-cue-voice-span

We do have some code already that parses `<v[voice="foo"]>` tags in order to resolve the associated styles: https://github.com/search?q=repo%3Aandroidx%2Fmedia+path%3Avtt+voice&type=code

But the voice information isn't directly exposed in the resulting `Cue` object.

I suspect this would need to be exposed using a ['custom span'](https://developer.android.com/develop/ui/views/text-and-emoji/spans#custom-spans) in `Cue.text`, like we do for Japanese rubies: https://github.com/androidx/media/blob/release/libraries/common/src/main/java/androidx/media3/common/text/RubySpan.java

I'll mark this as an enhancement. I'm afraid we are unlikely to work on this ourselves soon, but we would consider a high quality PR implementing this.

icbaker:
Closing this because https://github.com/androidx/media/pull/1652 has been merged - thanks!

PR #1652, opened 2024-08-23
Add VTT voice spans to cues

MiSikora:
This adds voice spans available in WebVTT cues. Spans were already parsed but they were never exposed and attached to the output `Spannable` created by `WebvttCueParser`. For more info see: #1632

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/text/VoiceSpan.java
+++ b/libraries/common/src/main/java/androidx/media3/common/text/VoiceSpan.java
@@ -31,7 +31,7 @@ import java.util.Set;
  *   voice spans</a>.
  */
 @UnstableApi
-public final class VoiceSpan implements LanguageFeatureSpan {
+public final class VoiceSpan {
 
   /** The speaker name. */
   public final String speakerName;
--- a/libraries/common/src/main/java/androidx/media3/common/text/VoiceSpan.java
+++ b/libraries/common/src/main/java/androidx/media3/common/text/VoiceSpan.java
@@ -23,10 +23,10 @@ import androidx.media3.common.util.UnstableApi;
 import androidx.media3.common.util.Util;
 
 /**
- * A span representing a speaker.
+ * A span representing the speaker of the spanned text.
  *
- * <p>More information on <a href="https://www.w3.org/TR/webvtt1/#webvtt-cue-voice-span">
- *   voice spans</a>.
+ * <p>For example a <a href="https://www.w3.org/TR/webvtt1/#webvtt-cue-voice-span">WebVTT voice
+ * span</a>.
  */
 @UnstableApi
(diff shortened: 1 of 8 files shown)


libraries/common/src/main/java/androidx/media3/common/text/VoiceSpan.java
```diff
@@ -0,0 +1,62 @@
...
+ *   voice spans</a>.
+ */
+@UnstableApi
+public final class VoiceSpan {
+
+  /** The speaker name. */
+  public final String speakerName;
+
+  /** The classes associated with the text. It can specify things like "first", "loud", etc. */
+  public final Set<String> classes;
```
icbaker, line 40: aiui these aren't really useful without access to the rest of the styling info. Do you have a use-case/need for these values?

libraries/common/src/main/java/androidx/media3/common/text/VoiceSpan.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * A span representing a speaker.
+ *
+ * <p>More information on <a href="https://www.w3.org/TR/webvtt1/#webvtt-cue-voice-span">
+ *   voice spans</a>.
+ */
+@UnstableApi
+public final class VoiceSpan {
+
+  /** The speaker name. */
```
icbaker, line 36: the webvtt spec uses the phrase "voice name", so I think we can use that here:

```
/** The voice name. */
public final String name;
```

icbaker (changes requested):
Thanks for the contribution, and thorough test changes too!

libraries/common/src/main/java/androidx/media3/common/text/VoiceSpan.java
MiSikora: No I don't have a user for those. Do you want me to remove it?

icbaker: Yeah, let's remove it for now, if a use-case comes up later we can re-add it.

MiSikora: [2a4a29b](https://github.com/androidx/media/pull/1652/commits/2a4a29b47c627cbb9d04d0845c3242456a643726)

libraries/common/src/main/java/androidx/media3/common/text/VoiceSpan.java
MiSikora: [340c14a](https://github.com/androidx/media/pull/1652/commits/340c14ad425062df753d1189114886130286d6e3)

icbaker:

icbaker (approved):

PR #1653, opened 2024-08-24
Add a way to set PlayerControlView fullscreen button state

### LINKED ISSUE #1590: Enable the developer to set the default state of isFullscreen
### Use case description
When multiple video players views are chained, if the user presses the fullscreen button in the ``PlayerControlView`` (the state of the button would go from ``fullScreenButton`` to ``minimalFullScreenButton``) and then navigates to a second ``PlayerControlView`` while still being in fullscreen, the button will be in its default position (``fullScreenButton``) in this new ``PlayerControlView``.

### Proposed solution
Create a public facing method in ``PlayerView`` called ``setIsFullscreen`` which would call ``PlayerControlView.updateFullScreenButtonForState`` with the new state (also add the corresponding ``fullScreenButton`` and would update ``isFullscreen``)

theskyblockman:
This fixes #1590.

I added and documented the new public-facing methods, moved some code to a new private method and used it to not duplicate code.

### DIFF
--- a/libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
+++ b/libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
@@ -1575,26 +1575,31 @@ public class PlayerControlView extends FrameLayout {
   }
 
   private void onFullScreenButtonClicked(View v) {
-    updateIsFullScreen(!isFullScreen);
+    updateIsFullscreen(!isFullScreen);
   }
 
   /**
-   * Updates whether the controller is in full screen to its icon
-   * and reports it to to the listener.
-   * @param isFullScreen If the view is in full screen.
+   * Updates whether the controller is in fullscreen, changing its fullscreen icon and reports it
+   * to to the listener.
+   *
+   * <p>For {@code isFullscreen} equals {@code true} the icon will be set
+   * to {@link R.drawable#exo_styled_controls_fullscreen_exit} or else
+   * {@link R.drawable#exo_styled_controls_fullscreen_enter}.
+   *
+   * @param isFullscreen If the view is in full screen.
    */
-  public void updateIsFullScreen(boolean isFullScreen) {
-    boolean fullScreenModeChanged = this.isFullScreen != isFullScreen;
+  public void updateIsFullscreen(boolean isFullscreen) {
+    boolean fullscreenModeChanged = this.isFullScreen != isFullscreen;
 
-    if(fullScreenModeChanged && onFullScreenModeChangedListener == null) {
+    if (fullscreenModeChanged && onFullScreenModeChangedListener == null) {
       return;
     }
 
-    this.isFullScreen = isFullScreen;
-    updateFullScreenButtonForState(fullScreenButton, isFullScreen);
-    updateFullScreenButtonForState(minimalFullScreenButton, isFullScreen);
-    if(fullScreenModeChanged && onFullScreenModeChangedListener != null) {
-      onFullScreenModeChangedListener.onFullScreenModeChanged(isFullScreen);
+    this.isFullScreen = isFullscreen;
+    updateFullScreenButtonForState(fullScreenButton, isFullscreen);
+    updateFullScreenButtonForState(minimalFullScreenButton, isFullscreen);
+    if (fullscreenModeChanged && onFullScreenModeChangedListener != null) {
+      onFullScreenModeChangedListener.onFullScreenModeChanged(isFullscreen);
     }
   }
 
--- a/libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
+++ b/libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
@@ -1575,14 +1575,25 @@ public class PlayerControlView extends FrameLayout {
   }
 
   private void onFullScreenButtonClicked(View v) {
-    if (onFullScreenModeChangedListener == null) {
+    updateIsFullScreen(!isFullScreen);
+  }
+
+  /**
+   * Updates whether the controller is in full screen to its icon
+   * and reports it to to the listener.
+   * @param isFullScreen If the view is in full screen.
+   */
+  public void updateIsFullScreen(boolean isFullScreen) {
+    boolean fullScreenModeChanged = this.isFullScreen != isFullScreen;
+
+    if(fullScreenModeChanged && onFullScreenModeChangedListener == null) {
       return;
     }
 
-    isFullScreen = !isFullScreen;
+    this.isFullScreen = isFullScreen;
     updateFullScreenButtonForState(fullScreenButton, isFullScreen);
     updateFullScreenButtonForState(minimalFullScreenButton, isFullScreen);
-    if (onFullScreenModeChangedListener != null) {
+    if(fullScreenModeChanged && onFullScreenModeChangedListener != null) {
       onFullScreenModeChangedListener.onFullScreenModeChanged(isFullScreen);
     }
   }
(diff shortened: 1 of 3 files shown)


libraries/ui/src/main/java/androidx/media3/ui/PlayerView.java
```diff
@@ -1138,6 +1138,16 @@ public void setFullscreenButtonClickListener(@Nullable FullscreenButtonClickList
     controller.setOnFullScreenModeChangedListener(componentListener);
   }
 
+  /**
+   * Sets whether the fullscreen button is fullScreenButton or minimalFullScreenButton.
```
oceanjules, line 1142: Please see the difference between `fullScreenButton` and `minimalFullScreenButton`. They are not the same button. Both buttons act as fullscreen toggles and hence can have exit/enter fullscreen icons

Screenshot 2024-09-04 at 00 01 56

libraries/ui/src/main/java/androidx/media3/ui/PlayerView.java
```diff
@@ -1138,6 +1138,16 @@ public void setFullscreenButtonClickListener(@Nullable FullscreenButtonClickList
...
+   *
+   * @param isFullscreen Whether the player is currently in fullscreen.
```
oceanjules, line 1144: Fullscreen state is semantically inverse to the button display.

* When the player is fullscreen -> button should show "exit fullscreen"
* When the player is minimized -> button should show "enter fullscreen"

It might be best to document this method in terms of the icon drawable you will be setting (since you can't really claim to be imperatively setting the player to a fullscreen activity)

* exo_styled_controls_fullscreen_enter
* exo_styled_controls_fullscreen_exit

libraries/ui/src/main/java/androidx/media3/ui/PlayerView.java
```diff
@@ -1138,6 +1138,16 @@ public void setFullscreenButtonClickListener(@Nullable FullscreenButtonClickList
...
+   */
+  public void setFullscreenButtonState(boolean isFullscreen) {
+    Assertions.checkStateNotNull(controller);
+    controller.updateIsFullScreen(isFullscreen);
```
oceanjules, line 1148: Just a stylistic note - you seem to be mixing `Fullscreen` and `FullScreen`. I understand that a mixture of spellings are currently used throughout the repo, but let's align on not capitalizing "S", since "fullscreen" is one word

libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
```diff
@@ -1574,14 +1574,25 @@ private void seekToTimeBarPosition(Player player, long positionMs) {
...
+
+  /**
+   * Updates whether the controller is in full screen to its icon
+   * and reports it to to the listener.
+   * @param isFullScreen If the view is in full screen.
+   */
+  public void updateIsFullScreen(boolean isFullScreen) {
+    boolean fullScreenModeChanged = this.isFullScreen != isFullScreen;
+
+    if(fullScreenModeChanged && onFullScreenModeChangedListener == null) {
```
oceanjules, line 1588: I think this line can be simplified to `if(onFullScreenModeChangedListener == null)`.

No matter whether the mode changed or not, if the listener is not there, we don't do anything

libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
```diff
@@ -1574,14 +1574,25 @@ private void seekToTimeBarPosition(Player player, long positionMs) {
...
 
   private void onFullScreenButtonClicked(View v) {
-    if (onFullScreenModeChangedListener == null) {
+    updateIsFullScreen(!isFullScreen);
+  }
+
+  /**
+   * Updates whether the controller is in full screen to its icon
+   * and reports it to to the listener.
+   * @param isFullScreen If the view is in full screen.
```
oceanjules, line 1583: I would reword this javadoc (see PlayerView comment)

oceanjules (commented):
Thanks for a PR!

libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
icbaker: Don't we want to update the internal state whether or not we have a listener attached?

Consider:

```java
playerView.setFullscreenButtonState(true);
playerView.setFullscreenButtonClickListener(isFullScreen -> ...);
```

Registering the listener means the button is shown, but the "enter full screen" button will be shown even though we've been told the content is full screen above, and so the "exit full screen" **should** be shown.

libraries/ui/src/main/java/androidx/media3/ui/PlayerView.java
theskyblockman: I rephrased in 2f1b08fa the javadoc:
Sets whether the player is currently in fullscreen, this will change the displayed icon.

If isFullscreen is true, ``exo_styled_controls_fullscreen_exit`` will be displayed or else ``exo_styled_controls_fullscreen_enter``.

libraries/ui/src/main/java/androidx/media3/ui/PlayerView.java
theskyblockman: I have rephrased this javadoc too in 2f1b08fa:
Updates whether the controller is in fullscreen, changing its fullscreen icon and reports it to to the listener.

For isFullscreen equals true the icon will be set to ``exo_styled_controls_fullscreen_exit`` or else ``exo_styled_controls_fullscreen_enter``.

libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
theskyblockman: I updated this javadoc too:
Updates whether the controller is in fullscreen, changing its fullscreen icon and reports it to to the listener.

For isFullscreen equals true the icon will be set to ``exo_styled_controls_fullscreen_exit`` or else ``exo_styled_controls_fullscreen_enter``.

libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
oceanjules: Sorry, yes, you are right - got confused with the previous impl of:

```
  private void onFullScreenButtonClicked(View v) {
    if (onFullScreenModeChangedListener == null) {  <-- can't be true, the button is not visible if the listener==null
      return;
    }
   ...
```

So I guess this line could be changed to `if (!fullScreenModeChanged)` to exit early?

libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java
```diff
@@ -1574,26 +1574,31 @@ private void seekToTimeBarPosition(Player player, long positionMs) {
...
+   *
+   * @param isFullscreen If the view is in full screen.
    */
-  public void updateIsFullScreen(boolean isFullScreen) {
-    boolean fullScreenModeChanged = this.isFullScreen != isFullScreen;
+  public void updateIsFullscreen(boolean isFullscreen) {
+    boolean fullscreenModeChanged = this.isFullScreen != isFullscreen;
 
-    if(fullScreenModeChanged && onFullScreenModeChangedListener == null) {
+    if (fullscreenModeChanged && onFullScreenModeChangedListener == null) {
```
oceanjules, line 1593: @theskyblockman, please see mine and icbaker's convo above:

tldr; 
* `if (!fullScreenModeChanged)` to exit quickly
* whether or not we have the listener is not important at this point
* If the mode DID change, we still want to update the state and go through the rest of this function

theskyblockman: Changed in 616e17af.

oceanjules:

theskyblockman:
Thank you!

PR #1665, opened 2024-08-28
Remove obsolete semicolon

JaroslavHerber:


tonihei:
We can't merge PRs into the release branch unfortunately, so you'd need to rebase it against the main branch. Alternatively we can just remove the semicolon ourselves, but also happy to merge the PR if you would like the credit :)

PR #1678, opened 2024-08-31
Create android.yml

myHerbDev:


marcbaechinger:
Can you please add a description what this yaml file is for? I don't think we will add that if theres isn't a good reason for this regarding improving building with gradle or working in Android Studio.

You'd also to sign the CLA after that justifaction went trough. Thank you for your understanding.

myHerbDev:
> Can you please add a description what this yaml file is for? I don't think we will add that if theres isn't a good reason for this regarding improving building with gradle or working in Android Studio.
> 
> You'd also to sign the CLA after that justifaction went trough. Thank you for your understanding.

It was an experiment, can be deleted.

PR #1736, opened 2024-09-18
Add method (dispatchEvent) to MediaSourceEventListener.EventDispatcher to allow invoking subclassed listeners

colinkho:


tianyif:
Hi @colinkho, I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review. Thanks!

PR #1740, opened 2024-09-19
Fix input type in `shortform` demo

MGaetan89:
Currently, in the `shortform` demo, the `How Many Players?` input has the `numberDecimal` type. When parsing the value in `MainActivity.afterTextChanged()`, the app may crash if the provided value is a decimal.
This PR changes the input type to `number`, so it is no longer possible to provide a decimal value.


oceanjules:
Hi @MGaetan89,
Thank you for the catch! I didn't realise `number` would also bring up the keyboard, but with `,` and `.` displayed and disabled.

PR #1752, opened 2024-09-24
Allows exact live sync for HLS

stevemayhew:
This pull request satisfies the enhancement request #702.   For playback longer than a few hours the fix for SntpClient is also required, see pull request #697

### DIFF
--- a/demos/main/src/main/java/androidx/media3/demo/main/IntentUtil.java
+++ b/demos/main/src/main/java/androidx/media3/demo/main/IntentUtil.java
@@ -26,6 +26,7 @@ import androidx.annotation.OptIn;
 import androidx.media3.common.C;
 import androidx.media3.common.MediaItem;
 import androidx.media3.common.MediaItem.ClippingConfiguration;
+import androidx.media3.common.MediaItem.LiveConfiguration;
 import androidx.media3.common.MediaItem.SubtitleConfiguration;
 import androidx.media3.common.MediaMetadata;
 import androidx.media3.common.Player;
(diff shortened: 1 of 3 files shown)


tonihei:
When we initially implemented support for low-latency HLS, this clock sync wasn't technically needed because all available segments and parts are explicitly published in the HLS playlist without the need for an explicit clock source (unlike DASH where this part of the core logic). The current live offset is always simply measured against the current device time and the "program data time" tags in the playlist. And the target offset is adjusted according to the HLS playlist settings, making it independent of the current system time (=no clock sync needed). 

Having said that, it also means the streams don't use the exact same target offset across devices. And if you manually set a target offset, it will be fully dependent on the system time again (this is what I think you are doing in your use case?)

More concretely:

- I think we can definitely add the offset from the `SntpClient` in `HlsMediaSource` as suggested
- The `IntentUtil `changes should probably be removed from the PR as they don't add a lot of value (we have lots of settings that can't be changed from a command line, that's fine for a demo app)
- The changes in `DefaultHlsPlaylistTracker` are more interesting.
   - You could argue that we don't need them at all. If an app cares about this exact sync, they can trigger the `SntpClient` sync manually before playback.
   - Why is the update triggered before every playlist load? I assume it's sufficient to only sync in the `start` method?
   - One problem of this change is that it postpones the initial playlist load for all HLS streams until the `SntpClient` reports back. What do you think about starting the time sync in the `start` method as already done. But instead of only triggering the playlist load after the time sync, do it immediately (in parallel). Then we'd need an additional check in `processLoadedPlaylist` to only process it if the time sync is already available.

PR #1754, opened 2024-09-24
Incorporate injection of custom Executor to the eventual Loader

colinkho:
This is enabled for ChunkSampleStream and ProgressiveMediaSource


colinkho:
Hey @tonihei, one area that I might have missed here is being able to listen on the Loader's onLoaderReleased callback for ProgressiveMediaSource. There isn't a way for us to know when to shutdown. Do you have any suggestions or if this should be added as a parameter?

tonihei:
> There isn't a way for us to know when to shutdown without it. Ideally, we should be able to pass in our own downloadExecutorReleaser if possible.

Thanks for highlighting this again! This is actually also what we discussed before and it just came up during the internal review again too. I will add another parameter for onDownloadExecutorReleased or similar during the import.

PR #1768, opened 2024-09-30
Incorporate MediaSourceEventListener.onLoadStarted with retryCount pa…

colinkho:
…rameter with backward compatibility

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollector.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollector.java
@@ -411,6 +411,20 @@ public class DefaultAnalyticsCollector implements AnalyticsCollector {
   }
 
   @Override
+  public final void onLoadStarted(
+      int windowIndex,
+      @Nullable MediaPeriodId mediaPeriodId,
+      LoadEventInfo loadEventInfo,
+      MediaLoadData mediaLoadData,
+      int retryCount) {
+    EventTime eventTime = generateMediaPeriodEventTime(windowIndex, mediaPeriodId);
+    sendEvent(
+        eventTime,
+        AnalyticsListener.EVENT_LOAD_STARTED,
+        listener -> listener.onLoadStarted(eventTime, loadEventInfo, mediaLoadData, retryCount));
+  }
+
+  @Override
   public final void onLoadCompleted(
       int windowIndex,
       @Nullable MediaPeriodId mediaPeriodId,
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollector.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollector.java
@@ -409,7 +409,7 @@ public class DefaultAnalyticsCollector implements AnalyticsCollector {
     EventTime eventTime = generateMediaPeriodEventTime(windowIndex, mediaPeriodId);
     sendEvent(
         eventTime,
-        C.INDEX_UNSET,
+        /* eventFlag= */ C.INDEX_UNSET,
         listener -> listener.onLoadStarted(eventTime, loadEventInfo, mediaLoadData));
     sendEvent(
         eventTime,
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollector.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollector.java
@@ -420,7 +420,7 @@ public class DefaultAnalyticsCollector implements AnalyticsCollector {
     EventTime eventTime = generateMediaPeriodEventTime(windowIndex, mediaPeriodId);
     sendEvent(
         eventTime,
-        AnalyticsListener.EVENT_LOAD_STARTED,
+        AnalyticsListener.EVENT_ON_LOAD_STARTED,
         listener -> listener.onLoadStarted(eventTime, loadEventInfo, mediaLoadData, retryCount));
   }
 
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollector.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollector.java
@@ -409,12 +409,11 @@ public class DefaultAnalyticsCollector implements AnalyticsCollector {
     EventTime eventTime = generateMediaPeriodEventTime(windowIndex, mediaPeriodId);
     sendEvent(
         eventTime,
-        /* eventFlag= */ C.INDEX_UNSET,
-        listener -> listener.onLoadStarted(eventTime, loadEventInfo, mediaLoadData));
-    sendEvent(
-        eventTime,
         AnalyticsListener.EVENT_LOAD_STARTED,
-        listener -> listener.onLoadStarted(eventTime, loadEventInfo, mediaLoadData, retryCount));
+        listener -> {
+          listener.onLoadStarted(eventTime, loadEventInfo, mediaLoadData);
+          listener.onLoadStarted(eventTime, loadEventInfo, mediaLoadData, retryCount);
+        });
   }
 
   @Override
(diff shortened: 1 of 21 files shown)


icbaker:
I think in its current form this stops calling any existing implementations of `AnalyticsListener.onLoadStarted(EventTime, LoadEventInfo, MediaLoadData)` (the one that's now deprecated). I checked this by adding some logging locally to this override in `EventLogger` and I see the logging before this change, but not after it.

We need to keep these implementations working - because although they're marked as deprecated, deprecated things are still expected to work.

Ensuring both the old and new callbacks continue working, without accidentally multiplying the calls, is somewhat tricky (you mentioned you experimented with putting some duplication in the `EventDispatcher` implementation, but this results in re-duplicating the events when they are forwarded inside e.g. `CompositeMediaSource`).

We can put the duplication in the callsites instead (i.e. in each `MediaSource` implementation). One question then becomes: how do we ensure that any custom `MediaSource` implementations call the new method (since we can't change them)? You could argue that this doesn't matter, because an app with a custom `MediaSource` impl also owns their custom `AnalyticsListener` impl, so they can migrate either both or neither - and therefore remain compatible with themselves.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ProgressiveMediaPeriod.java
```diff
@@ -869,7 +896,8 @@ private void startLoading() {
         C.SELECTION_REASON_UNKNOWN,
         /* trackSelectionData= */ null,
         /* mediaStartTimeUs= */ loadable.seekTimeUs,
-        durationUs);
+        durationUs,
+        0);
```
icbaker, line 900: please add a parameter comment here and elsewhere you pass a literal

libraries/exoplayer/src/test/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollectorTest.java
```diff
@@ -2199,6 +2200,15 @@ public void onLoadStarted(
       reportedEvents.add(new ReportedEvent(EVENT_LOAD_STARTED, eventTime));
```
icbaker, line 2200: you might need to use a different event type here, so we can test that both the new method **and** the old method are called

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/ServerSideAdInsertionMediaSource.java
```diff
@@ -427,20 +427,22 @@ public void onLoadStarted(
...
       LoadEventInfo loadEventInfo,
-      MediaLoadData mediaLoadData) {
+      MediaLoadData mediaLoadData,
+      int retryCount) {
     @Nullable
     MediaPeriodImpl mediaPeriod =
         getMediaPeriodForEvent(mediaPeriodId, mediaLoadData, /* useLoadingPeriod= */ true);
     if (mediaPeriod == null) {
-      mediaSourceEventDispatcherWithoutId.loadStarted(loadEventInfo, mediaLoadData);
+      mediaSourceEventDispatcherWithoutId.loadStarted(loadEventInfo, mediaLoadData, retryCount);
```
icbaker, line 436: this call (and the one below in this method) are basically forwarding from an inner `MediaSource` instance. SImilar to `CompositeMediaSource.ForwardingEventListener` we should implement both the old and new methods here, and forward each one to the respective `EventDispatcher` method.

Otherwise a custom `MediaSource` that hasn't yet switched to dispatching the new method will have all its `onLoadStarted` events dropped at this point.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/MediaSourceEventListener.java
```diff
@@ -47,12 +47,37 @@ public interface MediaSourceEventListener {
...
+   *     is zero for the first load attempt).
+   */
+  default void onLoadStarted(
+      int windowIndex,
+      @Nullable MediaPeriodId mediaPeriodId,
+      LoadEventInfo loadEventInfo,
+      MediaLoadData mediaLoadData,
+      int retryCount) {
+    if (retryCount == 0) {
+      onLoadStarted(windowIndex, mediaPeriodId, loadEventInfo, mediaLoadData);
```
icbaker, line 77: I think we should remove this attempted magic redirecting and keep it all in the `MediaSource` implementations.

This doesn't work when the call is redirected from `MediaSourceEventListener` to `AnalyticsListener` via `AnalyticsCollector` which overrides both the new and old `MediaSourceEventListener` methods (hence disabling this magic). `AnalyticsCollector` **must** override both methods for compat with custom `MediaSource` impls that aren't yet dispatching the new method.

icbaker (changes requested):

colinkho:
> I think in its current form this stops calling any existing implementations of `AnalyticsListener.onLoadStarted(EventTime, LoadEventInfo, MediaLoadData)` (the one that's now deprecated). I checked this by adding some logging locally to this override in `EventLogger` and I see the logging before this change, but not after it.
> 
> We need to keep these implementations working - because although they're marked as deprecated, deprecated things are still expected to work.
> 
> Ensuring both the old and new callbacks continue working, without accidentally multiplying the calls, is somewhat tricky (you mentioned you experimented with putting some duplication in the `EventDispatcher` implementation, but this results in re-duplicating the events when they are forwarded inside e.g. `CompositeMediaSource`).
> 
> We can put the duplication in the callsites instead (i.e. in each `MediaSource` implementation). One question then becomes: how do we ensure that any custom `MediaSource` implementations call the new method (since we can't change them)? You could argue that this doesn't matter, because an app with a custom `MediaSource` impl also owns their custom `AnalyticsListener` impl, so they can migrate either both or neither - and therefore remain compatible with themselves.

Thanks for the feedback and makes sense. I think duplicating the calls at each callsite makes sense to me since this is treating the updated API as a new API entirely. This seems cleaner too. I will adjust this

libraries/exoplayer/src/test/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollectorTest.java
colinkho: Any naming convention to suggest here? I was thinking `EVENT_ON_LOAD_STARTED` but not sure if that is confusing and outside of Google's naming conventions

icbaker: What about keeping `EVENT_LOAD_STARTED` for the new un-deprecated one and renaming the old one to `EVENT_LOAD_STARTED_WITHOUT_RETRY_COUNT`? It's a bit long, but it only needs to live as long as we keep the deprecated method.

icbaker: Ah because i was reading this in a test, i thought this constant was defined in the test - my bad, now i see it's defined in `AnalyticsListener` interface.

In that case i retract my comment, we shouldn't duplicate this constant imo - we should find a different way to test that both the new and old methods are called.

libraries/exoplayer/src/test/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollectorTest.java
```diff
@@ -2199,6 +2272,15 @@ public void onLoadStarted(
...
     }
 
+    @Override
+    public void onLoadStarted(
+        EventTime eventTime,
+        LoadEventInfo loadEventInfo,
+        MediaLoadData mediaLoadData,
+        int retryCount) {
+      reportedEvents.add(new ReportedEvent(DEPRECATED_EVENT_LOAD_STARTED, eventTime));
```
icbaker, line 2281: I think you've got this the wrong way round? This is an override of the new, un-deprecated method with the `retryCount` parameter but it's using the `DEPRECATED_` constant.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/DefaultAnalyticsCollector.java
```diff
@@ -406,10 +406,24 @@ public final void onLoadStarted(
     EventTime eventTime = generateMediaPeriodEventTime(windowIndex, mediaPeriodId);
     sendEvent(
         eventTime,
-        AnalyticsListener.EVENT_LOAD_STARTED,
+        C.INDEX_UNSET,
```
icbaker, line 409: I know we discussed this, and recommended this approach, but thinking about this again, this causes a problem if:
1. An app has a custom `MediaSource` which won't be updated to trigger the new overload
2. The same app registers an `AnalyticsListener` implementation and waits for `onEvents` to be called with `EVENT_LOAD_STARTED`

Because this implementation will no longer be triggered after this CL.

Discussing with @tonihei we're leaning towards making this a breaking change on the `MediaSourceEventListener` interface to avoid all these pitfalls of trying to keep everything working without breaking compilation. I think we should keep both the old and new methods on `AnalyticsListener`, as that is likely the part that most apps are overriding (and its easy to keep the compat behaviour at that level).

I'm going to play around with that approach and update this PR (I'll also fix the test thing i commented on) - so you don't need to take any action right now.

icbaker (approved):

PR #1792, opened 2024-10-09
Bugfix: segment_duration and media_time use different units

ybai001:
According to ISO 14496-12, section 8.6.6.3,

> segment_duration is an integer that specifies the duration of this edit segment in units of the
timescale in the Movie Header Box
media_time is an integer containing the starting time within the media of this edit segment (in
media time scale units, in composition time). If this field is set to –1, it is an empty edit. The last
edit in a track shall never be an empty edit. Any difference between the duration in the Movie
Header Box, and the track’s duration is expressed as an implicit empty edit at the end.

segment_duration is based on timescale in movie header box(mdhd)
media_time is based on timescale track header/media timescale

So segment_duration and media_time use different units.

The implementation in other Media3 codes/AOSP are correct, such as
(1) BoxParser.java line 712, 714, 716--718 and 731: https://github.com/androidx/media/blob/main/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java#L712
(2) On AOSP https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/module/extractors/mp4/MPEG4Extractor.cpp;l=583


rohitjoins:
I'm going to send this for internal review now. Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks!

PR #1794, opened 2024-10-09
SntpClient periodically re-syncs the offset to NTP server

stevemayhew:
This is the equivalent pull request to https://github.com/androidx/media/pull/697, just rebased (with simple conflict resolve) to the main branch.

@marcbaechinger feel free to pick one and close the other.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/util/SntpClient.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/util/SntpClient.java
@@ -49,9 +49,6 @@ public final class SntpClient {
   /** The default maximum time, in milliseconds, to wait for the SNTP request to complete. */
   public static final int DEFAULT_TIMEOUT_MS = 1_000;
 
-  /** Max time to allow before re-initializing with NTP server, default is 10 minutes */
-  private static final long MAX_ELAPSED_MS_TILL_UPDATE = 60 * 10 * 1_000L;
-
   /** Callback for calls to {@link #initialize(Loader, InitializationCallback)}. */
   public interface InitializationCallback {
 
@@ -100,6 +97,9 @@ public final class SntpClient {
   private static int timeoutMs = DEFAULT_TIMEOUT_MS;
 
   @GuardedBy("valueLock")
+  private static long maxElapsedTimeUntilUpdateMs = C.TIME_UNSET;
+
+  @GuardedBy("valueLock")
   private static long lastUpdateElapsedRealtime = C.TIME_UNSET;
 
   private SntpClient() {}
@@ -126,7 +126,6 @@ public final class SntpClient {
       if (!SntpClient.ntpHost.equals(ntpHost)) {
         SntpClient.ntpHost = ntpHost;
         isInitialized = false;
-        lastUpdateElapsedRealtime = C.TIME_UNSET;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


marcbaechinger:
This looks good to me as in how this would work regarding reinitialization. Many thanks, Steve.

I was wondering whether we want to use `C.TIME_UNSET` to never reinitialize as a default. Mainly to not change the behaviour of `DashMediaSource` without opting in. Then we could do such a getter similar to those for timeout and sntp host that allow opt-in in.

We would then wouldn't need to discuss how you came to choose 10 minutes as a default. :)

You as a user could then opt-in once at app level using the `SntpClient` API? I'm not really convinced. However, the static design of that client makes it a bit difficult to provide an API that could be enabled on  a media source level or similar. `SntpClient` being a proper object instance would probably be the right thing to do. 

How does adding a getter to override the default sound to you?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/util/SntpClient.java
```diff
@@ -49,6 +49,9 @@ public final class SntpClient {
   /** The default maximum time, in milliseconds, to wait for the SNTP request to complete. */
   public static final int DEFAULT_TIMEOUT_MS = 1_000;
 
+  /** Max time to allow before re-initializing with NTP server, default is 10 minutes */
+  private static final long MAX_ELAPSED_MS_TILL_UPDATE = 60 * 10 * 1_000L;
```
marcbaechinger, line 53: UNTIL ? :)

marcbaechinger: As mentioned in the PR comment. Is this a value that comes from your experience when operating live streams? I'm mostly curious...

marcbaechinger (commented):
As mentioned in the main comment. Consider adding a getter to override the default of C.TIME_UNSET and disable reinitialization for C.TIME_UNSET?

nikitapn:
This is what we need also, but could you remove hardcoded _MAX_ELAPSED_MS_TILL_UPDATE_?

From our experience with media boxes that we have, the sync interval of 12 hours is adequate.
We have some statistics gathered from deltas of offsets after some time. In our case, the drift is ~0.2 seconds per 12 hours. So we would like to decide sync interval for ourselves by having a setter. 
Also Android OS only does  _"a single time query approximately once a day"_ , and frequent NTP sync might possibly affect zap time every 10 minutes user switches a channel. 
Thanks.

stevemayhew:
Having it default to original behavior with a setter to enable the refresh (and set the time) sounds fine to me.  It is possible this issue is limited to Android Streamers as phones likely time sync with the carrier.

Also, for phones issuing the network request has at least some small battery penalty 

Feel free to add the commit with the change, you should have push access to the repo

marcbaechinger:
Ok, thanks! Will proceed that way.

marcbaechinger:
I pushed a commit with the getter and the release notes. I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review.

Let me know if you see something, Steve. Else I proceed as described above. Thanks again!

stevemayhew:
No worries.  Feel free to force push or whatever you need, branch is yours
know.

We’ll update our side so code matches

nikitapn:
In our case, users experience the issue when they bring the media box out of standby mode. During this process, we recreate the player (our wrapper with stuff) completely because the instance has been destroyed before entering standby, and "elapsedRealtimeOffsetMs" will be "TIME_UNSET" in the new instance.

I suspect we may still encounter some problems when users have been watching a live stream without turning the box off, right? In my opinion, this is a rare situation.

marcbaechinger:
Ah, I see. Yeah, obviously. Thanks for clarifying. I didn''t realized that it wouldn't re-initialize even when a new source is created. This case can be handled with this change as you say. The design of that `SntpClient` class with these static methods is, lets say, peculiar. But yeah, what you saying makes totally sense! Thanks!

Will submit as it is. If we need to enable this re-initialization inside the same `DashMediaSource` instance, we can look into this again.

Issue #1698, opened 2024-09-05
Stream/App freeze in DASH Multi-Period Live-Stream


mikef-dk:
### Version

Media3 1.4.1

### More version details

Issue reproduced in 1.4.1 and 1.4.0 - 1.3.1 is working fine.

After doing some `git bisect` we were able to pin down the issue to commit [d292f809](https://github.com/androidx/media/commit/d292f809265f2fa886da36d5f7f2e033eb4ed365). If we revert this particular commit on version 1.4.1 we cannot reproduce the issue anymore.

We observe the following Logcat messages when the issue occurs. In this particular case the "freeze" started at around 13:20:47 and application resumed at around 12:20:53.
```
13:20:45.705 AvrcpMediaPlayerList     D  sendMediaUpdate
13:20:45.705 AvrcpMediaPlayerList     I  sendMediaUpdate: Creating a one item queue for a player with no queue
13:20:45.705 AvrcpMediaPlayerList     D  sendMediaUpdate state=PlaybackState {state=3, position=1725535199867, buffered position=0, speed=1.0, updated=15856860, actions=895, custom actions=[], active item id=-1, error=null}
13:20:45.705 AvrcpTargetService       D  onMediaUpdated: track_changed=false state=true queue=false
13:20:45.705 AvrcpNativeInterface     D  sendMediaUpdate: metadata=false playStatus=true queue=false
13:20:45.705 AvrcpTargetJni           D  sendMediaUpdateNative
13:20:45.705 bt_stack                 I  [INFO:avrcp_service.cc(346)] virtual void bluetooth::avrcp::AvrcpService::SendMediaUpdate(bool, bool, bool) track_changed=0 :  play_state=1 :  queue=0
13:20:45.767 VDA                      I  [0]"ServiceDeviceTask: buffer counts: IN[90773,90773], OUT[90753,90751,90751] DEC[0->0] => DEVICE[3+0/3->0+2/14] => CLIENT[12]\n"
13:20:46.292 VDA                      I  [0]"ServiceDeviceTask: buffer counts: IN[90786,90786], OUT[90766,90764,90764] DEC[0->0] => DEVICE[3+0/3->0+2/14] => CLIENT[12]\n"
13:20:46.613 ThermalService           I  CPU temperatures: [73.3]
[...]

PR #1816, opened 2024-10-18
 Fix: Stream/App freeze in DASH Multi-Period Live-Stream Revert

Iragne:
Associated to this issue https://github.com/androidx/media/issues/1698
Revert commit

https://github.com/androidx/media/commit/d292f809265f2fa886da36d5f7f2e033eb4ed365#diff-db6bd1743a1139e100c534cebb1040df9b614cab6c49360f71fdfba8642f27a4R2063

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/ExoPlayerImplInternal.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/ExoPlayerImplInternal.java
@@ -495,7 +495,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
   @Override
   public void onPlaylistUpdateRequested() {
-    handler.removeMessages(MSG_DO_SOME_WORK);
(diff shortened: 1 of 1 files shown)


tonihei:
We haven't investigated #1698 yet to know what's happening exactly and how https://github.com/androidx/media/commit/d292f809265f2fa886da36d5f7f2e033eb4ed365 is related. However, I'm pretty sure we need to keep https://github.com/androidx/media/commit/d292f809265f2fa886da36d5f7f2e033eb4ed365 (as it fixes an inconsistency that should never have been there) and instead find the cause of the underlying issue and fix that instead. So I'll close the pull request again, but still many thanks for wanting to contribute and solve the issue!

jeanalexandre-iragne-deltatre:
@tonihei happy to see that you react on it. FYI we will remove this from our version. it's clearly visible on our playback and if you check the video dev chat we are not the only one. I understand the inconstancy point but the reality is that now we can't use any offical release for any SSAI stream.

PR #1817, opened 2024-10-18
Fix: Update the Default value for the keep alive session.

Iragne:
We saw that period can be more then 5 minutes resulting in a deletion of the session 
By default we should have a bigger value to avoid to delete it

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSessionManager.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSessionManager.java
@@ -297,7 +297,7 @@ public class DefaultDrmSessionManager implements DrmSessionManager {
   public static final int INITIAL_DRM_REQUEST_RETRY_COUNT = 3;
 
   /** Default value for {@link Builder#setSessionKeepaliveMs(long)}. */
-  public static final long DEFAULT_SESSION_KEEPALIVE_MS = 5 * 60 * C.MILLIS_PER_SECOND;
+  public static final long DEFAULT_SESSION_KEEPALIVE_MS = 8 * 60 * C.MILLIS_PER_SECOND;
 
   private static final String TAG = "DefaultDrmSessionMgr";
 


icbaker:
If you would like to use a larger value for this in your app, you can customize it using the `DefaultDrmSessionManager.Builder.setSessionKeepaliveMs(long)` method - I don't think we need to modify the default value in the library.

This keep-alive timeout is also a lot less important now that DRM sessions are pre-acquired ahead of the playback position (since ExoPlayer 2.14.0, https://github.com/google/ExoPlayer/issues/4133).

If you would still like us to consider this PR, please provide more info about why this default needs to change for **every** user of the library, including more detail on your specific case where this timeout is important (and pre-acquisition of DRM sessions is not sufficient).

jeanalexandre-iragne-deltatre:
Well this value is quite important in multi period manifest for ads. we experience the issue in a long ads break in clear.
I was expecting that the default value should manage the majority of the use case.
FYI in production today it's normal to have ads more than 5 minutes after at the our.
Yes we can customize the value but in the same time the value should cover 90% of the case.

I can share that in all our project ads are longer than 5 minutes in 70% of the case.

Happy to get your view on this

icbaker:
Just to check I'm understanding correctly: If you set this value to less than your ad break length, what do you observe? I would expect playback to still be seamless going in and out of the ad break, because the DRM session should be re-opened for the encrypted content well before the end of the ad break (this is pre-acquisition I mentioned above). So I'm guessing that the only difference is whether a session is re-used or closed and re-opened, and I think there's no user-visible difference between these cases - so is the only consideration the number of requests being made to your DRM provider, or something else?

----

You know the length of your ads better than we do, so it makes sense that you are able to customize the value more precisely to your needs.

We have to weigh this against all the uses of our library that play DRM content without any ad breaks (or other clear interruptions), and so by default 'waste' the DRM session system resource for `DEFAULT_SESSION_KEEPALIVE_MS` after every DRM protected playlist item.

jeanalexandre-iragne-deltatre:
FYI we are seeing a call to the DRM provider that is useless and we expect to not do that call.
Also in our implementation, is useless to re request DRM and it result in a stop of the playback.

I would like to improve the OOTB exo by providing a better realistic value.
But i can keep that value to me ;)

icbaker:
Thanks - I'm going to close this because it makes more sense for this specific knowledge of the length of your ad breaks to live in your app, rather than in the library.

PR #1823, opened 2024-10-23
Fix `Format.Builder` documentation

MGaetan89:
This PR fixes the documentation for `Format.Builder.setTileCountHorizontal` and `Format.Builder.setTileCountVertical`.


icbaker:
I think we still want to avoid touching the API checks in the `session.legacy` package for the same reason outlined here: https://github.com/androidx/media/pull/1549#issuecomment-2233286116

MGaetan89:
> Oh, I forgot about that, sorry. I will revert that change

Done. Now the PR only fixed the documentation for a couple of `Format.Builder` methods.

icbaker:

PR #1871, opened 2024-11-08
Added getter methods to ContainerMediaChunk & a callback method whenever a load loop is finished

colinkho:


tonihei:
Thanks! I'll import the pull request and will likely inline the the `nextLoadPositionUs` and the `isLoadCanceled` as parameters to `onLoadIterationEnded` given they are meant to be used in there. Please let me know if that wouldn't work for you.

tonihei:
> will likely inline the the nextLoadPositionUs and the isLoadCanceled as parameters to onLoadIterationEnded given they are meant to be used in there

Kept them in their originally proposed place to allow easier access to this information from other methods as well

PR #1889, opened 2024-11-15
Specify androidx.activity:activity-compose as 1.9.0

yschimke:
The Compose BOM doesn't provide this.
https://developer.android.com/develop/ui/compose/bom/bom-mapping

```
Failed to resolve: androidx.activity:activity-compose
```


yschimke:
Feel free to consider this a bug report, close and reopen internally if simpler.

PR #1898, opened 2024-11-18
Add AC-4 Level-4 format support for DASH

ybai001:
The main purpose of this pull request is to add support of parsing the number of channels from the value attribute of an AudioChannelConfiguration with schemeIdUri "tag:dolby.com,2015:dash:audio_channel_configuration:2015" as defined by table A.27 in ETSI TS 103 190-2 v1.2.1.


rohitjoins:

libraries/exoplayer/src/test/java/androidx/media3/exoplayer/e2etest/Mp4PlaybackTest.java
```diff
@@ -50,6 +50,7 @@ public static ImmutableList<String> mediaSamples() {
         "sample_ac3_fragmented.mp4",
         "sample_ac3.mp4",
         "sample_ac4_fragmented.mp4",
+        "sample_ac4_level4_fragmented.mp4",
```
rohitjoins, line 53: `Mp4PlaybackTest` fails with this newly added file as `channelCount` returned is either `18` or `21` and `Util.getAudioTrackChannelConfig` returns `CHANNEL_INVALID`.

Does these channel counts needs to map to 24 for it work as done in `DefaultTrackSelector.canBeSpatialized`?

Also fails with `sample_ac4_level4.mp4` for the same reason.

ybai001: Oh, you said I shouldn't update this code since you have submitted the internal review.

rohitjoins: Please feel free to update the code as required. I was not able to send it for internal review as it has failing tests. 

Once done, I'll re-sync the changes. Thanks!

ybai001: OK, I see.

ybai001: @rohitjoins , I updated the code by removing "sample_ac4_level4_fragmented.mp4" from Mp4PlaybackTest. The reasons are:
(1) Mp4PlaybackTest is a Robolectric test and it doesn't run on an actual physical device. When running "sample_ac4_level4_fragmented.mp4" for Mp4PlaybackTest, the fragmentedmp4extractor can parse the stream info successfully and channelCount = 21 is set for AudioFormat. This is expected behavior. But, in Robolectric test, there is no real AC-4 decoder used but a fake decoder named exotest.audio.ac4. If the input stream format with channelCount 21, the output buffer of this decoder is also with 21 channels. This results in the outputChannelConfig is [INVALID](https://github.com/androidx/media/blob/release-1.5.0-rc02/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L734) and an [exception](https://github.com/androidx/media/blob/release-1.5.0-rc02/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L777) is thrown. In real device, our AC-4 decoder will output buffer with channelCount = 24 so that there is no issue.
(2) Even if I mandatory modify the stream channel count from 21 to 24 [here](https://github.com/androidx/media/blob/main/libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java#L419), the test still failed because current Robolectric test is based on [SDK version 30](https://github.com/androidx/media/blob/main/constants.gradle#L21) and 24 channel is not supported in method [AudioTrack.getMinBufferSize()](https://github.com/androidx/media/blob/main/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L2113) and an [exception](https://github.com/androidx/media/blob/main/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L2114) is triggered again.

In summary, this test case failed because (1) A fake AC-4 decoder is used in Robolectric test and its behavior is different from real AC-4 decoder; (2) Current Robolectric test is based on SDK version 30. I didn't find a good workaround to solve this issue so that I suggest to remove this test case in current release. (Actually, I have test all kind of AC-4 streams (including local and DASH streaming) on my test device and there is no problem.)

rohitjoins: Makes sense. Thanks for clarifying. I'll send the latest code for internal review.

rohitjoins: I tried playing these AC4-level4 sample files on a Samsung device (API level 34), but it fails with an error in [AudioTrack.getMinBufferSize()](https://github.com/androidx/media/blob/main/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L2114) due to an assertion failure.

Which devices are you testing this on? What value is returned for encoding in MediaCodec.MediaFormat? For us, it currently fails because MediaCodec.MediaFormat reports AudioFormat.ENCODING_PCM_16BIT. This encoding has a maximum supported channel count of 12, but the MediaFormat specifies a channel count of 24. The error occurs because the channel count exceeds the maximum allowed for the given encoding.

rohitjoins: We can probably add E2E tests by customizing the `ShadowMediaCodecConfig` to change the channelCount as `MediaCodec` would do and then run it on API levels >= 32. But currently that fails because of the reason stated above.

rohitjoins:
Merged as https://github.com/androidx/media/commit/089eaa1d5d06832f5d8e1f74b85301b875b66446. Keeping the PR open to discuss unresolved https://github.com/androidx/media/pull/1898#discussion_r1850817681.

libraries/exoplayer/src/test/java/androidx/media3/exoplayer/e2etest/Mp4PlaybackTest.java
ybai001: > I tried playing these AC4-level4 sample files on a Samsung device (API level 34), but it fails with an error in [AudioTrack.getMinBufferSize()](https://github.com/androidx/media/blob/main/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java#L2114) due to an assertion failure.
> 
> Which devices are you testing this on? What value is returned for encoding in MediaCodec.MediaFormat? For us, it currently fails because MediaCodec.MediaFormat reports AudioFormat.ENCODING_PCM_16BIT. This encoding has a maximum supported channel count of 12, but the MediaFormat specifies a channel count of 24. The error occurs because the channel count exceeds the maximum allowed for the given encoding.

I'll write an email to you for technical details since some info is confidential.

PR #1905, opened 2024-11-19
Make MediaCodecVideoRenderer::shouldUsePlaceholderSurface protected.

khouzam:
This enables a derived renderer to disable the placeholder surface.

Not having a placeholder surface allows to delay instantiating the codecs until we have a surface. This allows to have more players ready without having their decoders loaded.

This follows the mode that `shouldUseDetachedSurface` is protected and overridable by a derived class.

Since `getSurfaceForCodec` must call `hasSurfaceForCodec`, the check for the placeholder surface in `getSurfaceForCodec` should not trigger if we disable the placeholder surface as `hasSurfaceForCodec` would have returned false.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -11,6 +11,10 @@
 *   Audio:
     *   Make `androidx.media3.common.audio.SonicAudioProcessor` final.
 *   Video:
+    *   Change `MediaCodecVideoRenderer.shouldUsePlaceholderSurface` to
+        protected so that applications can override to block usage of
+        placeholder surfaces
+        ([#1905](https://github.com/androidx/media/pull/1905)).
 *   Text:
 *   Metadata:
 *   Image:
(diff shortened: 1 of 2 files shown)


microkatz:
@khouzam 

Thanks for submitting your request! Just a curiosity, have you looked at [MediaCodecVideoRenderer::shouldInitCodec] for your use case? Its already overridable and utilized by `MediaCodecRenderer` to decide if the decoder should be initialized.

khouzam:
@microkatz Thanks for the suggestion, I'll experiment with it, but while this can prevent the initialization of the codec, it will still create the `PlaceholderSurface` when a `null` surface gets set.

khouzam:
@microkatz We've tried your suggestion. The results don't seem very encouraging. We see significant increases in ANRs, memory kills and native crashes. The additional resources used by the placeholder surface seems to really conflict with some of the fine tuning that we've done for our scenarios.

tonihei:
@microkatz: @khouzam  and I discussed this code path in person as well and concluded that the protected access makes sense because the method is also used from `setOutput` where there is no other way to prevent the usage of placeholder surfaces if the codec already exists and the surface is removed.

microkatz:

PR #1917, opened 2024-11-25
Add new consecutiveDroppedFrames callback

khouzam:
This is similar to the `droppedFrames` callback, but represents the number of consecutive frames that we dropped before rendering a frame or seeking or stopping the renderer.

While we already have a `maxConsecutiveDroppedFrame` available in the `DecoderCounters`, this doesn't provide enough visibility into the actual statistics of dropped frames.

If we get 200 dropped frames and a `maxConsecutive` of 20, we don't know if we dropped 20 frames in a row once and then dropped a single frame 180 times or if we dropped 20 frames 10 times.

We could add some code on our `OnDroppedFrames` callback to estimate if two calls are for consecutive frames, but that seems very fragile.

There is an open issue currently on how often to notify, and my preference would be to add a new `minConsecutiveDroppedFramesToNotify` similar to the `maxDroppedFramesToNotify` but that would only notify if more than X consecutive frames were dropped.


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/video/VideoRendererEventListener.java
```diff
@@ -205,6 +218,16 @@ public void droppedFrames(int droppedFrameCount, long elapsedMs) {
       }
     }
 
+    /** Invokes {@link VideoRendererEventListener#onConsecutiveDroppedFrames(int, long)}. */
+    public void consecutiveDroppedFrames(int consecutiveDroppedFrameCount, long elapsedMs) {
+      if (handler != null) {
+        handler.post(
+            () ->
+                castNonNull(listener)
+                    .onConsecutiveDroppedFrames(consecutiveDroppedFrameCount, elapsedMs));
```
microkatz, line 227: Can you add a test to `MediaCodecVideoRendererTest` to make sure that provided a very late buffer that is dropped that this event is invoked? You can look at `MediaCodecVideoRendererTest::render_withLateBuffer_dropsBuffer`. May also be helpful to have unit tests for the event testing the output for if dropped frames occur prior to any successful renders and between successful renders.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/analytics/AnalyticsCollector.java
```diff
@@ -250,6 +250,19 @@ void onVideoInputFormatChanged(
    */
   void onDroppedFrames(int count, long elapsedMs);
 
+  /**
+   * Called to report the number of consecutive frames dropped by the video renderer. Consecutive
+   * dropped frames are reported when a frame is renderered after a the previous consecutive frames
+   * were not rendered optionally, whenever the consecutive dropped frame count is above a specified
+   * threshold whilst the renderer is started.
```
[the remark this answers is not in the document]
khouzam: The comment has been fixed and should be much clearer as to the behavior.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/video/VideoRendererEventListener.java
khouzam: Added tests covering both the API being invoked and only being invoked when there are enough frame drops.

colinkho:
Hi @microkatz @tonihei @icbaker , is this ready to be pushed upstream? I see Gilles has addressed most comments

microkatz:
@khouzam 

Apologies for the delay!

I was holding off from additional comments and edits as we instituted a new Builder pattern into MediaCodecVideoRenderer rather than adding new constructors for all these new variables/thresholds etc. This PR then got lost in the weeds after that was completed here, https://github.com/androidx/media/commit/decfb9b0a9da075eba4629f470fe116a026f61e7.

There are a few requests for the PR. First, would you be able to rebase to the tip of main and transition to a Builder parameter for your consecutiveDroppedFrames threshold?

Secondly, in discussions with the team, it was thought better if we could still have a single listener event for dropped frames rather than two. The idea would be to add an additional parameter to onDroppedFrames for maxConsecutive. Currently, onDroppedFrames may be invoked when the dropped frames threshold is reached. This would be changed so that the invocation would occur where you are currently checking for consecutive frame drops notification.  At onStopped, onDisabled, and after a successful frame render. 

Would this suggestion still work to provide you the granularity of analytics you need?

Please reach out if you have additional questions.

khouzam:
Thanks for the feedback @microkatz, a single listener would also work, as long as we make sure to invoke it appropriately. We'll update the pull request with a rebase and the changes.

PR #1936, opened 2024-12-01
Reevaluate buffer during a playWhenReady change

colinkho:


microkatz:
@colinkho

Sorry about the delay! 

Looks like a simple change to me and I understand your desire to check trackSelection.shouldCancelLoad whenever playback is canceled. Currently your PR has the command running also when `setPlayWhenReady` is set to true? When `setPlayWhenReady` is set to true, ExoPlayer immediately runs a `doSomeWork` task which in `maybeUpdateLoadingPeriod` runs `queue.reevaluateBuffer(rendererPositionUs);`.

What would you think about moving your suggested code change to inside the conditional for when `!shouldPlayWhenReady()`? So that if setPlayWhenReady is called with true then `queue.reevaluateBuffer` is not called twice?

colinkho:
> @colinkho
> 
> Sorry about the delay!
> 
> Looks like a simple change to me and I understand your desire to check trackSelection.shouldCancelLoad whenever playback is canceled. Currently your PR has the command running also when `setPlayWhenReady` is set to true? When `setPlayWhenReady` is set to true, ExoPlayer immediately runs a `doSomeWork` task which in `maybeUpdateLoadingPeriod` runs `queue.reevaluateBuffer(rendererPositionUs);`.
> 
> What would you think about moving your suggested code change to inside the conditional for when `!shouldPlayWhenReady()`? So that if setPlayWhenReady is called with true then `queue.reevaluateBuffer` is not called twice?

That works for us! @microkatz

colinkho:
@microkatz I have made the change and updated this PR

microkatz:

PR #1963, opened 2024-12-10
Add MediaCodecInfo to MediaCodecRenderer::onReadyToInitializeCodec

khouzam:
This adds the MediaCodecInfo to the `onReadyToInitializeCodec` to allow for renderers to know which codec is being initialize, not only which format the codec will be initialized for. With multiple codecs supporting a format, knowing which one is being used would be helpful.


microkatz:

PR #1999, opened 2024-12-26
[fix #1997] restrict milliseconds field in VTT and SRT text caption time codes to 3 digits

### LINKED ISSUE #1997: subtitles parser doesn't handle time codes that include more than 3 digits in the milliseconds field
### Version

Media3 main branch

### More version details

you ask: why would the milliseconds field ever contain more than 3 digits
I reply: it shouldn't, but for whatever reason.. captions on PBS do

example:
* [VTT](https://ga.pbs-video.pbs.org/live2vod/videos/newshour/2024-12-23/fullshow_captions.vtt) for [PBS News Hour](https://www.pbs.org/show/newshour/)
  - episode: [12/23/2024](https://www.pbs.org/video/december-23-2024-pbs-news-hour-full-episode-1734930001/)
  - notice that all time codes have 6 digits in the milliseconds field
    * ex: `00:00:00.000000 --> 00:00:00.100000`
    * I didn't dig into the code to see how the parser is processing this field, but I can tell you that in ExoPlayer.. no subtitles show for a while.. and then (at about 7 minutes) a giant dump of text covers the screen
  - quick test:
    * I downloaded the file
    * opened it in a text editor
    * performed the following regex operation on its contents:
      - search: `(\.\d{3})\d{3}`
      - replace: `$1`
    * note: this effectively truncates the milliseconds field from 6 digits to 3 digits
    * loaded the resulting vtt file in ExoPlayer
    * issue fixed.. captions play as expected

note: I haven't yet tested their SRT captions; I suspect the same time code format. TBD..

OS: Android
app: ExoAirPlayer [v3.7.0](https://github.com/warren-bank/Android-ExoPlayer-AirPlay-Receiver/releases/tag/v3.7.0)
lib: AndroidX Media3 [v1.5.0](https://github.com/androidx/media/releases/tag/1.5.0)

_No response_

Not tested

### Reproduction steps

load captions from: [VTT](https://ga.pbs-video.pbs.org/live2vod/videos/newshour/2024-12-23/fullshow_captions.vtt)

### Expected result

captions display at the correct time

### Actual result

captions don't display at the correct time

### Media

* [HLS video URL](https://ga.pbs-video.pbs.org/live2vod/videos/newshour/2024-12-23/fullshow_2.m3u8)
* [referer header](https://www.pbs.org/video/december-23-2024-pbs-news-hour-full-episode-1734930001/)
  - note: I don't remember if PBS requires this header or not.. only mentioning it in case they do
* [external VTT captions URL](https://ga.pbs-video.pbs.org/live2vod/videos/newshour/2024-12-23/fullshow_captions.vtt)

-or-

* [ExoAirPlayer sender](http://webcast-reloaded.frii.site/airplay_sender.html#/watch/aHR0cHM6Ly9nYS5wYnMtdmlkZW8ucGJzLm9yZy9saXZlMnZvZC92aWRlb3MvbmV3c2hvdXIvMjAyNC0xMi0yMy9mdWxsc2hvd18yLm0zdTgjdmlkZW8ubTN1OA%253D%253D/subtitle/aHR0cHM6Ly9nYS5wYnMtdmlkZW8ucGJzLm9yZy9saXZlMnZvZC92aWRlb3MvbmV3c2hvdXIvMjAyNC0xMi0yMy9mdWxsc2hvd19jYXB0aW9ucy52dHQjdGV4dC52dHQ%253D/referer/aHR0cHM6Ly93d3cucGJzLm9yZy92aWRlby9kZWNlbWJlci0yMy0yMDI0LXBicy1uZXdzLWhvdXItZnVsbC1lcGlzb2RlLTE3MzQ5MzAwMDEv)

### Bug Report

warren-bank:


warren-bank:
For reference, the [VTT spec](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API/Web_Video_Text_Tracks_Format#ttt) says:

The timestamps can be specified in one of the following two formats:
* `mm:ss.ttt`
* `hh:mm:ss.ttt`

Where `ttt` represents milliseconds and must be between 000 and 999, inclusive.

warren-bank:
Also for reference, though there is no official "spec" for SRT, [this page](https://www.speechpad.com/captions/srt) summarizes the common conventions:

Timecodes have the following format:
* `hours:minutes:seconds,milliseconds`

Note that a comma is used instead of a period to separate the seconds and milliseconds.
This is because the SubRib format originated in France.
The first three components (hours, minutes, and seconds) each consist of two zero-padded digits,
and the last component (milliseconds) is __three zero-padded digits__.

(emphasis added)

icbaker:
As you've noted above, the WebVTT spec unambiguously requires exactly 3 decimal digits. Any media not conforming to this can expect undefined behaviour when played. ExoPlayer currently parses an undefined number of digits and assumes it's milliseconds (i.e. assumes it was 3 digits). Your fix here changes this to truncate more than 3 digits, but doesn't change the way we handle less than 3 digits. I'd be tempted to instead change the regex to match **exactly** 3 digits. Then the invalid subtitles with more or less than 3 digits will simply fail to parse (instead of successfully parsing with an incorrect timestamp, and then showing up at wrong/surprising times during playback). 

----

As you've said, SubRip has a less clearly defined spec than WebVTT, but this is the first time I'm aware of content with more (or fewer) decimal places - and the fact both SubRip and WebVTT files come from the same content producer suggests this is a localised bug in their content creation pipeline - so I suggest we can continue assuming SubRip content has exactly 3 decimal places too (or no decimal part at all - ExoPlayer already handles this).

---

In general, the specs exist to ensure a common understanding between media producers and consumers. Adding workarounds that accept media outside the spec introduces confusion - because suddenly something that works with player A doesn't work with player B. In this case I'd recommend reporting this as a bug to the content producer and getting them to fix the media to conform to the spec.

I don't think we should introduce additional complexity to our parsing code to correctly handle an arbitrary number of decimal places when this is clearly not permitted by the spec(s) in question.

icbaker:
If you switch to modify the regex, please also add some tests to this PR - in these files:
* https://github.com/androidx/media/blob/main/libraries/extractor/src/test/java/androidx/media3/extractor/text/webvtt/WebvttParserTest.java
* https://github.com/androidx/media/blob/main/libraries/extractor/src/test/java/androidx/media3/extractor/text/subrip/SubripParserTest.java

warren-bank:
I rather like that the parser(s) attempt to be as forgiving as possible.
When the data matches the spec, there's no issue.
But when it doesn't, I wouldn't want the video player to simply fail;
I'd much prefer that it make sane choices.

If there are fewer than 3 digits for ms, then the parser behaves as I would expect:
* `00:00.1` = 1ms
* `00:00.10` = 10ms
* `00:00.100` = 100ms

The question is.. if there are ever more than 3 digits,
then is it more likely that the intention is to convey a value larger than 1 second,
or that (for whatever reason) it's applying zero-padding on the right,
which can be safely removed?
I would assume the latter.. hence my PR.
Maybe others would assume differently.. idk.
A discussion.. and consensus.. would be a good idea.

Thanks for your response.

icbaker:
> If there are fewer than 3 digits for ms, then the parser behaves as I would expect:
> 
> * `00:00.1` = 1ms
> * `00:00.10` = 10ms
> * `00:00.100` = 100ms

Interesting - this is not what I would expect. `00:00.1` looks to me like '0 minutes and 0.1 seconds', which is 100ms (not 1ms).

-----

> your comments about strictly enforcing spec reminded me of when browsers attempted a migration to XHTML..
> and the web pretty much broke over night..
> until we all begrudgingly accepted that quirks mode is a necessary evil.

These problems come about when either specs are loosely documented or loosely implemented, and content producers then come to rely on these loose consumer implementations. If specs are strictly implemented by consumers from the beginning, then producers are forced (from the beginning) to conform to the spec. This avoids consumers having to add additional complexity which penalises well-behaved content produces in order to accommodate badly-behaved ones. In this case, ExoPlayer has supported WebVTT and SubRip with exactly 3 decimal places for many years, and this is the first time I'm aware of content which doesn't conform to this expectation/requirement. I don't think this is a "web broke overnight" situation. The WebVTT ecosystem seems to be working just fine with exactly 3 decimal digits, there's just a single content producer with a clear bug in their systems, which should be easy to fix.

warren-bank:
> Interesting - this is not what I would expect. 00:00.1 looks to me like '0 minutes and 0.1 seconds', which is 100ms (not 1ms).

The millis field is: `"1"`, and ..
`Long.parseLong("1")` = 1L

> I don't think this is a "web broke overnight" situation. The WebVTT ecosystem seems to be working just fine with exactly 3 decimal digits

whoa.. please calm down.. I didn't mean any offense.
What I did mean was to draw an amusing metaphor to what can happen when a parser decides that anything that isn't perfectly spec compliant must be invalid.

In any case, obviously this is a weird edge case that only rarely happens.
This is the first time that I encountered it.
I just thought I'd bring it to your attention, and offer a simple solution.
Take it, leave it, do something else.. it's entirely your call.
But again.. I'd urge to support quirky things like this..
where it's pretty obvious what what intended, and normalization can be done in the parser.

PR #2008, opened 2024-12-30
Add support for SSA (v4+) MarginL, MarginR, MarginV style

szaboa:
Reopening https://github.com/google/ExoPlayer/pull/10169. 

This PR is about adding support for `MarginL`, `MarginR`, `MarginV` from both the `Style` and `Dialogue` lines. I've used VLC as reference.

- Non-zero margins defined in `Dialogue` lines takes priority over the margins in `Style` lines.
- In case we have a `{\pos}` override, then no margins will be applied neither from `Dialogue` or `Style` (same as in VLC).
- In case we have a `{\an}` override, then VLC ignores the `Dialogue` margin but applies the `Style` margin which makes no sense for me, so in our case I've just ignored both `Dialogue` and `Style` margins (same as `{\pos}` case).

_Suggestion_: maybe we could update the `https://storage.googleapis.com/exoplayer-test-media-1/ssa/test-subs-position.ass` subtitle file with these new lines, then no need for a new media in `media.exolist.json`.


peerless2012:
Add ass feature for exoplayer use Java code mabe a huge work。

libraries/extractor/src/main/java/androidx/media3/extractor/text/ssa/SsaParser.java
```diff
@@ -461,11 +426,104 @@ private static Cue createCue(
...
 
+    // Apply margins if there are no overrides and we have valid positions.
+    if (styleOverrides.alignment == SsaStyle.SSA_ALIGNMENT_UNKNOWN
+        && styleOverrides.position == null
+        && cue.getPosition() != Cue.DIMEN_UNSET
+        && cue.getLine() != Cue.DIMEN_UNSET) {
+
+      // Margin from Dialogue lines takes precedence over margin from Style line.
+      float marginLeft = dialogueMarginLeft != 0f
+          ? dialogueMarginLeft / screenWidth
```
icbaker, line 441: nit: imo double-nested ternaries are quite hard to read

I would suggest unrolling this to:

```java
float marginLeft;
if (dialogueMarginLeft != 0f) {
  marginLeft = dialogueMarginLeft / screenWidth;
} else if (style != null) {
  marginLeft = style.marginLeft / screenWidth;
} else {
  marginLeft = 0f;
}
```

libraries/extractor/src/main/java/androidx/media3/extractor/text/ssa/SsaParser.java
```diff
@@ -461,11 +426,104 @@ private static Cue createCue(
...
+          : style != null ? style.marginLeft / screenWidth : 0f;
+      float marginRight = dialogueMarginRight != 0f
+          ? dialogueMarginRight / screenWidth
+          : style != null ? style.marginRight / screenWidth : 0f;
+
+      // Apply margin left, margin right.
+      if (SsaStyle.hasLeftAlignment(style)) {
```
icbaker, line 448: Shouldn't you use the `alignment` local from L411 here, rather than just the alignment derived from the `style`?

(related to comment above)

libraries/extractor/src/main/java/androidx/media3/extractor/text/ssa/SsaParser.java
```diff
@@ -461,11 +426,104 @@ private static Cue createCue(
       cue.setPosition(styleOverrides.position.x / screenWidth);
       cue.setLine(styleOverrides.position.y / screenHeight, LINE_TYPE_FRACTION);
     } else {
-      // TODO: Read the MarginL, MarginR and MarginV values from the Style & Dialogue lines.
       cue.setPosition(computeDefaultLineOrPosition(cue.getPositionAnchor()));
       cue.setLine(computeDefaultLineOrPosition(cue.getLineAnchor()), LINE_TYPE_FRACTION);
     }
 
+    // Apply margins if there are no overrides and we have valid positions.
+    if (styleOverrides.alignment == SsaStyle.SSA_ALIGNMENT_UNKNOWN
```
icbaker, line 434: why do we check specifically `styleOverrides.alignment` here when we've already resolved `alignment` from both `styleOverrides` and `style` on L411?

(related comment below)

libraries/extractor/src/main/java/androidx/media3/extractor/text/ssa/SsaParser.java
```diff
@@ -461,11 +426,104 @@ private static Cue createCue(
...
+      if (SsaStyle.hasLeftAlignment(style)) {
+        cue.setPosition(cue.getPosition() + marginLeft);
+        cue.setSize(1 - marginRight - marginLeft);
+      } else if (SsaStyle.hasRightAlignment(style)) {
+        cue.setPosition(cue.getPosition() - marginRight);
+        cue.setSize(1 - marginRight - marginLeft);
+      } else {
+        // Center alignment or unknown.
+        cue.setPosition(cue.getPosition() + (marginLeft - marginRight) / 2);
+        cue.setSize(1 - marginRight - marginLeft);
```
icbaker, line 457: this line is the same is all 3 `if` blocks - could probably pull it out?

libraries/extractor/src/main/java/androidx/media3/extractor/text/ssa/SsaParser.java
```diff
@@ -461,11 +426,104 @@ private static Cue createCue(
...
+      // Apply margin left, margin right.
+      if (SsaStyle.hasLeftAlignment(style)) {
+        cue.setPosition(cue.getPosition() + marginLeft);
+        cue.setSize(1 - marginRight - marginLeft);
+      } else if (SsaStyle.hasRightAlignment(style)) {
+        cue.setPosition(cue.getPosition() - marginRight);
+        cue.setSize(1 - marginRight - marginLeft);
+      } else {
+        // Center alignment or unknown.
+        cue.setPosition(cue.getPosition() + (marginLeft - marginRight) / 2);
```
icbaker, line 456: I'm not sure I understand the maths here (or for the `hasRightAlignment` case above).

`Cue.position` always measures from the left hand side of the screen (for horizontal text cues), no matter what value `Cue.textAlignment` has. So shouldn't it always be increased by `marginLeft`?

----

In general applying these margins after the size/position has already been calculated seems hard to reason about - why not apply them as part of computing the size and position?

icbaker (changes requested):
I only part-reviewed this, because my main comment is about whether we could make the code easier to reason about by applying the margins when initially calculating position/size, rather than trying to apply them in a second pass - and this might result in relatively major changes.

Issue #2048, opened 2025-01-16
Cache opened widevine DRM Sessions until `Player.release()`


stevemayhew:
###  Use case description
The use case involves streaming a large linear channel list (>100 channels) consisting of Live HLS and DASH playlists, protected by Widevine. Playback is continuous, and "channel zapping" or "channel surfing" is a frequent activity. Channels may be licensed individually or in packages.

Most Android streaming hardware supports a significant number of open `MediaDrm` session slots (e.g., Google TV Chromecast, based on AmLogic SoC, supports 100 active license slots). 

Currently, ExoPlayer retains open sessions for `MediaItem`s in a timed cache, however this cache is flushed whenever any `MediaSource` that references it is released.

The [Playlist](https://developer.android.com/media/media3/exoplayer/playlists) avoids this as the `MediaItem`'s in it are associated with `MediaSource`s are not fully released until the `MediaSourceList` (playlist) is released.  

In contrast, changing channels by calling `Player.setMediaItems()` with a single channel's `MediaItem` releases the current `MediaSource`, effectively flushing the DRM session cache every time.

The use case precludes the use of a Playlist due to the following challenges:  
1. The Playlist API does not directly preclude multiple live items, however it makes little sense contextually
2. Retaining a fully prepared **live** `MediaSource` consumes significant CPU, memory, and network bandwidth, as it retains source buffers and continues polling live playlist refreshes.  
3. Linear channel provisioning often occurs on-the-fly, making it impractical to pre-create a playlist of 100+ channels before playback starts.  

### Related Requests
The issue https://github.com/androidx/media/issues/394 presented a very similar request, however it looked like the request was to persist licnese after process exit.   

The Widevine CDM platform will not support this as the open MediaDrm sessions owned by a process are released on process exit.  In fact it is not really valid to retain open MediaDrm sessions while the Activity is not in the foreground as they are a shared hardware resource.

### Proposed solution

Pull request https://github.com/androidx/media/pull/2049

### Alternatives considered

The Playlist could  be made acceptable, but some mechanism for quiescing the `Timeline.Window` live windows that are not currently playing.  The quiesce would stop the playlist tracker and free cached media without fulling releasing the `MediaSource`.   This would be a much more significant change than what is proposed in the pull request.

icbaker:
> however this cache is flushed whenever any `MediaSource` that references it is released.

I'm not sure I completely agree with this statement.

I think a more precise statement would be:

[...]

Issue #394, opened 2023-05-16
DRM Key-caching for Streaming Licenses


carlosmesquita:
Hello,

I'll try to be as concise as possible:

I am trying to cache streaming/playback DRM keys, for Widevine specifically, so I don't have to request a new key everytime new media is requested. Imagine an IPTV scenario, where everytime the user swaps channel, a new license is being requested at the moment, but it is not needed, because the same license should be used!

I have tried reimplementing the DrmSessionManager and MediaDrmCallback, but without success. 

I have also tried using the **OfflineLicenseHelper**, but it only works for offline licenses, and my DRM server doesn't allow fetching offline licenses for streaming content, which makes total sense to me. I tried reimplementing the **OfflineLicenseHelper**, and calling the **blockingKeyRequest** method passing **MODE_PLAYBACK** as parameter, but it didn't work... I assume it's some kind of security mechanism?

Does anyone have a use case similar to this? Can anyone help me understanding how can i get the current DRM Key being used, so i can store in the local KeyStore, and use it later for new streaming content, so the Player doesn't request a new DRM Key?

Best regards, and thanks in advance.

icbaker:
Streaming keys can be re-used by using the same 'session'. In ExoPlayer that generally means using the same `DrmSession` instance. ExoPlayer's `DefaultDrmSessionManager` will aim to re-use session in two ways:
* If `multiSession` is false then all subsequent calls to `acquireSession` with a non-null `format.drmInitData` return the same `DrmSession` instance.
* If `multiSession` is true then calls to `acquireSession` return an existing `DrmSession` instance if there's one where `DefaultDrmSession.schemeDatas` matches the `schemeDatas` derived from `format.drmInitData`.

When switching between different MediaItems in the playlist (which it sounds like what you're doing with 'swapping channel'?) then this re-use can only work if the same `DefaultDrmSessionManager` instance is used for both items. This re-use is decided based on the equality of `MediaItem.drmConfiguration` here:
https://github.com/androidx/media/blob/3c01488f8d478caaa54038431ee6339444ea6e04/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/drm/DefaultDrmSessionManagerProvider.java#L83-L87

I suggest you spend some time debugging (and maybe add some logging) in order to determine whether:
* Firstly, the same `DefaultDrmSessionManager` instance is used for both the before and after playbacks after you switch channels in your app.
[...]

PR #2049, opened 2025-01-16
DrmSession Caching Proposal

stevemayhew:
This series of commits implements extending the functionality of the `DefaultDrmSessionManagersetSessionKeepaliveMs()` to allow keeping open `MediaDrm` sessions beyond the live time of any one `MediaSource`

Our fork of ExoPlayer is using this implementation to support a large live channel lineup while reducing the tune load on the DRM License Server significantly (we see hit ratios over 50% on AmLogic hardware).

The related issue Is #2048 

One improvement would be to add a "player release listener" API to ExoPlayer to allow automatically calling the `releaseAllSessions()` method on `DefaultDrmSessionManager` rather than requiring the application to do this manually.


stevemayhew:
This is also a possible solution for issue androidx/media#394

icbaker:
Closing this - see https://github.com/androidx/media/issues/2048#issuecomment-2682662780 for the app-side solution we settled on without any changes to the library.

PR #2107, opened 2025-02-04
Deprecate `Util.DEVICE`, `Util.MANUFACTURER` and `Util.MODEL`

MGaetan89:
This PR deprecates `Util.DEVICE`, `Util.MANUFACTURER` and `Util.MODEL` and replaces them with `Build.DEVICE`, `Build.MANUFACTURER` and `Build.MODEL`, respectively.

This addresses the corresponding TODOs in the project and [issue 384699964](http://issuetracker.google.com/issues/384699964) (I hope, since it is not public).

**Note:** I've submitted all three replacements in a single PR, since they are part of the same ticket, but I've used one commit for each of them. If you prefer multiple PRs, let me know.

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/util/GlUtil.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/GlUtil.java
@@ -33,6 +33,7 @@ import android.opengl.GLES20;
 import android.opengl.GLES30;
 import android.opengl.GLUtils;
 import android.opengl.Matrix;
+import android.os.Build;
 import androidx.annotation.IntRange;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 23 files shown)


icbaker:

MGaetan89:
@icbaker since this has been released in 1.6.0, is it ok if I send a PR to remove these fields?

icbaker:
We generally don't remove deprecated things so quickly unless there's a pressing need or they are causing significant tech debt (e.g. we are still working through removing symbols that were marked deprecated in exoplayer 2.15.0). Is there a need to prioritize these?

MGaetan89:
No specific needs from me. It was just a question to know if it was worth opening a PR or not.

PR #2113, opened 2025-02-07
Allow shouldStartPlayback have the last rebuffer time

colinkho:
- **Add lastRebufferRealtimeMs to LoadControl.Parameters**

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/LoadControl.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/LoadControl.java
@@ -15,6 +15,7 @@
  */
 package androidx.media3.exoplayer;
 
+import android.os.SystemClock;
 import androidx.media3.common.C;
 import androidx.media3.common.Player;
 import androidx.media3.common.Timeline;
@@ -81,6 +82,18 @@ public interface LoadControl {
     public final long targetLiveOffsetUs;
 
     /**
+     * Sets the time at which the last rebuffering occurred, in milliseconds since boot including
+     * time spent in sleep.
+     *
+     * <p>The time base used is the same as that measured by {@link SystemClock#elapsedRealtime}.
+     *
+     * <p><b>Note:</b> If rebuffer events are not known when the load is started or continued, or if
+     * no rebuffering has occurred, or if there have been any user interactions such as seeking or
+     * stopping the player, the value will be set to {@link C#TIME_UNSET}.
+     */
+    public final long lastRebufferRealtimeMs;
+
+    /**
      * Creates parameters for {@link LoadControl} methods.
      *
      * @param playerId See {@link #playerId}.
@@ -92,6 +105,7 @@ public interface LoadControl {
      * @param playWhenReady See {@link #playWhenReady}.
      * @param rebuffering See {@link #rebuffering}.
      * @param targetLiveOffsetUs See {@link #targetLiveOffsetUs}.
+     * @param lastRebufferRealtimeMs see {@link #lastRebufferRealtimeMs}
      */
     public Parameters(
         PlayerId playerId,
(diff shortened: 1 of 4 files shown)


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/LoadControl.java
```diff
@@ -80,6 +81,18 @@ final class Parameters {
...
+     * Sets the time at which the last rebuffering occurred, in milliseconds since boot including
+     * time spent in sleep.
+     *
+     * <p>The time base used is the same as that measured by {@link SystemClock#elapsedRealtime}.
+     *
+     * <p><b>Note:</b> If rebuffer events are not known when the load is started or continued, or if
+     * no rebuffering has occurred, or if there have been any user interactions such as seeking or
+     * stopping the player, the value will be set to {@link C#TIME_UNSET}.
+     */
+    public final long lastRebufferRealtimeMs;
```
marcbaechinger, line 94: Please make this `Us` so it is consistent to the other position units we use in LoadControl.Parameters

libraries/exoplayer/src/test/java/androidx/media3/exoplayer/DefaultLoadControlTest.java
```diff
@@ -91,7 +91,8 @@ public void shouldContinueLoading_untilMaxBufferExceeded() {
                     SPEED,
                     /* playWhenReady= */ false,
                     /* rebuffering= */ false,
-                    /* targetLiveOffsetUs= */ C.TIME_UNSET)))
+                    /* targetLiveOffsetUs= */ C.TIME_UNSET,
+                    C.TIME_UNSET)))
```
marcbaechinger, line 95: please add inline comments when using a constant, here and elsewhere

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/LoadControl.java
icbaker: I guess the difference between this and `playbackPositionUs` & `bufferedDurationUs` is that this value is measured in the `SystemClock.elapsedRealtimeMs` timebase, rather than the 'media timebase' (for which we pretty consistently use microseconds inside ExoPlayer). 

We seem to have a mixture of micros and millis throughout out the codebase for this:
* https://github.com/search?q=repo%3Aandroidx%2Fmedia%20realtimeus&type=code (36 files)
* https://github.com/search?q=repo%3Aandroidx%2Fmedia%20realtimems&type=code (41 files)

icbaker: These values are already kind of (deliberately) 'inconsistent' with each other by being measured in different time bases anyway, whether they use the same units or not. As in, it would probably not be meaningful to add/subtract/compare `lastRebufferRealtimeUs` with `playbackPositionUs`.

I'm not saying we should definitely use millis or micros, just that there's arguments for both I think.

marcbaechinger: Yeah, I think I see you point of the last buffer timestamp being taken from the `clock.elapsedRealtime()` which is MS. Looking at the arguments of `EPII.updateRebufferingState(boolean isRebuffering, boolean resetLastRebufferRealtimeMs)` and the naming of the boolean flag I'm not sure how deliberate this was. 

But looks like falling into a democratic mode is quickest to resolve this dispute which makes me loose 2:1. :)

marcbaechinger:
Thanks @colinkho for your PR!

I'm going to send this for internal review. Please ignore the comment around inline comments in the test class. I'm going to fix this myself and will push this to the PR once done.

colinkho:
Hi, any update on the PR here?

marcbaechinger:

PR #2115, opened 2025-02-07
Replace usages of `Util.areEqual()` with `Objects.equals()`

MGaetan89:
The `Util.areEqual()` method is deprecated in favor of `Objects.equals()`. So this PR migrates the remaining usages of `Util.areEqual()` to the recommended alternative.

Since Guava also provides an `Objects` class which is used in some places for equality and hash code, I've also migrated Guava's `Objects` class to Java's one.


tonihei:

PR #2136, opened 2025-02-12
Fix permitted parameter range for handleReplaceMediaItems in javadoc

bubenheimer:
Align javadoc with code


tonihei:

PR #2137, opened 2025-02-12
Fix cue render order for ass/ssa.

### LINKED ISSUE #2124: ASS render in a wrong order.
### Version

Media3 1.5.1

### More version details

1.5.1 release version.

Pixel 5

None (I think)

Yes

### Reproduction steps

Chose a video with a customer ass.

Play

### Expected result

The  red latter should render behind the white, and before the gray.

Image

### Actual result

The gray is before the white, and red before all.

Image

### Media

The sample will post to media email later.

### Bug Report

- [x] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

peerless2012:
This should fix #2124 
* Add layer in cue and output.
* Add layer support for ssa parser.

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/text/Cue.java
+++ b/libraries/common/src/main/java/androidx/media3/common/text/Cue.java
@@ -310,6 +310,7 @@ public final class Cue {
   public final float shearDegrees;
 
   /** The z index for cue, the larger index will render above the smaller index. */
+  @UnstableApi
   public final int zIndex;
 
   private Cue(
--- a/libraries/common/src/main/java/androidx/media3/common/text/Cue.java
+++ b/libraries/common/src/main/java/androidx/media3/common/text/Cue.java
@@ -310,9 +310,9 @@ public final class Cue {
   public final float shearDegrees;
 
   /**
-   * The layer for cue, the larger cue will render above the smaller cue.
+   * The z index for cue, the larger index will render above the smaller index.
    */
-  public final int layer;
+  public final int zIndex;
 
   private Cue(
       @Nullable CharSequence text,
@@ -332,7 +332,7 @@ public final class Cue {
       int windowColor,
       @VerticalType int verticalType,
       float shearDegrees,
-      int layer) {
+      int zIndex) {
     // Exactly one of text or bitmap should be set.
     if (text == null) {
       Assertions.checkNotNull(bitmap);
@@ -362,7 +362,7 @@ public final class Cue {
     this.textSize = textSize;
     this.verticalType = verticalType;
     this.shearDegrees = shearDegrees;
-    this.layer = layer;
+    this.zIndex = zIndex;
   }
 
   /** Returns a new {@link Cue.Builder} initialized with the same values as this Cue. */
@@ -399,7 +399,7 @@ public final class Cue {
         && textSize == that.textSize
         && verticalType == that.verticalType
         && shearDegrees == that.shearDegrees
-        && layer == that.layer;
+        && zIndex == that.zIndex;
   }
 
   @Override
@@ -422,7 +422,7 @@ public final class Cue {
         textSize,
         verticalType,
         shearDegrees,
-        layer);
+        zIndex);
   }
 
   /** A builder for {@link Cue} objects. */
@@ -445,7 +445,7 @@ public final class Cue {
     @ColorInt private int windowColor;
     private @VerticalType int verticalType;
     private float shearDegrees;
-    private int layer;
+    private int zIndex;
 
     public Builder() {
       text = null;
@@ -484,7 +484,7 @@ public final class Cue {
       windowColor = cue.windowColor;
       verticalType = cue.verticalType;
       shearDegrees = cue.shearDegrees;
-      layer = cue.layer;
+      zIndex = cue.zIndex;
     }
 
     /**
@@ -817,19 +817,19 @@ public final class Cue {
       return verticalType;
     }
 
-    /** Sets the layer for this Cue. */
+    /** Sets the zIndex for this Cue. */
     @CanIgnoreReturnValue
-    public Builder setLayer(int layer) {
-      this.layer = layer;
+    public Builder setZIndex(int zIndex) {
+      this.zIndex = zIndex;
       return this;
     }
 
     /**
-     * Gets the layer for this Cue.
+     * Gets the zIndex for this Cue.
      */
     @Pure
-    public @VerticalType int getLayer() {
-      return layer;
+    public int getZIndex() {
+      return zIndex;
     }
 
     /** Build the cue. */
@@ -852,7 +852,7 @@ public final class Cue {
           windowColor,
           verticalType,
           shearDegrees,
-          layer);
+          zIndex);
     }
   }
 
@@ -875,7 +875,7 @@ public final class Cue {
   private static final String FIELD_WINDOW_COLOR_SET = Util.intToStringMaxRadix(14);
   private static final String FIELD_VERTICAL_TYPE = Util.intToStringMaxRadix(15);
   private static final String FIELD_SHEAR_DEGREES = Util.intToStringMaxRadix(16);
-  private static final String FIELD_LAYER = Util.intToStringMaxRadix(19);
+  private static final String FIELD_Z_INDEX = Util.intToStringMaxRadix(19);
 
   /**
    * Returns a {@link Bundle} that can be serialized to bytes.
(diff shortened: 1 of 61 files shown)


tonihei:
@icbaker Could you take a look? Thanks!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/text/MergingCuesResolver.java
```diff
@@ -52,6 +52,12 @@
           .compound(
               Ordering.<Long>natural().reverse().onResultOf((CuesWithTiming c) -> c.durationUs));
 
+  /**
+   * An {@link Ordering} which sorts cues in ascending layer priority
+   */
+  private static final Ordering<Cue> CUES_LAYER_PRIORITY_COMPARATOR =
+      Ordering.<Integer>natural().onResultOf(c -> c.layer);
```
icbaker, line 59: I think this logic should be part of `CUES_DISPLAY_PRIORITY_COMPARATOR` rather than a separate `Ordering` object

libraries/common/src/main/java/androidx/media3/common/text/Cue.java
```diff
@@ -309,6 +309,11 @@ public final class Cue {
    */
   public final float shearDegrees;
 
+  /**
+   * The layer for cue, the larger cue will render above the smaller cue.
+   */
+  public final int layer;
```
icbaker, line 315: nit: TTML (which has a much more robust & unambiguous spec than SSA) uses the name `zIndex` for this. It seems to have the same "higher index renders above lower index" semantics.

Shall we use that nomenclature here? So `Cue.zIndex`

We should keep the `layer` terminology everywhere in the `ssa` package.

libraries/common/src/main/java/androidx/media3/common/text/Cue.java
```diff
@@ -806,6 +817,21 @@ public Builder setShearDegrees(float shearDegrees) {
...
+    public Builder setLayer(int layer) {
+      this.layer = layer;
+      return this;
+    }
+
+    /**
+     * Gets the layer for this Cue.
+     */
+    @Pure
+    public @VerticalType int getLayer() {
```
icbaker, line 831: If this isn't needed I suggest we remove it

libraries/extractor/src/main/java/androidx/media3/extractor/text/ssa/SsaParser.java
```diff
@@ -361,6 +369,19 @@ private void parseDialogueLine(
...
+  /**
+   * Parse int in SSA.
+   * @param intString The string to parse.
+   * @return The parsed int.
+   */
+  private static int parseInt(String intString) {
+    try {
+      return Integer.parseInt(intString.trim());
+    } catch (Exception exception) {
+      return C.LENGTH_UNSET;
```
icbaker, line 381: I think we should default to `0` here for `layer`, same as `Cue`

That probably means you need to either return `null` here, or take a default value as a parameter, or don't define this general purpose function and just in-line it for the layer case.

I'd suggest in-lining

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -184,7 +185,8 @@ public void draw(
         && this.parentLeft == cueBoxLeft
         && this.parentTop == cueBoxTop
         && this.parentRight == cueBoxRight
-        && this.parentBottom == cueBoxBottom) {
+        && this.parentBottom == cueBoxBottom
+        && this.cueLayer == cue.layer) {
```
icbaker, line 189: What happens if you remove this comparison? It seems a bit surprising you need to store `cueLayer` into a field and then **only** use it in this comparison. I think all the other fields being compared here are then later used to compute the layout (hence they need to be checked to see if the layout needs to be invalidated).

libraries/extractor/src/main/java/androidx/media3/extractor/text/CuesWithTiming.java
```diff
@@ -68,7 +75,7 @@ public class CuesWithTiming {
 
   /** Creates an instance. */
   public CuesWithTiming(List<Cue> cues, long startTimeUs, long durationUs) {
-    this.cues = ImmutableList.copyOf(cues);
+    this.cues = ImmutableList.sortedCopyOf(CUES_LAYER_PRIORITY_COMPARATOR, cues);
```
icbaker, line 78: `CuesWithTiming.cues` has no documented ordering behaviour wrt layout on screen.

`CuesGroup.cue` **does** document how ordering affects layout.

Is this reordering needed here? It seems it shouldn't be, since the field has no ordering guarantee. Maybe this should be moved to a point where a `CueGroup` is being created?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/text/MergingCuesResolver.java
peerless2012: CUES_DISPLAY_PRIORITY_COMPARATOR with CuesWithTiming, but the layer is in Cue, not in CuesWithTiming.

libraries/common/src/main/java/androidx/media3/common/text/Cue.java
peerless2012: I'm not familiar with TTML, but I can chenage the layer to zIndex, if you think it is more stand.

libraries/common/src/main/java/androidx/media3/common/text/Cue.java
peerless2012: > If this isn't needed I suggest we remove it

Yes, I forget to delete the type annotation.

libraries/extractor/src/main/java/androidx/media3/extractor/text/ssa/SsaParser.java
peerless2012: parseInt is a general method, and I will set layer to 0 when the result is unset in line 333.

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
peerless2012: Due to your explain, it seems no need to compair the layer.

libraries/extractor/src/main/java/androidx/media3/extractor/text/CuesWithTiming.java
peerless2012: Yes, you are right. That's a good way. and it seems there is no need to modify the MergingCuesResolver.

peerless2012:
@icbaker hello, please make a review for the new commit, and hope this pr can merge to 1.6.0 final release.

libraries/extractor/src/main/java/androidx/media3/extractor/text/ssa/SsaParser.java
icbaker: `parseInt` may be trying to be a general method, but `LENGTH_UNSET` (-1) is not a generally safe 'invalid' value for all possible integer fields (i.e. there may be some integer fields where we expect to read -1 from the file). In fact, tbh it's not clear to me that -1 (or any negative number) is an invalid value for `layer`. The SSA "spec" I'm reading says:
> Layer (any integer)

Which could reasonably be interpreted to include negative numbers.

So please either in-line this method (my preference) or return `null` when the parsing fails.

icbaker:
Please can you add some tests for this?

I suggest:
1. Some parsing tests in `SsaParserTest` (including negative & unparseable layer values) - assert that the right value ends up in `Cue.zIndex`
2. An end-to-end test in `SubtitlePlaybackTest`? This will cover both your parsing code and the re-ordering in the `CueGroup` constructor.

For (2) I suggest you craft an SSA test file that has cues overlapping in time with different layer values, to demonstrate why the re-ordering has to be done 'outside' `SsaParser`.

peerless2012:
I add `parseLayer` and `parseInvalidLayer` in `SsaParserTest` to cover the layer value in ssa.

How can I add a test in `SubtitlePlaybackTest`? I have no idea.

icbaker:
Thanks for adding the tests to `SsaParserTest`, I've added `SsaPlaybackTest` to cover the layer-based re-ordering inside `CueGroup`.

peerless2012:
> I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks!

Thanks a lot for your help. 🫶

Issue #2069, opened 2025-01-22
This commit will cause H265 ts to fail


FongMi:
### Version

Media3 main branch

### More version details

https://github.com/androidx/media/commit/029b8bad86f6ba3d734f1202e8d115b622c267c2
If I revert the modification of H265Reader, it won't get stuck.

Before revert
[problem.zip](https://github.com/user-attachments/files/18508788/problem.zip)

After revert
[no problem.zip](https://github.com/user-attachments/files/18508789/no.problem.zip)

### Devices that reproduce the issue

![Image](https://github.com/user-attachments/assets/3f32a3a9-da4d-4eda-a9b5-1390edb6cdf9)

### Devices that do not reproduce the issue

None

### Reproducible in the demo app?

Not tested

### Reproduction steps

I don't have the actual video URL
I asked users to test
Tested from media3 1.3.1 ~ 1.4.1
The problem was found to occur from 1.4.0-alpha02
And I compare 1.4.0-alpha01 and 1.4.0-alpha02
I just found a commit about H265
https://github.com/androidx/media/commit/029b8bad86f6ba3d734f1202e8d115b622c267c2

### Expected result

Don't get stuck

[no problem.zip](https://github.com/user-attachments/files/18508789/no.problem.zip)

### Actual result

get stuck

[problem.zip](https://github.com/user-attachments/files/18508788/problem.zip)

### Media
Sample 1
https://drive.google.com/file/d/1k5sfTgYWf2r1fdnFlknkfoZ0Gr9PL0i5/view?usp=sharing

Sample 2
http://otttv.bj.chinamobile.com/PLTV/88888888/224/3221226893/1.m3u8?GuardEncType=2&accountinfo=%7E%7EV2.0%7EcnoJZd_sZxCC6bZYZh4R6g%7EtP4-l0lmSfjwLWEfK_el1vH_mv-s1zo4AQJwdedaVwG9xkuFTDg8J26cwOrNJzn20BErrHdLhuZ9EzLUCD3PMW-OMx4MGteHV2vLeW6BqoY%2CEND

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

FongMi:
@dsparano Do you have any ideas?

FongMi:
OK, maybe it will be available in the morning, the users is asleep.

FongMi:
@dsparano I have edit media url.
But maybe some device won't happened, only special device will happened.

[...]

PR #2145, opened 2025-02-13
Fix issue #2069

dsparano:
Fix for issue #2069 consisting of:

- Change implementation of packetFinished() at end of input in H264 and H265 readers, using end & start dummy NALu to trigger output of the very last sample. This seems more simple and reliable than SampleReader.end() which is now removed;
  - Get rid of double output samples for H265 at end of input;
- Position NAL unit start offset correctly at beginning of start code prefix in case of a 4 byte start code 00 00 00 01, thus fixing access unit length
- Fix H264 and H265 TS extractor dumps
- Add a few .gitignore files to fix CRLF mismatch test failures in Windows


libraries/test_data/src/test/assets/.gitattributes
```diff
@@ -0,0 +1 @@
+*.dump -text
```
tonihei, line 1: We don't have such files anywhere in our code yet. As this seems to be unrelated to the actual fix, I'll remove them from the PR for now. If you think this is needed, could you open a new issue to explain the difference? Then we can look into adding these everywhere where needed and not only to the paths that happened to be touched by this change.

tonihei:

libraries/test_data/src/test/assets/.gitattributes
dsparano: I understand, will propose a separate PR.

PR #2149, opened 2025-02-16
Minor Util.postOrRun refactor: fix skipped reuse of variable created …

bubenheimer:
…for reuse

It looks like a variable got created to only obtain the Looper once (the Looper is stored in a `final` field), but then it's not actually being reused. A minor nit I tripped over during code review.


tonihei:

PR #2163, opened 2025-02-19
Fixed break presence of the next segment for hls

ArtemyukViktor03:
Hi developers
in  [PR](https://github.com/androidx/media/pull/419) discused some context about why it's correct to only do this if `mode != MODE_HLS`
The autor [PR]( https://github.com/androidx/media/pull/419) could not defend his opinion about HLS check in `TsExtractor ->  read()` method

for_pul

please approve our PR,  since our hls's work fine in `ExoPlayer` lib and after migtration to `Media3` lib our hls's work fine up to `1.1.1` version
But our hls's freeze on `1.2.0 - alpha01` version and above (was added [PR]( https://github.com/androidx/media/pull/419))
in our PR we added only HLS check in` TsExtractor ->  read()` method

Without of our check`mode != MODE_HLS` would break HLS mode, because of the presence of the next segment.

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/ts/TsExtractor.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/ts/TsExtractor.java
@@ -449,12 +449,14 @@ public final class TsExtractor implements Extractor {
 
     if (!fillBufferWithAtLeastOnePacket(input)) {
       // Send a synthesized empty pusi to allow for packetFinished to be triggered on the last unit.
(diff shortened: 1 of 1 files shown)


FongMi:
Can you provide the hls url for testing?

FongMi:
@ArtemyukViktor03 Why closed?

ArtemyukViktor03:
[FongMi](https://github.com/FongMi) hi, found a better solution and created a new one  PR because I saw that this had already been discussed [#1189](https://github.com/androidx/media/pull/1189/)

PR #2183, opened 2025-02-25
Allow CmcdConfiguration.Factory to return null to disable CMCD

MattiasBuelens:
We would like to enable or disable CMCD selectively for specific media items. This already works today by simply returning `null` from `CmcdConfiguration.Factory.createCmcdConfiguration()`, however this behavior is not specified anywhere.

This PR updates the return type to be `@Nullable`, so this behavior is properly specified and better supported for Kotlin users.


MattiasBuelens:
I'll reopen this soon from [the DolbyLaboratories fork](https://github.com/DolbyLaboratories/media) after internal review. Apologies for the confusion!

PR #2186, opened 2025-02-25
Adds support for fragmented mp4 with multiple sidx boxes

anthonybajoua:
## Summary:
Addresses [Github issue #9373](https://github.com/google/ExoPlayer/issues/9373) 
Implements solution 1 proposed by [Google maintainer](https://docs.google.com/document/d/1NUheADYlqIVVPT8Ch5UbV8DJHLDoxMoOD_L8mvU8tTM/edit?tab=t.0#heading=h.2zer3ifwclzq).

## Algorithm/Solution
1. Upon encountering multiple sidx atom in track:
2. For each atom in the track:
    1. If sidx atom then aggregate  ChunkIndex into master list, else skip
    2. Merge them all into one large chunk index.
    3. Reset the seek pointer to where it was last
Retry above on I/O failures.


colinkho:
Hey @tonihei @icbaker , would you guys be able to take a look? This would help us maintain a vanilla fMp4 extractor!

oceanjules:
@anthonybajoua Hey, a reminder to complete the CLA instruction if you want us to look at your code :)

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -509,26 +515,56 @@ public void release() {
     // Do nothing
   }
 
+  private void readRemainingSidxAtomsIntoTrack(ExtractorInput input) throws IOException {
```
rohitjoins, line 518: This method scans the entire input to collect all `sidx` boxes before playback starts, which could increase startup latency, especially for large files.

Instead, could we consider building the seek map incrementally i.e. starting playback with the first `sidx` and extending the seek map as we encounter additional `sidx` boxes during playback or buffering?

This could significantly reduce startup time while still allowing full seekability to be built up progressively.

anthonybajoua: This is an interesting idea trading usability for startup latency. If users seek to somewhere out of the buffer, what'd the behavior be? 

In local testing on an SSD, this startup latency is sub 1s for videos around an hour of length. The blast radius is only videos of this particular format.

Regarding buffering in online playback, any web server serving mp4 should really be using DASH.

anthonybajoua: That is fine by me, please let me know if there's anything at all needed on my side.

colinkho: @rohitjoins Is there an ETA for this? Just want to get this in so we are unblocked. Thank you!

cc: @tonihei @icbaker

rohitjoins: @colinkho I'm preparing the PR for internal review and should be merged this week. Thanks!

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -511,24 +526,36 @@ public void release() {
...
+          default:
+            if (readSample(input)) {
+              return RESULT_CONTINUE;
+            }
+        }
+      }
+    } finally {
+      if (seekPositionBeforeSidxProcessing != C.INDEX_UNSET) {
+        seekPositionBeforeSidxProcessing = C.INDEX_UNSET;
       }
```
rohitjoins, line 559: We're still investigating the use of `finally` block here. Extractor should be able to restart from an IOException without having this. Removing this block results in test failures with IO Errors simulation.

anthonybajoua: Would it be preferable to catch IOException, reset the `seekPositionBeforeSidxProcessing` in the catch block and rethrow the exception? That is functionally equivalent.

anthonybajoua: @rohitjoins Do you to steward the changes and take through another review pass?

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -653,19 +679,52 @@ private void processAtomEnded(long atomEndPosition) throws ParserException {
...
+        haveOutputSeekMap = true;
+      } else if ((flags & FLAG_MERGE_FRAGMENTED_SIDX) != 0
+          && !haveOutputSeekMapFromMultipleSidx
+          && chunkIndexMerger.size() > 1) {
+        try {
+          processRemainingSidxAtoms(input);
+          seekPositionBeforeSidxProcessing = inputPosition;
+          haveOutputSeekMapFromMultipleSidx = true;
+        } finally {
+          extractorOutput.seekMap(chunkIndexMerger.merge());
```
rohitjoins, line 701: @anthonybajoua Same with this finally block, why do we need to emit seekMap ( which may be incomplete ), when an IO error occurs?

rohitjoins: `haveOutputSeekMapFromMultipleSidx` was never set to `true` in this case. Due to how the `IOError` simulation was configured, we got lucky that it triggered consistently. Had it occurred intermittently (e.g., every second read), we might have ended up skipping sample data or headers.

I've refactored the code to explicitly track state and ensure the seek map is only output once all `sidx` boxes have been processed. Thanks for your patience!

anthonybajoua: Amazing cleanup! Thanks so much for working through this (:

Issue #1931, opened 2024-11-30
24 bit and 32 bit int MediaCodec and AudioTrack output


nift4:
### [REQUIRED] Use case description
Some manufacturers utilize 24 bit and 32 bit int output support in their flagship device's hardware.  Utilizing this hardware mode in MediaCodec and AudioTrack would avoid wasting memory and CPU time in converting to 32bit float values and would also avoid quality loss of converting to 16bit int.
Additionally, considering #415, this is required to output bit perfect audio if files are in 24 bit int or 32 bit int format.

### Proposed solution
In https://github.com/google/ExoPlayer/issues/6749, 32 bit float output support was added to ExoPlayer. Similar to that issue, 24 bit and 32 bit int support would solve this issue.

### Alternatives considered
Keep converting to 32 bit float or 16 bit int, wasting battery and not fully utilizing user's hardware.

PR #2218, opened 2025-03-08
Try input format PCM encoding first in MediaCodecAudioRenderer

nift4:
...and remove hard requirement for 16-bit PCM support in the audio sink even if that's not needed to playback.

If the audio file is using 24-bit or 32-bit (int) PCM, MediaCodecAudioRenderer tries to output float, and if that is not supported, falls back to 16-bit PCM. This commit changes this behavior to first trying the media file format, then any close high-resolution audio formats (such as float) and then fall back to 16-bit PCM. The behaviour with DefaultAudioSink does not change, but if an audio sink supports any more optimal formats, they will now be used with MediaCodecAudioRenderer.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/MediaCodecAudioRenderer.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/MediaCodecAudioRenderer.java
@@ -26,7 +26,6 @@ import static java.lang.Math.min;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.media.AudioDeviceInfo;
-import android.media.AudioFormat;
 import android.media.MediaCodec;
 import android.media.MediaCrypto;
 import android.media.MediaFormat;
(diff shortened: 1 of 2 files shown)


nift4:
Sadly forgot to add this to the commit message, but it's a first step towards #1931

nift4:
@microkatz sorry to bump this but it's been a while, any chance you could take a look at this PR? thanks!

nift4:
@microkatz @icbaker hi, please could you review this PR? it's been waiting for 2 months, and I'm somewhat blocked by this being reviewed/merged before upstreaming more things. thanks in advance!

nift4:
@microkatz Hi, can you please take a look at this PR, as it has been a while? Thanks in advance!

nift4:
Hi @microkatz @tonihei, is there any chance to get this PR reviewed, please?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/MediaCodecAudioRenderer.java
```diff
@@ -338,14 +337,38 @@ public String getName() {
...
+        return RendererCapabilities.create(C.FORMAT_UNSUPPORTED_SUBTYPE);
+      }
+    } else {
+      // For all other input formats, MediaCodec can do some conversions for us. Check if the sink
+      // supports any of the formats we can get out of MediaCodec.
+      boolean pcmEncodingSupported = false;
+      // On API levels before 24, any non-PCM input format is decoded to 16-bit PCM.
+      if (SDK_INT >= 24) {
+        int[] platformEncodings = Util.getClosestPlatformPcmEncodings(format.pcmEncoding);
+        for (int platformEncoding : platformEncodings) {
```
microkatz: Can you explain the purpose of this code? 

This method just checks format support by the audio sink and its not necessary to check other platform encodings at the moment.

Basically at default, MediaCodec should support 16-bit PCM at minimum. The point of this method is to check the track support in general. That is why the original check just evaluates the 16-bit PCM.

Do you have a use-case where the audio sink is distinctly not support 16-bit PCM and only supporting higher-fidelity audio depth?

microkatz:
@nift4 

Thank you for your patience. Can you describe in greater detail what your goal is?

To note, the supportsFormat method in MediaCodecAudioRenderer method just checks format support by the audio sink. At default 16-bit PCM is the default and therefore that should all be needed to check baseline track support. Do you have a use-case where the audio sink is distinctly not support 16-bit PCM nor FLOAT(which is preferred to 24)? 

Also as it stands, I don't think your code change will work as expected. If you checked further through to AudioTrackAudioOutputProvider.java then you would see that your PR needs additional code to check other encoding support.

nift4:
> Can you describe in greater detail what your goal is?

If file is 16-bit, I want to output 16 bit and if file is 24-bit, I want to output 24 bit. The rationale is that I want audio sink to get unconverted data if possible, not converted by MediaCodec. It's also not a problem for performance because AudioTrack can do all conversions (like 24->16 or 24->float or whatever the device uses) natively in platform code. I'm trying to avoid additional effort of going 24 bit (file)->float (MediaCodec output)->24 bit again if file is 24bit in the first place.

> Do you have a use-case where the audio sink is distinctly not support 16-bit PCM nor FLOAT(which is preferred to 24)?

No, but I have use-case where FLOAT is not supported. Some Qualcomm devices have unofficial API for PCM offload which supports only 16 and 24 bit. 

> Also as it stands, I don't think your code change will work as expected. If you checked further through to AudioTrackAudioOutputProvider.java then you would see that your PR needs additional code to check other encoding support.

I have a custom audio sink in my app currently, which is patched for supporting this.
The relevant code change for upstreaming is #2445. As you can see it currently conflicts because I did not get to rebasing (I'll do it soon), but it should outline the general idea and it has the code which you identified as missing here.
I chose to make two seperate PRs because they work independently (although there is no benefit unless you make a custom audio sink with only this PR, when both are merged, the benefit will appear). As the other PR seems to be blocked by some internal draft it would be nice if maybe this one can get merged first, before the other one is done.

nift4:
Ah now I see what you were thinking about,

> To note, the supportsFormat method in MediaCodecAudioRenderer method just checks format support by the audio sink. At default 16-bit PCM is the default and therefore that should all be needed to check baseline track support. Do you have a use-case where the audio sink is distinctly not support 16-bit PCM nor FLOAT(which is preferred to 24)?

I do not have a usecase where the explicit check is absolutely needed, I just thought it would be nice to not hardcode assumptions in the code.

Issue #1932, opened 2024-11-30
call .cloneAndRemove().cloneAndInsert() on ShuffleOrder for moveMediaItems() in ExoPlayer


nift4:
### [REQUIRED] Use case description
I wrote a custom ShuffleOrder class with four major goals:
1. when enabling shuffle in one specific item, it will be the first item in shuffled playlist so everything else plays afterwards
2. when user presses "play next", it will play next even if shuffle is enabled (inserted items are shuffled among themselves and inserted after previous song from normal playlist in shuffle playlist)
3. support saving and loading from disk
This mostly works fine with more or less hacky approaches for each of the desired features.
However, my custom shuffle order has no chance of reacting to moveMediaItems(): unlike set/add/remove methods, moveMediaItems does not clone and edit the shuffle order. Due to that, moving a playlist item from UI while shuffle is enabled acts unpredictably (insertion points of DefaultShuffleOrder is random) while my implementation acts as expected by users (insertion point is after the previous item in shuffled list).

### Proposed solution
Call shuffleOrder.cloneAndRemove().cloneAndInsert() inside ExoPlayerImpl when moveMediaItems() is called to inform the shuffle order of changes. This allows custom shuffle order implementations to react however they like while not impacting DefaultShuffleOrder (it's still random, just a bit differently random).

### Alternatives considered
ForwardingPlayer that, when moveMediaItems() is called, extracts the current copy of the shuffle order using reflection, calling the clone methods and then setShuffleOrder() manually. This is fragile and causes ExoPlayer to reload timeline. However, it works as intended.

Issue #1381, opened 2024-05-17
ExoPlayer.moveMediaItems does not update the shuffle order (Copy)


OxygenCobalt:
### Version

Media3 main branch

### More version details

(This is a reposted issue since the last one was locked for inactivity. I've only recently realized that this issue was not my fault.)

Media3 1.3.1

### Devices that reproduce the issue

All

### Devices that do not reproduce the issue

N/A

### Reproducible in the demo app?

Yes

### Reproduction steps

Occurs at any queue move in a shuffled playlist that does not happen to result in the indended shuffled playlist (It's hard to define the specific criteria, sorry). For example, this pseudo-kotlin code:

```
player.shuffleModeEnabled = true
player.setMediaItems(listOf(dummy1, dummy2, dummy3)) // Assume these are initialized MediaItems
player.moveMediaItems(0, 2) 
```

### Expected result

Assume that the shuffled playlist is: [dummy1, dummy3, dummy2] (i.e shuffled index list of [0, 2, 1]), and the ordered playlist [dummy1, dummy2, dummy3].

Moving the media items as specified in the reproduction steps should as a result:

    Move the ordered playlist, resulting in [dummy3, dummy2, dummy1].
    Reindex the shuffle order, resulting in a shuffled index list of [2, 0, 1].
    Move the items in the shuffle order, resulting in a shuffled index list of [1, 0, 2] and a shuffled playlist of [dummy2, dummy3, dummy1]

In practice, this should involve another ShuffleOrder override, something like cloneAndMove, that performs this reindexing and item moving.

### Actual result

The ShuffleOrder is not even cloned when MediaItems are moved judging by [ExoPlayerImpl](https://github.com/androidx/media/blob/f6fe90f30ba022c7e04e14a3dc5c28c568a4d1e2/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/ExoPlayerImpl.java#L706). Instead, what happens 
- Move the ordered playlist, resulting in [dummy3, dummy2, dummy1].
- Do not reindex the shuffle order, resulting in still a shuffled index list of [0, 2, 1]
- Do not move anything in the shuffle order, resulting in a shuffled index list of [0, 2, 1] and a shuffled playlist of [dummy3, dummy1, dummy2]. This is entirely incorrect.

### Media

N/A

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

oceanjules:
[...]

PR #2226, opened 2025-03-11
notify ShuffleOrder when media item(s) are moved

nift4:
Issue: #1932
Issue: #1381

--

(Extracting a seperate method `moveMediaSourceHolders` was done to align with insert/remove code. A new `default` method instead of just implementing that in `DefaultShuffleOrder` was done because I believe the move detection is not required by simpler shuffle order implementations, and this makes upgrades easier as devs don't have to implement a new function.)

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ShuffleOrder.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ShuffleOrder.java
@@ -267,7 +267,10 @@ public interface ShuffleOrder {
   ShuffleOrder cloneAndInsert(int insertionIndex, int insertionCount);
 
   /**
-   * Returns a copy of the shuffle order with a range of elements moved.
+   * Returns a copy of the shuffle order to be used after a move.
+   *
+   * <p>The default implementation is a no-op. Custom implementation can choose to re-shuffle when
+   * items are moved.
    *
    * @param indexFrom The starting index in the unshuffled order of the range to move, from before
    *     the move occurs.
@@ -278,8 +281,7 @@ public interface ShuffleOrder {
    * @return A copy of this {@link ShuffleOrder} with the elements moved.
    */
   default ShuffleOrder cloneAndMove(int indexFrom, int indexToExclusive, int newIndexFrom) {
-    return cloneAndRemove(indexFrom, indexToExclusive)
-        .cloneAndInsert(newIndexFrom, indexToExclusive - indexFrom);
+    return this;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


nift4:
@rohitjoins @icbaker any chance one of you could take a look?

nift4:
@marcbaechinger ping since it has been a month, any chance you could take a look at this PR?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ShuffleOrder.java
```diff
@@ -266,6 +266,22 @@ public ShuffleOrder cloneAndClear() {
...
+   *
+   * @param indexFrom The starting index in the unshuffled order of the range to move, from before
+   *     the move occurs.
+   * @param indexToExclusive The smallest index (must be greater or equal to {@code indexFrom}) that
+   *     is not included in the range of elements moved, from before the move occurs.
+   * @param newIndexFrom The starting index in the unshuffled order of the range to move, from after
+   *     the move occurs.
+   * @return A copy of this {@link ShuffleOrder} with the elements moved.
+   */
+  default ShuffleOrder cloneAndMove(int indexFrom, int indexToExclusive, int newIndexFrom) {
```
marcbaechinger, line 280: If we implement this method like this, then I think there isn't much benefit, as calling the two other methods would be equivalent and we don't have to introduce this new method.

If we would implement this method standalone without using the other methods, you could raise a point that we can do it in `DefaultShuffleOrder` with a single implementation that probably has to iterate only once over the array. Do you think there is sufficient benefit in adding this new API and then implementing it here in the suggested way?

However we implement this. If we accept the change we'd need to have a unit test I think. I'd add that myself also, so this is not a condition to take that change if we think we have enough benefit because we can do it with better performance than the higher-order composite calls.

marcbaechinger (commented):
Thanks for your PR

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ShuffleOrder.java
nift4: Hi @marcbaechinger,

the benefit of adding a seperate API for this is that shuffle order may decide to not reshuffle in the individual items when a group of items is moved.

My goal with my custom ShuffleOrder is to make add/remove/move operations on the unshuffled ExoPlayer playlist map directly to shuffled playlist. My custom ShuffleOrder will reshuffle the added item group individually for insert action, but I do not want to shuffle them again when moving.

When using the existing API, ShuffleOrder can't know if this is a remove+insert done by the user, or a real move operation.

I'm not sure about the performance, but people who don't care about knowing when moving happens wouldn't need to implement any additional code if there is a default implementation. And of course, if the performance is increased when combining the call, DefaultShuffleOrder can still override cloneAndMove without the default implementation being removed. But personally, I do not feel strongly about the default implementation and it could be removed - but a seperate "move" API is needed for my usecase.

marcbaechinger:

nift4:
Sounds good to me, thanks :)

nift4:
Thanks, Marc :)

PR #2259, opened 2025-03-23
expand format support for audio processors used by DefaultAudioSink

nift4:
This allows future extensions such as direct 24 bit / 32 bit (int) playback in DefaultAudioSink, and is a step towards Issue: #1931

Run audio processors possibly discarding data first to avoid wasting CPU on data that'll be discarded anyway.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java
@@ -540,6 +540,7 @@ public final class DefaultAudioSink implements AudioSink {
   private final TrimmingAudioProcessor trimmingAudioProcessor;
   private final ImmutableList<AudioProcessor> toIntPcmAvailableAudioProcessors;
   private final ImmutableList<AudioProcessor> toFloatPcmAvailableAudioProcessors;
+  private final ImmutableList<AudioProcessor> forPreProcessingAvailableAudioProcessors;
   private final AudioTrackPositionTracker audioTrackPositionTracker;
   private final ArrayDeque<MediaPositionParameters> mediaPositionParametersCheckpoints;
   private final boolean preferAudioTrackPlaybackParams;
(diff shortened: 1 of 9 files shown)


tonihei:
@ivanbuper Could you take a look at this extension?

nift4:
@ivanbuper ping since it has been a month, any chance you could take a look at this PR?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java
```diff
@@ -702,6 +701,7 @@ public void configure(Format inputFormat, int specifiedBufferSize, @Nullable int
       inputPcmFrameSize = Util.getPcmFrameSize(inputFormat.pcmEncoding, inputFormat.channelCount);
 
       ImmutableList.Builder<AudioProcessor> pipelineProcessors = new ImmutableList.Builder<>();
+      pipelineProcessors.addAll(forPreProcessingAvailableAudioProcessors);
```
ivanbuper, line 709: Can you please explain what is the goal of `forPreprocessingAvailableAudioProcessors`? In the current state of the PR, you might add the same processor twice to the pipeline. Is this intentional?

nift4: Hi @ivanbuper,

> Can you please explain what is the goal of forPreprocessingAvailableAudioProcessors?

Right now, some audio processors are added based on if it's float or int output. But Trimming and ChannelMapping audio processor work with any PCM format after this PR, including int8, int16, int24, int32, float32. So they can always be added without problems if we enter PCM codepath, even if (which I want to PR later) a format which is neither int16 nor float32 is used.

> In the current state of the PR, you might add the same processor twice to the pipeline. Is this intentional?

No, the DefaultAudioSink edit is not supposed to change anything and should only prepare for future edit. I can't follow why it'd be added twice (I've removed them from `toFloatPcmAvailableAudioProcessors` and `toIntPcmAvailableAudioProcessors` both - either of one was always added - and added them to `forPreprocessingAvailableAudioProcessors` which now also is always added).

ivanbuper: Thanks, that makes sense. I missed that change to `toIntPcmAvailableAudioProcessors` and `toFloatPcmAvailableAudioProcessors`.

ivanbuper:
I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback (e.g. formatting). Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks!

nift4:
Thanks :)

PR #2260, opened 2025-03-24
Use the existing `TAG` constant in `AudioManagerCompat`

MGaetan89:
https://github.com/androidx/media/blob/d5dcbf4a12bddbcf26e02fdfb29105f5c288ef5d/libraries/common/src/main/java/androidx/media3/common/audio/AudioManagerCompat.java#L43

This also addresses the `Private field 'TAG' is never used` warning.


microkatz:

PR #2285, opened 2025-03-28
Deprecate `Util.SDK_INT` in favor of `Build.VERSION.SDK_INT`

MGaetan89:
`Util.SDK_INT` was introduced to be able to simulate any SDK version during tests. This is possible by using Robolectric's `@Config(sdk)` annotation.
All usages of `Util.SDK_INT` have been replaced by `Build.VERSION.SDK_INT`.

This is a similar change to what was done in #2107.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -3,6 +3,7 @@
 ### Unreleased changes
 
 *   Common Library:
+    *   Deprecate `Util.SDK_INT`. Use `Build.VERSION.SDK_INT` instead.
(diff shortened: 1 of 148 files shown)


tonihei:
Thanks for trying to replace all these usages :) 

The value was in fact not only used for Robolectric, but also to allow special logic for pre-releases of Android that don't have a real version number yet. We haven't used this field for that purpose in years though, and if we wanted to, we could just a more localized check while the pre-release hasn't got its new number.

PR #2330, opened 2025-04-11
Fix StrictMode unsafe intent launch violation

nift4:


tonihei:
Could you provide more details on what the issue is? I don't see any warnings/lint issues in this area in Android Studio, but it sounds like you've seen a warning using some StrictMode.

nift4:
Hi @tonihei ,

in **debug** builds of my app I enabled the built-in platform feature StrictMode with the VmPolicy set to (among others) detect unsafe intent launches:
https://developer.android.com/reference/android/os/StrictMode.VmPolicy.Builder#detectUnsafeIntentLaunch()
with penaltyDeath (the app process will be terminated when an issue occurs). This helps catch programming mistakes.

One such issue detected by StrictMode is that when starting playback with a wired headset's play button using media resumption, the MediaButtonRetriever takes the recieved Intent, modifies it and sends it again. According to the docs, this is a possible security problem:
https://developer.android.com/reference/android/os/strictmode/UnsafeIntentLaunchViolation
However, in this case, because setComponent() is always set to our own app's component, there is no security impact at all. I still believe it's good to fix this, because it's technically not a false positive (the Intent with which the service is started does originate from another app).

And fixing this will allow me to remove the exemption of this detection from penaltyDeath and hence detect other programming mistakes that might be a real issue.

nift4:
Hi @tonihei, do you need any further information to proceed here?

nift4:
@tonihei sorry for the ping, could you please take a look?

tonihei:
Sorry for the delay! I checked the pull request and it seems to work well to prevent the warning but also doesn't seem to remove any functionality from the incoming Intent. I'm going to import the PR now.

nift4:
Thank you!

PR #2367, opened 2025-04-23
Updated Composition Demo app

SigmanZero:
This PR rewrites the Composition Demo app from the ground up, now using Compose and modern app architecture best practices. We've also added preset layouts to arrange sequences in a 2x2 or PiP composition, in addition to the sequential compositions that the demo app already showcased.


droid-girl:
Already merged. Closing this PR

PR #2386, opened 2025-04-30
Allow `CmcdConfiguration.Factory` to return `null` to disable CMCD

MattiasBuelens:
We would like to enable or disable CMCD selectively for specific `MediaItem`s, while still using the same `MediaSource.Factory` instance. This already works today by returning `null` from `CmcdConfiguration.Factory.createCmcdConfiguration()`, however this behavior is not specified anywhere.

This PR updates the return type of `createCmcdConfiguration()` to be `@Nullable`, so this existing behavior is properly specified and better supported for Kotlin users.

We've considered a few alternative solutions, but we ultimately think allowing `null` is the most straightforward solution:
* Return a "dummy" `CmcdConfiguration` that doesn't report any CMCD keys. This is possible by passing a `requestConfig` to the `CmcdConfiguration` constructor whose `isKeyAllowed()` method always returns `false`. However, it's not very efficient to have ExoPlayer do all of the work of setting up a `CmcdData.Factory` and creating `CmcdData` objects and then to never actually use them.
* Make `MediaSource.Factory.setCmcdConfigurationFactory()` accept `null`, so you can revert it back to its initial state (with no CMCD configuration factory). However, none of the other methods on `MediaSource.Factory` accept `null`, so this would feel a bit out of place.
* Avoid using `MediaSource.Factory` at all, and instead manually create `MediaSource`s. We would then need to call `ExoPlayer.setMediaSource()` instead of `ExoPlayer.setMediaItem()`. This would work, but it would be unfortunate that we could no longer use `setMediaItem()` for our use case.

Supersedes #2183.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/CmcdConfiguration.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/upstream/CmcdConfiguration.java
@@ -127,8 +127,9 @@ public final class CmcdConfiguration {
      * Creates a {@link CmcdConfiguration} based on the provided {@link MediaItem}.
      *
      * @param mediaItem The {@link MediaItem} from which to create the CMCD configuration.
-     * @return A {@link CmcdConfiguration} instance.
+     * @return A {@link CmcdConfiguration} instance, or {@code null} to disable CMCD.
      */
+    @Nullable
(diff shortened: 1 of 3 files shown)


rohitjoins:
Thanks for the contribution. I'm going to start the internal review process.

PR #2390, opened 2025-05-02
Enable multiple audio presentations parsing for AC-4 in MP4

sqshet-dlb:
* Parse multiple audio presentations for AC-4 content
* Populate audio presentations in a Format object associated with the audio track. Also allow to convert multiple audio presentations to Bundle using a unique key associated with Parcelable AudioPresentation objects.
* Add unit test to evaluate the multiple audio presentations parsing

Test: Run libraries/extractor/src/test/java/androidx/media3/extractor/
      mp4/Mp4ExtractorNonParameterizedTest.java unit test.


sqshet-dlb:
Create the pull request on the main brach instead.

PR #2399, opened 2025-05-05
Add missing commands to `MediaIntToString#playerCommandMap`

MGaetan89:
The following commands were missing from `MediaIntToString#playerCommandMap` in `:testapps`:
- `COMMAND_SET_DEVICE_VOLUME_WITH_FLAGS`
- `COMMAND_ADJUST_DEVICE_VOLUME_WITH_FLAGS`
- `COMMAND_SET_AUDIO_ATTRIBUTES`
- `COMMAND_SET_TRACK_SELECTION_PARAMETERS`
- `COMMAND_GET_TRACKS`
- `COMMAND_RELEASE`

Also, the `COMMAND_SEEK_TO_DEFAULT_POSITION` command was listed twice.


icbaker:
I've just written a test to enforce this going forward - and it made clear that since `COMMAND_ADJUST_DEVICE_VOLUME` and `COMMAND_ADJUST_DEVICE_VOLUME_WITH_FLAGS` have the same value (one is deprecated), we shouldn't try and add them both to the map. Same for `COMMAND_SET_DEVICE_VOLUME`. I'll add a commit to this PR removing the deprecated ones.

icbaker:
Ah right, good point, I saw the case of the deprecated fields with duplicate values and didn't look closely enough to see that didn't apply to all the deprecated values - thanks for pointing that out, I've removed my commit from this PR and finessed my test :)

PR #2413, opened 2025-05-12
RTSP H265 Aggregation packet support

ArtsemKurantsou:
Aggregation packet support for H265 RTSP added according to the specification.
The implementation is based on the assumption that DONL won't be present in the packet since `sprop-max-don-diff != 0` is not supported

### DIFF
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpH265Reader.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpH265Reader.java
@@ -102,8 +102,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
     if (payloadType >= 0 && payloadType < RTP_PACKET_TYPE_AP) {
       processSingleNalUnitPacket(data);
     } else if (payloadType == RTP_PACKET_TYPE_AP) {
-      // TODO: Support AggregationPacket mode.
-      throw new UnsupportedOperationException("need to implement processAggregationPacket");
+      processAggregationPacket(data);
     } else if (payloadType == RTP_PACKET_TYPE_FU) {
       processFragmentationUnitPacket(data, sequenceNumber);
     } else {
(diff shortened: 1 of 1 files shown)


ArtsemKurantsou:
Hi @tonihei, could you please help with review and moving forward with the change?

ArtsemKurantsou:
Hi @microkatz thank you for review and comment, could you please review again?

microkatz:

microkatz:
@ArtsemKurantsou 

Would you be able to add a test for this? I believe that you would create a file and test similar to the one in [RtpH263ReaderTest](https://github.com/androidx/media/blob/release/libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH263ReaderTest.java)

Thank you!

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpH265Reader.java
```diff
@@ -167,6 +166,58 @@ private void processSingleNalUnitPacket(ParsableByteArray data) {
...
+    //   |                               +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    //   |                               :
+    //   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
+    // Since sprop-max-don-diff != 0 is not supported, DONL won't present in the packet.
+
+    data.setPosition(2); // skipping payload header (2 bytes)
+    do {
+      short nalUnitSize = data.readShort(); // 2 bytes of NAL unit size
+      int nalHeaderType = (data.getData()[data.getPosition()] >> 1) & 0x3F;
```
microkatz, line 208: I believe the data.getData()[data.getPosition()] would be data.peekUnsignedByte().

Or the whole line as NalUnitUtil.getH265NalUnitType(data.getData(), -3).

ArtsemKurantsou:
> @ArtsemKurantsou
> 
> Would you be able to add a test for this? I believe that you would create a file and test similar to the one in [RtpH263ReaderTest](https://github.com/androidx/media/blob/release/libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH263ReaderTest.java).
> 
> Thank you!

Yes, sure, I can add tests for that

libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/reader/RtpH265Reader.java
```diff
@@ -167,6 +166,58 @@ private void processSingleNalUnitPacket(ParsableByteArray data) {
...
+      if (data.bytesLeft() < nalUnitSize) {
+        throw ParserException.createForMalformedManifest(
+            "Malformed Aggregation Packet. NAL unit size exceeds packet size.", /* cause= */ null);
+      }
+
+      fragmentedSampleSizeBytes += writeStartCode();
+      trackOutput.sampleData(data, nalUnitSize);
+      fragmentedSampleSizeBytes += nalUnitSize;
+      bufferFlags = getBufferFlagsFromNalType(nalHeaderType);
+    } while (data.bytesLeft() > 0);
```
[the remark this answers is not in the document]
ArtsemKurantsou: That is correct, however in this case we may miss "invalid" packet with 1-2 bytes "extra"

ArtsemKurantsou:
Hi @microkatz I added unit tests, could you please take a look?

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH265ReaderTest.java
```diff
@@ -0,0 +1,208 @@
...
+  ) {
+    return new RtpPacket.Builder()
+        .setTimestamp(AP_PACKET_TIMESTAMP)
+        .setSequenceNumber(AP_PACKET_SEQUENCE_NUMBER)
+        .setMarker(true)
+        .setPayloadData(Bytes.concat(AP_NALU_HEADER, Bytes.concat(nalUnits)))
+        .build();
+  }
+
+  private static void consumeFirstPacket(
```
microkatz: All your tests only consume one packet? Is this static method needed? 

If you can please just move these lines to each test.

Side note: Maybe it would be beneficial to create a test with multiple packets similar to RtpH263ReaderTest::consume_outOfOrderPackets.

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH265ReaderTest.java
```diff
@@ -0,0 +1,208 @@
...
+  private static final RtpPayloadFormat H265_FORMAT =
+      new RtpPayloadFormat(
+          new Format.Builder()
+              .setSampleMimeType(MimeTypes.VIDEO_H265)
+              .setWidth(1920)
+              .setHeight(1080)
+              .build(),
+          /* rtpPayloadType= */ 98,
+          /* clockRate= */ (int) MEDIA_CLOCK_FREQUENCY,
+          /* fmtpParameters= */ ImmutableMap.of(),
```
microkatz: Currently the RtpH265Reader does not handle reading the decoder initialization data from the stream. It requires the addition of the fmtp parameters in the rtp transport layer.

If you copied these example packets from a stream, would you be so kind as to provide the fmtp parameter data from that stream into this format data as well?

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH265ReaderTest.java
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  @Test(expected = ParserException.class)
+  public void consume_invalidAggregationPacket_extraByte() throws ParserException {
+    RtpH265Reader h265Reader = new RtpH265Reader(H265_FORMAT);
+    h265Reader.createTracks(extractorOutput, /* trackId= */ 0);
+    consumeFirstPacket(h265Reader, INVALID_AP_PACKET_EXTRA_BYTE);
+  }
+
+  @Test(expected = ParserException.class)
```
microkatz: Rather than utilize the (expected = thrownexception), the style is to use assertThrows with the specific line intending to provide the error. Would you update these unit tests using the following as an example?

    assertThrows(ParserException.class, () -> consumeFirstPacket(h265Reader, INVALID_AP_PACKET_EXTRA_BYTE));

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH265ReaderTest.java
```diff
@@ -0,0 +1,208 @@
...
+    rtpPacket = copyPacket(rtpPacket);
+    h265Reader.consume(
+        new ParsableByteArray(rtpPacket.payloadData),
+        rtpPacket.timestamp,
+        rtpPacket.sequenceNumber,
+        rtpPacket.marker
+    );
+  }
+
+  private static RtpPacket copyPacket(RtpPacket packet) {
```
microkatz: Currently your tests are not re-using RtpPackets across multiple unit tests so I'm not sure you need to run copyPacket? Your tests don't seem to fail if copyPacket is not executed.

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH265ReaderTest.java
ArtsemKurantsou: I think it make sense to keep this method to not repeat `h265Reader.onReceivingFirstPacket` in every test as it seems to be an expected call before first RTP packet is consumed. 
I also added test with multiple packets

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH265ReaderTest.java
ArtsemKurantsou: The payloads are just dummy data, but I've tested on the real stream and added parameters and format from the real stream.

libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/reader/RtpH265ReaderTest.java
ArtsemKurantsou: yeah, removed this, just added based on the the H263 implementation, but doesn't seems to be required since packet is not being modified anywhere.

Issue #1691, opened 2024-09-04
Exoplayer 2 Crashing when used with media3 dependency.


ShivamGoyal1899:
### Version

Media3 1.4.1

### Devices that reproduce the issue

- Pixel 7 Pro running Android 13

### Reproducible in the demo app?

Not tested

### Reproduction steps

- Add both media3 and exoplayer2 dependency in the project.
- Implement Separate Players for both dependencies.

### Expected result

Both Videos should play separately.

### Actual result

Media3 Player works but Exoplayer 2 crashes w/ below log.

```kotlin
15:16:22.925 AndroidRuntime           E  FATAL EXCEPTION: main
                                         Process: com.example.app, PID: 20460
                                         android.view.InflateException: Binary XML file line #116 in com.example.app:layout/video_player_widget_layout: Binary XML file line #116 in com.example.app:layout/video_player_widget_layout: Error inflating class com.google.android.exoplayer2.ui.PlayerView
                                         Caused by: android.view.InflateException: Binary XML file line #116 in com.example.app:layout/video_player_widget_layout: Error inflating class com.google.android.exoplayer2.ui.PlayerView
                                         Caused by: java.lang.reflect.InvocationTargetException
                                         	at java.lang.reflect.Constructor.newInstance0(Native Method)
                                         	at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
                                         	at android.view.LayoutInflater.createView(LayoutInflater.java:858)
                                         	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:1010)
                                         	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:965)
                                         	at android.view.LayoutInflater.rInflate(LayoutInflater.java:1127)
                                         	at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:1088)
                                         	at android.view.LayoutInflater.rInflate(LayoutInflater.java:1130)
                                         	at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:1088)
                                         	at android.view.LayoutInflater.inflate(LayoutInflater.java:686)
                                         	at android.view.LayoutInflater.inflate(LayoutInflater.java:538)
                                         	at androidx.databinding.DataBindingUtil.inflate(DataBindingUtil.java:126)
                                         	at androidx.databinding.DataBindingUtil.inflate(DataBindingUtil.java:95)
[...]

Issue #95, opened 2022-06-27
Crash when using both media3 and exoplayer2


phongbm:
My project use both media3 and exoplayer2. I use exoplayer2 before, and now I want to use media3 (keep exoplayer2 because I don't want to refactor project). When I run app, it crash with log:
```
 Caused by: java.lang.ClassCastException: com.google.android.exoplayer2.ui.AspectRatioFrameLayout cannot be cast to androidx.media3.ui.AspectRatioFrameLayout
```

![Screenshot 2022-06-27 111534](https://user-images.githubusercontent.com/7350399/175858946-a89f8936-2dbf-412b-94b0-21ec0b19dd39.png)

How do fix that?
Thanks.

lemontmoon:
you cannot use exoplayer2 libraries together with media3 player libraries,, I had the same problem. It looks like theese two libraries and not compatible together.

marcbaechinger:
I'd recommend migrating to `androidx.media3`. Using both libraries in the same project is not something we provide support for. I can't see any benefit in using both libraries and saving time with refactoring seems like this will bite you sooner or later. The stack trace above tells us it's is rather sooner than later. 

I'd take the time you invest in writing questions to forums for refactoring your code which is a sustainable investment into your project compared to investing time in moving your project into difficult waters with using both libs at a time.

I leave the issue open for a while in case someone else wants to support you.

PR #2430, opened 2025-05-15
rename resources to not conflict with exoplayer 2

j-roskopf:
media3 and exoplayer2 libraries ship a layout with the same name

https://github.com/androidx/media/blob/839c4a90f2ab36e48be73e1b5e907f3283dce72e/libraries/ui/src/main/res/layout/exo_player_view.xml

https://github.com/google/ExoPlayer/blob/dd430f7053a1a3958deea3ead6a0565150c06bfc/library/ui/src/main/res/layout/exo_player_view.xml

This results in a crash when integrating both exoplayer2 and media3 into the same app:

https://github.com/androidx/media/issues/1691
https://github.com/androidx/media/issues/95


microkatz:
Hi @j-roskopf 

It's not a supported configuration from the project perspective to have the "same" library with different package names in the same project/app. There shouldn't be any reason to depend on both ExoPlayer2 and Media3 and I'd recommend fully migrating to 'androidx.media3'.

Code that is using the old package names needs to be [migrated](https://developer.android.com/media/media3/exoplayer/migration-guide).

PR #2445, opened 2025-05-19
Better support for PCM audio formats around DefaultAudioSink

nift4:
This change:
- abstracts away the fact that SonicAudioProcessor does not support float PCM from the audio sink to the audio processor chain (easier to replace)
- makes the audio sink allow custom audio processors with any output format
- adds a new mode to DefaultAudioSink that allows any supported linear PCM format to be output into the AudioTrack

Issue: https://github.com/google/ExoPlayer/issues/10516
Issue: https://github.com/androidx/media/issues/1931

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/audio/AudioProcessorChain.java
+++ b/libraries/common/src/main/java/androidx/media3/common/audio/AudioProcessorChain.java
@@ -15,6 +15,7 @@
  */
 package androidx.media3.common.audio;
 
+import androidx.media3.common.Format;
 import androidx.media3.common.PlaybackParameters;
 import androidx.media3.common.util.UnstableApi;
 
@@ -33,7 +34,7 @@ public interface AudioProcessorChain {
    * during initialization, but audio processors may change state to become active/inactive during
    * playback.
    */
-  AudioProcessor[] getAudioProcessors();
+  AudioProcessor[] getAudioProcessors(Format inputFormat);
 
   /**
    * Configures audio processors to apply the specified playback parameters immediately, returning
@@ -50,7 +51,7 @@ public interface AudioProcessorChain {
    * value. Only called when processors have no input pending.
    *
    * @param skipSilenceEnabled Whether silences should be skipped in the audio stream.
-   * @return The new value.
+   * @return The value that was actually applied.
    */
   boolean applySkipSilenceEnabled(boolean skipSilenceEnabled);
 
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/DefaultRenderersFactory.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/DefaultRenderersFactory.java
@@ -120,7 +120,7 @@ public class DefaultRenderersFactory implements RenderersFactory {
   private long allowedVideoJoiningTimeMs;
   private boolean enableDecoderFallback;
   private MediaCodecSelector mediaCodecSelector;
-  private boolean enableFloatOutput;
+  private boolean enableHighResolutionPcmOutput;
   private boolean enableAudioOutputPlaybackParameters;
   private boolean enableMediaCodecVideoRendererPrewarming;
   private boolean parseAv1SampleDependencies;
@@ -231,19 +231,19 @@ public class DefaultRenderersFactory implements RenderersFactory {
   }
 
   /**
-   * Sets whether floating point audio should be output when possible.
+   * Sets whether to enable high resolution PCM output with more than 16 bits.
    *
-   * <p>Enabling floating point output disables audio processing, but may allow for higher quality
-   * audio output.
+   * <p>Parts of the default audio processing chain (for example, speed adjustment) will not be
+   * available when output formats other than 16-bit integer are in use.
    *
    * <p>The default value is {@code false}.
    *
-   * @param enableFloatOutput Whether to enable use of floating point audio output, if available.
    * @return This factory, for convenience.
    */
   @CanIgnoreReturnValue
-  public final DefaultRenderersFactory setEnableAudioFloatOutput(boolean enableFloatOutput) {
-    this.enableFloatOutput = enableFloatOutput;
+  public final DefaultRenderersFactory setEnableHighResolutionPcmOutput(
+      boolean enableHighResolutionPcmOutput) {
+    this.enableHighResolutionPcmOutput = enableHighResolutionPcmOutput;
     return this;
   }
 
@@ -436,7 +436,7 @@ public class DefaultRenderersFactory implements RenderersFactory {
         renderersList);
     @Nullable
     AudioSink audioSink =
-        buildAudioSink(context, enableFloatOutput, enableAudioOutputPlaybackParameters);
+        buildAudioSink(context, enableHighResolutionPcmOutput, enableAudioOutputPlaybackParameters);
     if (audioSink != null) {
       buildAudioRenderers(
           context,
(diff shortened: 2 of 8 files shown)


ivanbuper (changes requested):
Hi @nift4, thanks for your PR. From what I can see, the PR seems to be doing a few different things:

1. Modifying `DefaultAudioSink` to allow an arbitrary PCM encoding (including changes in `DefaultRenderersFactory)`.
2. Adding support in the codebase for `PCM_DOUBLE` and `DSD`.

These two items seem to be independent changes. Can you please split them out onto separate PRs to ease the review process? Thanks!

nift4:
Hi @ivanbuper, no problem, I have split the PR into this one and #2500. Thanks in advance for the review :)

ivanbuper:
@microkatz Would you mind taking a look at this one? I can do #2500. Thanks!

tonihei:
@microkatz (or whoever will look into this one): there is also the internal draft cl/592161260 that is roughly the same as this PR, so we should check where they differ and what makes most sense for the API. 

@nift4: Some question based on a quick look:
 - Our internal draft so far just had `setEnableHighResolutionPcmOutput` instead of `setEnableFloatOutput`. Are the additional options provided by this PR needed for your integration or just meant to provide all the options? 
 - The internal draft also specifically checks whether the output path actually supports the PCM version directly (which is different from the version in this PR that always claims all PCM versions are supported on newer API versions). Is this to enable specific audio processors with a different PCM version? Otherwise, if the actual output path doesn't support a certain PCM bit depth, it will need to be transcoded again to avoid unexpected crashes. 
 - And last question: there doesn't seem to be a change in the MediaCodecAudioRenderer logic that determines the PCM version produced by the codec. I think this means that the codec will always produce float PCM at the moment and the new paths are not actually used unless you play PCM (without decoding) directly?

nift4:
Hi @tonihei ,
1.
I just deprecated `setEnableFloatOutput` but kept it for backwards compatibility. I don't see any need for a float output filter in the longer term, and if backward compat of unstable API, is not of concern, removing is totally fine for me.
2.
Audioflinger can do the transcoding in optimized native code (compared to ART AOT java code) no matter if actual output device supports it or not, so if API level (hence system-side audioflinger) supports the PCM mode, it should always work, or am I missing something? Which function was actually used for checking (because the 'getDirectPlaybackSupport' check will cause false negatives for normal AudioTrack usage as some devices do not support direct PCM at all - for example Pixel 7a)?
3.
MediaCodec integration is pending since 3 months, #2218 I would appreciate review :) I tested this PR together with #2218 for expected functionality and it works fine.

nift4:
Oh I just realized I missed a question,

> Is this to enable specific audio processors with a different PCM version?

No, the idea is that audio sink does least possible work. For example, imagine following scenario:
- Audio output device supports int32 only (an audio device in android only ever has one format sent to it, and this never changes until device reboot or device eject, with the exception of USB devices)
- Song is int24, sink will not output int24 because device doesn't support it
- So sink falls back to int16 and converts in java
- But because device only supports int32 audioflinger will convert a second time which wastes resources

always avoiding work in audio sink and letting audioflinger do the work avoids these without additional effort in coding, and will provide best battery life because conversion will be done in native code

nift4:
Sorry for the review request, it seems to have requested it from the wrong person. It seems like I don't have permission to assign / request microkatz

nift4:
@microkatz Hi, can you please take a look at this PR, as it has been a while? Thanks in advance!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioTrackAudioOutputProvider.java
```diff
@@ -494,15 +494,21 @@ private void verifySinglePlaybackLooper() {
...
-              && format.pcmEncoding == C.ENCODING_PCM_FLOAT)) {
-        return FORMAT_SUPPORTED_DIRECTLY;
+      if (format.pcmEncoding != C.ENCODING_PCM_16BIT
+          && !formatConfig.enableHighResolutionPcmOutput) {
+        // Only allow 16-bit PCM because high resolution PCM output is disabled.
+        return FORMAT_UNSUPPORTED;
       }
-      // We can resample all linear PCM encodings to 16-bit integer PCM, which AudioTrack is
-      // guaranteed to support.
-      return FORMAT_SUPPORTED_WITH_TRANSCODING;
```
nift4: cc @tonihei since you did the AudioOutputProvider split refactor:
the transcoding behaviour of DefaultAudioSink is hardcoded in AudioTrackAudioOutputProvider here. In the diff I have an example of how I would think a proper split would look like. Do you think that is a good idea?

tonihei: Yes, I agree, these constants currently encode behavior of the audio processors in DefaultAudioSink. I'll probably update the code directly because this fix is not fully related to the PR here as far as I can see.

tonihei: That was done in https://github.com/androidx/media/commit/4f4ad5f82d156767f89b9a6dec44c37bb8d153f4 (matching the current logic of the class)

nift4: Thanks, very nice! I've rebased this PR to fix the resulting conflicts. (and cleaned it up a little)

nift4: Hi @tonihei, related to the commit you just added, I have another additional comment: the audio sink still assumes that if it can transcode something to float or int16, then the audio output provider supports that. This assumption made sense when it was always AudioTrack because float is supported since SDK 23 and int16 since SDK 1, but now that audio output provider can be replaced, there is no guarantee for it to support both of these specific format anymore. (There could even be an audio output provider that supports neither. Of course, considering the sink currently only supports outputting either float or int16, that does not make sense with current main branch, but when this PR, the internal draft or anything in between is finally merged, it starts making sense. And considering the diverse possible choices for output provider - it could be USB device via libusb instead of AudioTrack, or Wi-Fi speaker system like AirPlay that gets a transcoded stream because it wants decoded PCM audio, or something else entirely, I don't think it should be hardcoded assumption.)

TL;DR

- output provider that supports only float32 would not be able play back low-resolution PCM formats because DefaultAudioSink will just assume unsupported without trying to use ToFloatPcmAudioProcessor.
- the same check should also be done for float32, as assumption that float output is enabled == float32 is supported will be invalidated by this PR (or the internal draft) eventually.

I have included a fix for this in this PR, but I think it might still not fully related to this PR, so let me know if you prefer to fix it seperately too, I would appreciate that.

nift4:
Hi @microkatz,

This PR has changed a bit since it was initially opened because of the audio output provider refactor. Part of what was in here is now obsolete, so I will describe the goals of this PR again:

* Change float output mode -> high resolution pcm mode. Implies that now, all high resolution PCM formats will be used, as long as they're supported by the AudioOutputProvider.
* For use with the mode described above, allow custom audio processor chain that supports all PCM formats. This involves abstracting the capabilties of the default audio processors into the DefaultAudioProcessorChain, which can be easily replaced, instead of hardcoding this in the audio sink itself.

This PR is mergable individually, it will not have any effect on a default setup though.

To achieve a noticable difference, PR #2218 is needed as well. That, too, is mergable individually, but will not have any effect on a default setup. Do you prefer to merge these seperately or should I combine them?

Thanks,
Nick

PR #2457, opened 2025-05-22
Should skip SEI payload when SEI type mismatch

### LINKED ISSUE #2456: Unexpected exception loading stream when playing hevc video with SEI
### Version

Media3 1.7.1 (same as 1.6.1)

### More version details

Since  https://github.com/androidx/media/commit/f673ef43b468159edb08fae3d292607ac497b8bc

Any devices

_No response_

Yes

### Reproduction steps

Play the attached sample file in demo app.

### Expected result

The media file should plays successfully.

### Actual result

1. "Playback failed" message on screen
2. Check logcat:
[stacktrace.log](https://github.com/user-attachments/files/20393606/stacktrace.log)

### Media

https://github.com/user-attachments/assets/fa6f0d25-fe07-4cbe-8c0d-a3f754abdf12

### Bug Report

pengbins:
This PR tries to fix  #2456

The function `parseH265Sei3dRefDisplayInfo` tries to find SEI of type 179 and extract 3d ref DisplayInfo from it. But when the SEI type does not match 179, it fails to properly skip the unprocessed SEI payload. Instead, it continues searching for the number 179 within the payload content. This may mistakenly treat payloads of other types SEI as the target type (179), leading to incorrect parsing and eventual through an exception.

The missing else branch is added to skip payload bytes.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -22,6 +22,8 @@
         placeholder seek points), and fall back to binary search seeking if the
         duration of the file is known
         ([#2327]()https://github.com/androidx/media/issues/2327).
+    *   Fix parsing of H.265 SEI units to fully skip unrecognized SEI types
+        ([#2456]()https://github.com/androidx/media/issues/2456).
 *   DataSource:
 *   Audio:
     *   Add support for all linear PCM sample formats in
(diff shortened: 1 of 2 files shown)


icbaker:
Thanks for the fix!

Some questions:
* Can we use `sample_with_sei_.mp4` to add a test for this (I'll do this in a follow-up after merging this PR)?
* Did you craft it from an existing MP4 sample asset?
    * Do you have an easy command for this?
    * And which sample file did you start with (so I can mark it as correctly branched in our VCS)?

icbaker:
Oh can you also re-target this PR agains the `main` branch please?

icbaker:
> Oh can you also re-target this PR agains the `main` branch please?

Alternatively I'm happy to make the change directly internally and send it for review, if you're happy to get the fix but without merging this PR - let me know :)

icbaker:

Issue #1764, opened 2024-09-27
Add flag to `onPlaybackResumption` to indicate whether playback should be started or not.


kelmer44:
### Use case description

`MediaLibrarySessionCallback`'s `onPlaybackResumption`  is called from `MediaSession.handleMediaControllerPlayRequest` to handle incoming play requests from MediaControllers, or from `MediaLibrarySession.getRecentMediaItemAtDeviceBootTime` to populate the media notification at boot time. While the former should end up with playback (or else resulting in a `ForegroundServiceDidNotStartInTimeException`), the latter should never explicitly call playback to avoid playing content upon device reboot.

Although for simpler cases media3 can handle this situation internally, for more complex cases where playback can't be resolved by merely returning a list of MediaItems in `onPlaybackResumption`, it would be useful to have a way to determine within this function whether the call should result in playback or not.

### Proposed solution
`onPlaybackResumption` may have an extra boolean parameter to indicate this situation, or the incoming controller info might contain an indication that the request is coming from MediaLibrary to populate the last media sesion within the OS.

### Alternatives considered
I currently inspect the call stack to reach this conclusion but this is hacky and far from ideal.

tonihei:
Thanks for the suggestion! 

> for more complex cases where playback can't be resolved by merely returning a list of MediaItems in onPlaybackResumption

Not sure if that's already clear, but `onPlaybackResumption` returns a `ListenableFuture`. This means you can create for example a `SettableFuture` that can be resolve asynchronously, whenever you are ready.  Does that already help with your integration. If not, it might be interesting to know how the logic differs assuming you have enough time to resolve the item that should be used for playback resumption.

kelmer44:
> Thanks for the suggestion!
> 
> > for more complex cases where playback can't be resolved by merely returning a list of MediaItems in onPlaybackResumption
> 
> Not sure if that's already clear, but `onPlaybackResumption` returns a `ListenableFuture`. This means you can create for example a `SettableFuture` that can be resolve asynchronously, whenever you are ready. Does that already help with your integration. If not, it might be interesting to know how the logic differs assuming you have enough time to resolve the item that should be used for playback resumption.

[...]

Issue #325, opened 2023-04-15
When adding new MediaItems while shuffleModeEnabled==true, the shuffleOrder is updated too late.


tzugen:
### Media3 Version

Media3 1.0.0

### Devices that reproduce the issue

All

### Devices that do not reproduce the issue

--

### Reproducible in the demo app?

Yes

### Reproduction steps

In https://github.com/androidx/media/issues/273#issuecomment-1472952138 Marc suggested to use `controller?.currentTimeline?.getFirstWindowIndex()` to get the windowIndex where playback should start.

Unfortunately, the underlying `shuffledWindowIndices` will be in regular sequential order after adding new items, and are only later being shuffled in ExoPlayerImpl.java#L2266.

This means that after enabling shufflePlayback and adding new items it is currently impossible to directly seek to the first item in the shuffled list (without resorting to some hack and waiting until the order is shuffled)

### Expected result

---

### Actual result

---

### Media

---

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to dev.exoplayer@gmail.com after filing this issue.

tzugen:
More information:
Directly after calling .addMediaItems() a Timeline change event is emitted. When I listen to this event and query the firstWindowIndex it is still 0.
One second later, another Timeline change event is emitted. Only during this second later event the playlist has been shuffled and the firstWindowIndex is actually some real value:

```
2023-04-15 15:05:48.442  7261-7261  MediaPlayerController   org.moire.ultrasonic.debug2          W  (shuffle) Adding media items
2023-04-15 15:05:48.455  7261-7261  MediaPlaye...$listeners org.moire.ultrasonic.debug2          W  On timeline changed. First shuffle play at index: 0
2023-04-15 15:05:49.287  7261-7261  MediaPlaye...$listeners org.moire.ultrasonic.debug2          W  On timeline changed. First shuffle play at index: 2
```

tonihei:
The underlying reason for this is that there is no good concept of `PlaybackOrder`/`ShuffleOrder` in the `Player` interface that would allow a `MediaController` to set or predict the order. 

I will turn this into a feature request to add such a concept to the `Player` interface, which should include the fix to the problem you are describing here.

tzugen:
Thanks, that sounds like a good plan.

naveensingh:
Could you provide an approximate timeline for when we can expect this feature request to be implemented?

[...]

PR #2458, opened 2025-05-22
Fix a bug and add a feature for media resumption

nift4:
Fix a bug where media resumption would break if prepare() was called in onCreate()

Media resumption code relied on the assumption that if no media
controller sends any commands, the player will stay in STATE_IDLE.
This optimization however breaks if a developer calls prepare()
in or shortly after (in my case, which caused this to be a race)
onCreate(). Instead, use playWhenReady to determine whether we need
to do a query with effort, because playWhenReady really shouldn't
be set to true on device bootup. :)
SystemUI asks for resumption support while a session is active and just
started playing (playWhenReady=true), hence the optimization is still valid.

--- 

Formal API to allow apps to handle playback resumption themselves

This could previously be achieved by inspecting stack trace of
onPlaybackResumption() and then returning settable future that will
never be set, but this API adds a cleaner way and makes the caveats
of this clearer.

Issue: https://github.com/androidx/media/issues/1764


tonihei:
Thanks for the suggestions!  This PR currently mixes multiple things together, and it may be helpful to untangle this a bit. As far as I can see, there are 3 proposals here:
 1. Update check for whether this is the boot time SysUI query for playback resumption to use `!playWhenReady` instead `state==IDLE`. 
    - This makes sense for the scenario you describe, but may fail if System UI checks playback resumption at a different time. The player may reasonably be in a paused state for example, which would trigger the wrong path here. I guess a better distinction is to check whether System UI started the service just for this check (vs connecting to an already existing service). Do you want to file a new issue for this so we can see if that's possible to implement? 
 2. Add new parameter to the onPlaybackResumption callback to distinguish the "information only" from the "playback will be started soon" case. This is actually quite useful to avoid work and I think we can use this PR to merge this part of the change.
 3. Add a way to intercept playback resumption handling entirely to do something else. It would be worth figuring out what "something else" is to see how it fits into the picture. The general assumption from our perspective was that you'd be able to call additional setup methods from onPlaybackResumption. In any case, it looks like this isn't implemented by this PR fully (or I missed the part where the new exception is thrown) and it's better to handle that separately.

nift4:
Hi @tonihei, thanks for your feedback.

> This makes sense for the scenario you describe, but may fail if System UI checks playback resumption at a different time. The player may reasonably be in a paused state for example, which would trigger the wrong path here. 

Yes, I don't think it's a problem to do that though as while the wrong path might be more expensive, it will not be functionally wrong. I also did some basic testing and didn't see SystemUI querying playback resumption while paused during that testing. Though I admit that it'd be better to detect that, I am not sure how that would work. I'll file an issue (edit: https://github.com/androidx/media/issues/2578) to discuss further and drop it from the PR for now. If we find a better way, great, and if not I can push the commit again into a new PR.

> it would be worth figuring out what "something else" is to see how it fits into the picture.

My usecase is actually just setShuffleOrder. Can't do that before setMediaItems (I get IllegalArgumentException iirc), but if library calls setMediaItems I don't see any non-racy way to add a setShuffleOrder call after that. For now I've hacked my ShuffleOrder to just return a predefined result in cloneAndInsert if something is loaded to disk, but it's rather suboptimal hack.
An alternative solution I can imagine is adding ShuffleOrder to MediaItemsWithStartPosition, but that would require going back to the drawing board in order to refactor for #325, which I do not see happening anytime soon (it's a hard problem - how do you serialize shuffle order and their possible mutations for media controllers sending them via Binder? if you want to support setting shuffle orders via Player interface, you'd need to handle that).

> or I missed the part where the new exception is thrown

The idea is that app code calls setException on the SettableFuture it returns from onPlaybackResumption when it wants to manually do playback resumption.

> Add new parameter to the onPlaybackResumption callback to distinguish the "information only" from the "playback will be started soon" case. This is actually quite useful to avoid work and I think we can use this PR to merge this part of the change.

I agree this is useful on it's own, but I specifically added it to go hand-in-hand with manually handling playback resumption, as manual handling only makes sense if we want to actually play stuff. I am not opposed to adding that first if you think the manual handling makes no sense or needs to be iterated upon first, though.

Let me know whether I should split out manual handling as well.

tonihei:
Thanks! I'll start the merge process for the additional `isForPlayback` parameter.

I have removed the `ManuallyHandlePlaybackResumption` for now as mentioned above. I agree that the shuffle order is a tricky case because it's not nicely integrated into the Player interface at the moment. In theory, you should be able to set an empty `ShuffleOrder` on the still empty `Player` from within this `onPlaybackResumption` callback. The `cloneWith...` methods can then be used to update the order to the real number of items in the playlist. This is needed to ensure the `ShuffleOrder` instance itself is deeply immutable and can be passed around between threads to match a specific playlist. There are other pending GH issues to make this setup more convenient if you want to customize the order to your needs. Returning your predefined result in `cloneAndInsert` is actually a good approach I think for now.

nift4:
Sounds good to me, thanks :D

PR #2465, opened 2025-05-23
Fix bug in CEA-608 caption display

bennettpeter:
I upgraded from 1.4.0 to 1.6.1. CEA-608 captions were displaying correctly in 1.4.0, but are now mangled. The correct letters are present but in the wrong order, for example see below (these are actual captions seen). The correct caption is in parentheses. This affects every caption, all are mangled.

    ybMae,ut b, still,ou y kno..w.    (Maybe, but still, you know...)
    Yeahyo, u t gotoe  bcarel.fu      (Yeah, you got to be careful.)
    I'gom nntea llat Prickha tt       (I'm gonna tell Patrick that)

The cause is that buffers are being processed out of order. I don't know why the ordering has changed between 1.4.0 and 1.6.1. Each buffer creates two characters, so pairs of characters are switched.

I made a simple change to fix it. If the buffer being processed and the next buffer are out of sequence, switch them. This fixes the problem in all of the above examples and others I have tested.

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/text/cea/CeaDecoder.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/text/cea/CeaDecoder.java
@@ -117,6 +117,14 @@ import java.util.ArrayDeque;
         return outputBuffer;
       }
 
+      // Cater for buffers out of sequence
+      CeaInputBuffer next = queuedInputBuffers.peek();
+      if (next != null &&  next.timeUs < inputBuffer.timeUs) {
+        queuedInputBuffers.poll(); // remove next from queue
+        queuedInputBuffers.addFirst(inputBuffer);
+        inputBuffer = next;
+      }
+
       decode(inputBuffer);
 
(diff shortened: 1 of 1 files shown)


icbaker:
Please can you try `1.8.0-alpha01` to see if this is fixed by https://github.com/androidx/media/commit/e5d32366491db3c3a43162fc87a1f71e5c5a9a23.

If not, please can you file an issue with a stream we can use to reproduce the problem.

CEA subtitles are now re-ordered by this file: https://github.com/androidx/media/blob/main/libraries/container/src/main/java/androidx/media3/container/ReorderingBufferQueue.java

Re-ordering in the decoder as you are doing here will not be possible when we move the CEA-608 parsing to happen as part 
of extraction (to be consistent with how other subtitle formats are now handled in ExoPlayer).

bennettpeter:
Than you for the prompt response.
1.8.0-alpha01 fixes the captions. 
Will you fix it in 1.6.1? 
Otherwise I can  use 1.6.1 with my patch until 1.8.0 is stable.

PR #2471, opened 2025-05-26
[b/372449691] Remove `packagingOptions` from `build.gradle` files

MGaetan89:
AGP is now set to 8.8.1, so the `packagingOptions` setup can be removed.


tonihei:
Thanks! I'll import this change now.

PR #2472, opened 2025-05-26
[b/306324391] Remove `try`/`catch` in `AudioCapabilities.Api33#getDefaultRoutedDeviceForAttributes()`

MGaetan89:
This mentioned commit (https://github.com/robolectric/robolectric/commit/442dff) was released in Robolectric 4.12.
Media3 now uses Robolectric 4.14.1, so this `try`/`catch` is not needed anymore.


tonihei:
Thanks for this change too, will import it to our internal repo and it should appear here as merged soon.

Issue #2433, opened 2025-05-16
1.7.0-alpha01 release has been marked on maven as 1.7.0 (stable)


icbaker:

MGaetan89:
Version 1.7.0 says to either stay on 1.6.1 or to update to 1.7.1.
But version 1.7.1 says to not use 1.7.1 and stay on 1.6.1.

Could you please align the message? Don't forget the release note files, if needed 🙂 

<img width="946" alt="Image" src="https://github.com/user-attachments/assets/c0a9f3a0-1d1d-4d23-b705-99d6b345462e" />

icbaker:
Thanks for flagging. 1.7.0 should be avoided, 1.7.1 is safe (but pointless) to upgrade to from 1.6.1. I've fixed the inconsistency:

* On the `main` branch: https://github.com/androidx/media/commit/4a21653dd8e8effdd45f783336f9dfcb1fa43054
* On the `release` branch: https://github.com/androidx/media/commit/b732b2a0aa3c186edb75c5b3dadc810b849071e9
* In the GitHub release info: https://github.com/androidx/media/releases

MGaetan89:
Thanks 👍🏻

PR #2473, opened 2025-05-26
[b/399861060] Remove `ChannelMixingMatrix#create()`

MGaetan89:
This method is deprecated and marked for removal in media3 1.8.0, which is currently in alpha.


icbaker:
This TODO is stale due to https://github.com/androidx/media/issues/2433. I've removed them in https://github.com/androidx/media/commit/1842aec64d38ab0a5b707cbe5efb2273b1f7eaa6 and they will be removed as part of our routine deprecation removal process (which works from oldest deprecations to newest).

PR #2496, opened 2025-06-02
[b/353490583] Remove outdated TODO in `libraries/muxer/build.gradle`

MGaetan89:
I've ran `./gradlew :lib-muxer:lintDebug` locally, and it didn't report any issue.
So I've enabled Lint for the test source set and removed the corresponding TODO.


icbaker:

Issue #1642, opened 2024-08-21
Client-side Ads on DASH Multi period video not working


dejan-korchev-deltatre:
Hello,

We having a case where we set Client-side Ads on DASH Multi period video,
but video stop after add and we receive the error.

If possible, can you provide an example for the right implementation or some kind of help?

The version of ExoPlayer that I am using is Media3 1.4.0
Thank you

Custom Test Case:

> {
  "name": "Client-side Ads on DASH Multi period video",
  "uri": "https://dash.akamaized.net/dash264/TestCases/5a/nomor/1.mpd",
  "ad_tag_uri": "https://pubads.g.doubleclick.net/gampad/ads?sz=640x480&iu=/124319096/external/single_ad_samples&ciu_szs=300x250&impl=s&gdfp_req=1&env=vp&output=vast&unviewed_position_start=1&cust_params=deployment%3Ddevsite%26sample_ct%3Dlinear&correlator="
}

Error:
`Caused by: java.lang.IllegalArgumentException
                                                                                                          at androidx.media3.common.util.Assertions.checkArgument(Assertions.java:40)
                                                                                                          at androidx.media3.exoplayer.source.ads.AdsMediaSource.onChildSourceInfoRefreshed(AdsMediaSource.java:289)
                                                                                                          at androidx.media3.exoplayer.source.ads.AdsMediaSource.onChildSourceInfoRefreshed(AdsMediaSource.java:63)
                                                                                                          at androidx.media3.exoplayer.source.CompositeMediaSource.lambda$prepareChildSource$0$androidx-media3-exoplayer-source-CompositeMediaSource(CompositeMediaSource.java:117)
                                                                                                          at androidx.media3.exoplayer.source.CompositeMediaSource$$ExternalSyntheticLambda0.onSourceInfoRefreshed(D8$$SyntheticClass:0)
                                                                                                          at androidx.media3.exoplayer.source.BaseMediaSource.refreshSourceInfo(BaseMediaSource.java:90)
                                                                                                          at androidx.media3.exoplayer.source.MaskingMediaSource.onChildSourceInfoRefreshed(MaskingMediaSource.java:211)
                                                                                                          at androidx.media3.exoplayer.source.WrappingMediaSource.onChildSourceInfoRefreshed(WrappingMediaSource.java:154)
                                                                                                          at androidx.media3.exoplayer.source.WrappingMediaSource.onChildSourceInfoRefreshed(WrappingMediaSource.java:49)
[...]

PR #2501, opened 2025-06-04
Multi period content support for AdsMediaSource

kotucz:
Multi period DASH content ads support inspired by https://github.com/androidx/media/issues/1642#issuecomment-2772077256 
At `Timeline` level, the original `AdPlaybackState` is divided and adjusted for each of the periods.

Removing single period assertions from `AdsMediaSource`.
Providing `MultiPeriodAdTimeline` as an alternative to current `SinglePeriodAdTimeline`.
From input `AdPlaybackState`, new modified `AdPlaybackState` is created - one for each period:
- ad group time is offset relative to period start time
- ad groups after period end time are marked as skipped
- post-roll ad group is kept only for last period
- ad group count and indices are kept constant

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/MultiPeriodAdTimeline.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/MultiPeriodAdTimeline.java
@@ -24,7 +24,15 @@ import androidx.media3.exoplayer.source.ForwardingTimeline;
 
 
 /**
- * A custom {@link Timeline} for sources that have AdPlaybackState split among multiple periods.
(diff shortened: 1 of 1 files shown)


icbaker:
Please sign the CLA, then we can take a look at this PR

kotucz:
Company CLA should be used. It is in progress

DedSecTeam17 (approved):

kotucz:
CLA updated and PR rebased

kotucz:
@marcbaechinger when can we expect some feedback on this PR, please?

We have also identified one edge case, with timeline not being updated properly. What would be the best way to raise the fix? Creating new issue/PR or putting all together in this one? It is more complicated scenario with more fix options

PR #2503, opened 2025-06-04
Fix: Made cue builder setText accept null

LagradOst:
The main problem that this PR solves it the ability to set Cues text to null, after it has been assigned. To illustrate the problematic API of no null currently, if you have a cue with text you can not in any way change it into a bitmap. You are forced to copy each and every field of the cue, making it a semver hazzard.

Alternatively make setBitmap also set text to null. 

```kotlin
val baseCue = Cue.Builder().setText("Hello, world!").build()
```
Case 1.
```kotlin
baseCue.buildUpon().setBitmap(bitmap).build() // will crash due to "Assertions.checkArgument(bitmap == null);"
```
Case 2.
```kotlin
baseCue.buildUpon().setBitmap(bitmap).setText(null).build() // will not compile due to "Null can not be a value of a non-null type CharSequence"
```

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/text/Cue.java
+++ b/libraries/common/src/main/java/androidx/media3/common/text/Cue.java
@@ -481,11 +481,14 @@ public final class Cue {
      *
      * <p>Note that {@code text} may be decorated with styling spans.
      *
+     * <p>Note that this will also set the {@code bitmap} to null.
+     *
      * @see Cue#text
      */
     @CanIgnoreReturnValue
-    public Builder setText(@Nullable CharSequence text) {
+    public Builder setText(CharSequence text) {
       this.text = text;
+      this.bitmap = null;
       return this;
     }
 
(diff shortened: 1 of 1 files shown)


icbaker:
Thanks for the contribution - your reasoning/issue explanation makes sense.

> Alternatively make `setBitmap` also set text to null.

I think it's reasonable to do this? (and same for `setText` setting bitmap to null)

If we add the `@Nullable` as you propose, then when switching a `Cue` from text to bitmap you force the dev to always explicitly call `setText(null)`. It seems reasonably safe (and documentable) to do this 'automatically' instead. I think of the `text` and `bitmap` fields as a ["union type"](https://en.wikipedia.org/wiki/Union_type) in spirit - so automatically clearing one when the other is set makes sense in that context.

icbaker:
Thanks! Please can you re-target this against the `main` branch (as that's where we merge PRs).

LagradOst:
Github just gave up.

LagradOst:
Can I just redo the PR as the re-targeting with the webui just fucks everything up?

icbaker:
Yeah opening a new PR is fine

PR #2506, opened 2025-06-05
[b/181312195] Make `CuesWithTimingSubtitle` package-private

MGaetan89:
`ExoplayerCuesDecoder` doesn't seem to exist anymore, so `CuesWithTimingSubtitle` can be made package-private.

> [!NOTE]
> - `CuesWithTimingSubtitle` was part of the public API, and not in an `internal` package. I don't know if making it package-private directly is how you want to proceed, or if it should be deprecated first?
> - Should the `@UnstableApi` annotation be kept for internal APIs?

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/text/CuesWithTimingSubtitle.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/text/CuesWithTimingSubtitle.java
@@ -20,7 +20,6 @@ import static androidx.media3.common.util.Assertions.checkArgument;
 import androidx.media3.common.C;
 import androidx.media3.common.text.Cue;
 import androidx.media3.common.util.Log;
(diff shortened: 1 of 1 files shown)


icbaker:
> `ExoplayerCuesDecoder` doesn't seem to exist anymore

For completeness, `ExoplayerCuesDecoder` was deleted in https://github.com/androidx/media/commit/002ee0555dc35dce9570f1a991b33ec92743db10

> I don't know if making it package-private directly is how you want to proceed

I think it's fine to do this immediately without deprecating it - but thanks for checking.

> Should the `@UnstableApi` annotation be kept for internal APIs?

This annotation can (and should) be removed for package-private types.

MGaetan89:
Thanks for the feedback @icbaker.
I've updated the PR accordingly.

icbaker:

PR #2509, opened 2025-06-05
Fix: Make cue setText and setBitmap null the other field.

LagradOst:
See https://github.com/androidx/media/pull/2503


icbaker:

PR #2513, opened 2025-06-06
Remove some unused dependencies

MGaetan89:
This PR removes some unused dependencies in the project. I've used https://github.com/autonomousapps/dependency-analysis-gradle-plugin (not included here) to detect them, and I've addressed the simplest ones:
- [demo-composition] Use more specific AndroidX Lifecycle dependencies.
- [demo-composition] Replace the hardcoded AndroidX Lifecycle version with `androidxLifecycleVersion`.
- [demo-session] Use `lifecycle-runtime` instead of the `-ktx` variant, because it is empty since [AndroidX Lifecycle 2.8.0](https://developer.android.com/jetpack/androidx/releases/lifecycle#2.8.0) (2.8.7 is used here).
- [demo-session-automotive] Remove AndroidX Core, AndroidX AppCompat and Material Components dependencies.
- [demo-session-automotive] Add AndroidX Annotation Experimental dependency.
- [demo-session-service] Remove AndroidX AppCompat dependency.
- [demo-shortform] Remove AndroidX ConstraintLayout and AndroidX Core dependencies.
- [demo-shortform] Remove duplicated Material Components dependency.
- [demo-shortform] Remove test dependencies since the module doesn't have any tests.
- [demo-transformer] Remove AndroidX RecyclerView dependency.
- [library-common] Remove Android Test dependencies since the module doesn't have any Android Tests.
- [library-muxer] Remove Android Test dependencies since the module doesn't have any Android Tests.

### DIFF
--- a/demos/composition/build.gradle
+++ b/demos/composition/build.gradle
@@ -69,8 +69,8 @@ dependencies {
     implementation project(modulePrefix + 'lib-ui')
     implementation project(modulePrefix + 'lib-ui-compose')
     implementation 'androidx.annotation:annotation:' + androidxAnnotationVersion
-    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.8.7'
-    implementation 'androidx.lifecycle:lifecycle-viewmodel-compose:2.8.7'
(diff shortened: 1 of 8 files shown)


icbaker:

PR #2525, opened 2025-06-10
Enable setOutputSurface workaround for SberBox

KingOfRaccoon:


tonihei:
Do you know on which API levels this applies? The current changes add the device to the sections that are commented with "// Workaround for MiTV and MiBox devices which have been observed broken up to API 28.", which I think is not the case for this device. We may need to add it to a separate section instead with the proper API check.

PR #2526, opened 2025-06-11
Handle channel count parsing for AC-4 with discrete objects

ybai001:
Some AC-4 stream consists of sub stream with discrete objects. Due to limitations in the current AC-4 specification (ETSI TS 103 190-2 V1.2.1 (2018-02)), discrete object number information is not explicitly stated in the ISO Base Media File Format (Annex E). To prevent exception triggering, "Maximum number of tracks if audio presentation includes object audio" in table 77 of ETSI TS 103 190-2 V1.2.1 (2018-02) is used as the channel count.

This pull request provides the corresponding implementation in AC4 parser.


libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java
```diff
@@ -415,28 +418,59 @@ public static Format parseAc4AnnexEFormat(
...
+          channelCount = channelCount == 17 ? 21 : channelCount;
+        }
+      } else {
+        // This presentation includes a substream with discrete objects. Due to limitations in the
+        // current AC-4 specification (ETSI TS 103 190-2 V1.2.1 (2018-02)), discrete object number
+        // information is not explicitly stated in the ISO Base Media File Format (Annex E). To
+        // prevent exceptions, "Maximum number of tracks if audio presentation includes object
+        // audio" in table 77 of ETSI TS 103 190-2 V1.2.1 (2018-02) is used as the channel count.
+        switch (ac4Presentation.level) {
+          case 0:
```
rohitjoins, line 447: Should this case be part of the default handling in the switch case?

libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java
```diff
@@ -415,28 +418,59 @@ public static Format parseAc4AnnexEFormat(
...
+          case 1:
+            channelCount = 6;
+            break;
+          case 2:
+            channelCount = 8;
+            break;
+          case 3:
+            channelCount = 10;
+            break;
+          case 4:
```
rohitjoins, line 460: Table 77 in spec `ETSI TS 103 190-2 V1.2.1 (2018-02)` states the value `4-6` is reserved, where do we get the value `12` for this?

libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java
```diff
@@ -415,28 +418,59 @@ public static Format parseAc4AnnexEFormat(
...
+            channelCount = 12;
+            break;
+          default:
```
rohitjoins, line 463: Shouldn't we throw an error from here instead of defaulting to channel count 2?

libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java
ybai001: > Should this case be part of the default handling in the switch case?

In that case, there will be a log output: AC-4 level 0 has not been defined. But level 0 is a defined AC-4 level. That's why I added this "case 0" handling.

libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java
ybai001: > Table 77 in spec `ETSI TS 103 190-2 V1.2.1 (2018-02)` states the value `4-6` is reserved, where do we get the value `12` for this?

This is an internal version that is not public. Level 4 has been defined. It's 12 in that doc. I have double checked with our AC-4 architect to double confirm it.

libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java
ybai001: The idea of mine is if there is new level defined in future, e.g. level 5, and if the application which integrates Media3 is not updated at that time, when playing a level 5 stream, there will be an exception triggered. But that device may has capability to decode this stream. So setting channelCount to 2 provides a chance for device to play it even if this stream may exceed the capability of device. Do you remember our discussion about [FORMAT_EXCEEDS_CAPABILITIES](https://github.com/androidx/media/blob/ae9a2339679826f870690092509a135dffc0d7e4/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/MediaCodecAudioRenderer.java#L373)? In this case, even if the input channel count is larger than the device capability, Media3 still try to play it rather than triggering an exception or stopping playback.

libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java
rohitjoins: Thank you for the clarification.

libraries/extractor/src/main/java/androidx/media3/extractor/Ac4Util.java
rohitjoins: Yeah, makes sense!

rohitjoins:

PR #2530, opened 2025-06-11
Fix various small warnings

MGaetan89:
This commit addresses the following warnings:
- Remove unnecessary semi-colons.
- Use `isEmpty()` instead of comparing the size to 0 (for collections) or the value to `""` (for strings).
- Remove unnecessary calls to `toString()`.
- Remove redundant type declarations.
- Use Kotlin's property instead of getter/setter.
- Replace redundant `StringBuilder` with `String`.
- Remove unnecessary casts.


rohitjoins:

Issue #2510, opened 2025-06-05
Vobsub subtitles not working with mp4 files


bennettpeter:
### Version

Media3 1.6.1

### More version details

vobsub subtitles work perfectly with mkv files but exoplayer does not recognize the subtitle track in mp4 files. The mp4 files are created with HandBrakeCLI. The vobsub subtitles are recognized and display when playing the video with VLC.

### Devices that reproduce the issue

Android Studio emulator Television (1080p) API 34
onn 4K streaming box
Amazon fire stick 4K

### Devices that do not reproduce the issue

None

### Reproducible in the demo app?

Yes

### Reproduction steps

Play the media in the demo app.
Arrow up a few times
Select SELECT TRACKS

### Expected result

Video, Audio and Text tracks available.
2 text tracks, English and Spanish

### Actual result

Only Video and Audio tracks available. No text tracks

### Media

https://drive.google.com/file/d/1CicUvxkKT5IZIHzy_ZsNqFsI_9tLt74K/view?usp=sharing

### Bug Report

- [x] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

icbaker:
Please can you link to or otherwise provide a reference to the spec which defines how VobSub subtitles should be carried in an MP4 container?

That will help us understand how this support should be added to ExoPlayer.

("handbrake produces files that look like X" isn't a spec :))

bennettpeter:
I did not mean to create an enhancement request. I thought a bug was causing it to fail on mp4's. 

I could not find a standard for vobsub on mp4. Perhaps there is no official standard for it. There is a method that handbrake uses and VLC understands. ffmpeg understans the vobsub subtitles and it is used by both handbrake and VLC. 

I will look at the code myself and if I can figure out an easy way to support this I will create a pull request.

bennettpeter:
This is now fixed with pull request #2531

Ass-xxx:
Thank You.

Issue #2491, opened 2025-06-01
FLV does not support H265


360060316:
### Version

Media3 1.6.1

### More version details

test:
https://ali.hlspull.yximgs.com/live/bjwsbjm8ik9ol0p1qaz2wsx3edc4rfv5tgb6yxsrjy36t35ty54y36yu2uhy24h35h45yh463uyh46u4rtyjtyktyestj4642gdhgsw.flv

### Devices that reproduce the issue

xiaomi 11 pro

### Devices that do not reproduce the issue

FLV does not support H265

### Reproducible in the demo app?

Yes

### Reproduction steps

FLV does not support H265

### Expected result

FLV does not support H265

### Actual result

FLV does not support H265

### Media

FLV does not support H265

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

icbaker:
It looks like the FLV container itself doesn't support H.265? At least based on the info here: https://en.wikipedia.org/wiki/Flash_Video#Encoding

ldm0206:
> It looks like the FLV container itself doesn't support H.265? At least based on the info here: https://en.wikipedia.org/wiki/Flash_Video#Encoding

Yes, at least it seems that Adobe does not intend to update the standard. But in fact, there are already many sources using this format. FFmpeg previously also refused to support it for the same reason (https://trac.ffmpeg.org/ticket/6389), but now they do support it (https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/b76053d8bf322b197a9d07bd27bbdad14fd5bc15).

It is difficult to make a definite conclusion on this issue, but I think the situation is the same as #1484, which belongs to a part that is not officially defined but has been widely used.

icbaker:
> I think the situation is the same as [#1484](https://github.com/androidx/media/issues/1484), which belongs to a part that is not officially defined but has been widely used.

I think the difference between this and #1484 is the scope/complexity of implementing (and maintaining) support for a whole new video encoding in a container (this request), vs a relatively simple URI scheme 'redirect'.

We find that maintaining compatibility with un(der)-documented media container specs is frustrating and challenging. Bugs are reported and it's always unclear whether the media has been prepared/encoded incorrectly, or whether the player impl is incorrect - because there is no spec to act as a clear arbiter of "correct".

Given that, I'm afraid we are extremely unlikely to add support for this (including accepting a PR for it).

ldm0206:
[...]

PR #2531, opened 2025-06-11
Extend Support for Vobsub subtitles to mp4 files

bennettpeter:
Refer to Issue #2510 . Vobsub subtitles work in mkv files but not in mp4 files.

I cannot find an official document describing how vobsub is implemented in mp4. It appears to be an extension invented by Nero Inc. / Nero AG for their "Nero Digital format" in 2003. It is supported by ffmpeg and I have traced through ffmpeg to find the details.

There is a new subtitle box type 'subp' (subpicture) that occurs under the handler alongside the existing types "text", "subt" ,etc. There is a new box type "mp4s" that occurs inside the "stsd" box and contains an esds box with initialization data for vobsub.

Box hierarchy:
```
moov
 trak
  mdia
   hdlr
    subp
     minf
      stbl
       stsd
        mp4s
         esds
          vobsub initialization data

```
The vobsub initialization data here is 64 bytes: 16 x 4-byte YUV palette entries. These must be converted to RGB for the vobsub parser. Also the vobsub parser needs the video width and height. It needs to be formatted as a vobsub idx file string, as follows (example):

size: 720x480\n
palette: feff00, feff00, feff00, 1f1f1f, 7f7f00, bebebe, 00fefc, fd00fe, feff00, 00007e, 008000, 7e0000, 00807d, 7f007f, 7f7f00, fefefe\n

In order to supply the size values, the values from the video track are needed. These are not available at the time when the mp4s box in being parsed. I have used a nasty hack to get the width and height from the video track into the vobsub subtitle tracks. I have added static variables to BoxParser to store the width and height from the video track so they can be accessed when parsing the vobsub. It would be preferable to apply the size to the vobsub track initialization data in parseTraks after all tracks have been parsed, but all of the track variables involved are final so cannot be modified. If the vobsub track occurs before the video track, default values will be used for width and height. This is unlikely.

I have added a test file libraries/test_data/src/test/assets/media/mp4/sample_with_vobsub.mp4 which is the test mkv file converted to mp4.

### DIFF
--- a/libraries/exoplayer/src/test/java/androidx/media3/exoplayer/e2etest/Mp4PlaybackTest.java
+++ b/libraries/exoplayer/src/test/java/androidx/media3/exoplayer/e2etest/Mp4PlaybackTest.java
@@ -15,6 +15,8 @@
  */
 package androidx.media3.exoplayer.e2etest;
 
+import static org.robolectric.annotation.GraphicsMode.Mode.NATIVE;
+
 import android.content.Context;
 import android.graphics.SurfaceTexture;
 import android.view.Surface;
@@ -36,8 +38,10 @@ import org.junit.runner.RunWith;
 import org.robolectric.ParameterizedRobolectricTestRunner;
 import org.robolectric.ParameterizedRobolectricTestRunner.Parameter;
 import org.robolectric.ParameterizedRobolectricTestRunner.Parameters;
+import org.robolectric.annotation.GraphicsMode;
 
 /** End-to-end tests using MP4 samples. */
+@GraphicsMode(NATIVE)
 @RunWith(ParameterizedRobolectricTestRunner.class)
 public class Mp4PlaybackTest {
 
--- a/libraries/test_data/src/test/assets/playbackdumps/mp4/sample_with_vobsub.mp4.dump
+++ b/libraries/test_data/src/test/assets/playbackdumps/mp4/sample_with_vobsub.mp4.dump
@@ -529,7 +529,7 @@ TextOutput:
   Subtitle[1]:
     presentationTimeUs = 0
     Cue[0]:
-      bitmap = length 296960, hash B7D4981
+      bitmap = length 296960, hash 6C5C74A1
       line = 0.88611114
       lineType = 0
       lineAnchor = 0


libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java
```diff
@@ -114,6 +117,16 @@ public final class BoxParser {
...
 
+  /** This is a hack to get the width and height from the video track into any vobsub
+   * subtitle tracks. It would be preferable to modify the vobsub track initialization
+   * data in parseTraks after all tracks have been parsed, but all of the structures
+   * involved are final so cannot be modified. These are default values that will be
+   * updated as video tracks are parsed, then used when a vobsub track is parsed.
+   * If the vobsub occurs before the video track, the default values here will be used.
+   */
+  private static int width = 720;
+  private static int height = 576;
```
icbaker: I'm afraid we can't accept this PR with this static mutable state. It's quite possible that there could be two ExoPlayer instances in the same JVM, both handling VobSub subtitles in MP4 files, and this static state would result in very confusing "cross talk" between them.

I think you will need to re-visit this plumbing to remove the static mutability.

bennettpeter: I agree this is bad. Thanks for the feedback. I will try to work it in a different way.

bennettpeter:
I have redone it in a way that is better and gets rid of the static variables. Also does not depend on  the video track being before the subtitle tracks.

In a couple of places I have used  "instanceOf" to check the type of object before casting the object and calling a new method. Another, and possibly better, way would be to add the new method to the interface used by the object, with a default empty implementation, and avoid the "instanceOf". Let me know if you would like me to change it that way.

icbaker:
I should give you a general warning about the likelihood of this PR being accepted: Since there's no spec for this, we have to weigh up the cost/complexity of the change, and the potential maintenance burden going forward. You can see a bit more of my general reasoning on this topic here: https://github.com/androidx/media/issues/2491#issuecomment-2939613651

I'm not saying we won't ever accept VobSub-in-MP4 support - but the bar for simplicity of implementation is higher than when adding support for a better specified extension.

The mutable static state initially suggested was clearly beyond what we'd accept (tbh even for a spec-supported change).

The latest version that requires adding VobSub-in-MP4-specific methods in the otherwise very general `SubtitleTranscodingExtractorOutput` is also likely too complex given the lack of spec I'm afraid. It also assumes that if an MP4 container has multiple video tracks, they all have the same size (which I'm not sure can be assumed in general).

Would it be possible to extract the width and height you need from the `tkhd` box of the subtitle track? That would hopefully reduce the blast radius of the change to make it more acceptable. Analysing your test file with https://gpac.github.io/mp4box.js/test/filereader.html I see this for the subtitle track (these are encoded in `16.16` format, so they decode to `1280x720`):

```
width	83886080
height	47185920
```

This is slightly different to the values in the video track `tkhd` (which decode to `1080x720`):

```
width	70778880
height	47185920
```

ExoPlayer doesn't currently parse the `width` and `height` from `tkhd` boxes at all (we read the width & height for video tracks from the `VideoSampleEntry` box instead), but we could - by extending the code here (beware that atm we stop reading halfway through the `matrix`, so you'll need to skip the rest of it before reading the width & height): 

https://github.com/androidx/media/blob/4423af424b3ba492f5a62d851ed6f5453ae0c3aa/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java#L955-L962

So I think this would require:

1. Parse width & height from `tkhd` into new fields on `TkhdData` class
2. Pass this width & height from `parseTrak(...)` into `parseStsd(...)` and then into `parseTextSampleEntry(...)`
    * It might be easiest to replace the `rotationDegrees` param with the `TkhdData` object.

bennettpeter:
All done as suggested. It works perfectly. A much better solution. All changes are in BoxParser.java, apart from one new constant that is defined in MP4Box.java.

The difference between the size in the video track tkhd and the subtitle tkhd seem to be that when the source is anamorphic, the subtitle track tkhd and the video track VideoSampleEntry have the source width and height, while the video track tkhd has the size adjusted to the rendering aspect ratio. For the subtitles it is the source width and height that is needed, so this solution is good.

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java
```diff
@@ -1206,21 +1214,73 @@ private static void parseTextSampleEntry(
...
-            .setId(trackId)
+            .setId(tkhdData.id)
             .setSampleMimeType(mimeType)
             .setLanguage(language)
             .setSubsampleOffsetUs(subsampleOffsetUs)
             .setInitializationData(initializationData)
             .build();
   }
 
+  private static int clip (int a) {
```
icbaker: You can use `Util.constrainValue` for this

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java
```diff
@@ -1206,21 +1214,73 @@ private static void parseTextSampleEntry(
...
+
+  private static int yuvToRgb(int yuv) {
+    int y, u, v;
+    int r,g,b;
+    y = (yuv >> 16) & 0xFF;
+    v = (yuv >> 8)  & 0xFF;
+    u =  yuv        & 0xFF;
+
+    r = clip(y + 14075 * (v - 128) / 10000);
+    g = clip(y - 3455 * (u - 128) / 10000 - 7160 * (v - 120) / 10000);
```
icbaker: should this be `v - 128`?

icbaker:
Thanks for refactoring to keep everything within the parsing of the single subtitle track.

> YUV palette entries. These must be converted to RGB for the vobsub parser.

I think this is the next contentious/awkward part of this PR. Do you have any context for why the Nero Digital Format uses YUV? As far as I can see, the 'original' VobSub IDX files use an RGB palette, so converting to YUV (and presumably forcing any parser to convert it back again) seems quite awkward!

I think (?) your conversion algorithm is based on BT.601 (i.e. SD) coefficients. Are you sure this is correct, and it shouldn't be BT.709 (i.e. HD)? The example in this PR only has white and black, so the difference might not be that noticeable.

bennettpeter:
In ffmpeg V6.0 the code to create the idx for vobsub in mp4 files is here:
https://github.com/FFmpeg/FFmpeg/blob/d388c347d41e4eb516dec05910551c5461e65615/libavformat/mov.c#L2356

It has a call to convert to RGB a few lines down, here:
https://github.com/FFmpeg/FFmpeg/blob/d388c347d41e4eb516dec05910551c5461e65615/libavformat/mov.c#L2372

Since ffmpeg is used for handbrake and vlc that seems a good place to find the correct implementation. I agree that the whole YUV palette seems odd. It makes sense for  the vob to use RGB but does not make much sense that ithe palette is stored in YUV in the mp4 file.

I found it difficult to find a YUV to RGB formula that gave the correct results. Wikipedia was not helpful. The formula I eventually found was for the older SD conversion. I figured that any difference would not be noticable.  Also I assume that since vobsub subtitles were created for DVDs the SD formula would be correct.

I could not figure out the formula that ffmpeg uses for the YUV to RGB conversion. It uses bit manipulation and goes into low level code.

If you want me to change to the HD conversion coefficients can you point me to where they can be found. I could not make sense of the wikipedia article, which has a formula that converts to R', G', B' and these do not seem to be the same as R G and B. I could not find any good information on it.

icbaker:
> Also I assume that since vobsub subtitles were created for DVDs the SD formula would be correct.

This reasoning makes sense to me - thanks.

I think we'll keep the logic in this class, and when I import this PR for internal review I might rename (and document) the `yuvToRgb` method to be VobSub specific, so it doesn't accidentally get re-used in a more general ExoPlayer/media3 context in future where it might not have the right coefficients.

icbaker (changes requested):

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java
bennettpeter: > You can use `Util.constrainValue` for this

Will do

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java
bennettpeter: You are correct. i will fix it.
These are the formulas I found, in case I messed up something else:
R = Y + 1.4075 * (V - 128)
G = Y - 0.3455 * (U - 128) - (0.7169 * (V - 128))
B = Y + 1.7790 * (U - 128)

bennettpeter: I also messed up 7169

icbaker:

PR #2553, opened 2025-06-21
Fix RTL punctuation issues with mixed BiDi subtitles

orenskl:
This PR fixed the punctuation issues with subtitles text containing a mix of RTL and LTR languages.

https://github.com/google/ExoPlayer/issues/11214
https://github.com/jellyfin/jellyfin-androidtv/issues/4296

### DIFF
--- a/libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
+++ b/libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
@@ -25,10 +25,12 @@ import android.graphics.Paint;
 import android.graphics.Paint.Join;
 import android.graphics.Paint.Style;
 import android.graphics.Rect;
+import android.text.BidiFormatter;
 import android.text.Layout.Alignment;
 import android.text.SpannableStringBuilder;
 import android.text.Spanned;
 import android.text.StaticLayout;
+import android.text.TextDirectionHeuristics;
 import android.text.TextPaint;
 import android.text.TextUtils;
 import android.text.style.AbsoluteSizeSpan;
@@ -190,7 +192,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
       return;
     }
 
-    this.cueText = cue.text;
+    this.cueText = containsRTL(cue.text) ? wrapBidiText(cue.text) : cue.text;
     this.cueTextAlignment = cue.textAlignment;
     this.cueBitmap = cue.bitmap;
     this.cueLine = cue.line;
@@ -469,5 +471,68 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
     // Some CharSequence implementations don't perform a cheap referential equality check in their
     // equals methods, so we perform one explicitly here.
     return first == second || (first != null && first.equals(second));
+  }
+
+  /**
+   * Checks whether the given {@link CharSequence} contains any characters
+   * with right-to-left (RTL) directionality.
+   * <p>
+   * This method inspects each character's Unicode directionality and returns {@code true}
+   * if at least one character is classified as RTL. This includes characters with the following
+   * directionalities:
+   * <ul>
+   *     <li>{@link Character#DIRECTIONALITY_RIGHT_TO_LEFT}</li>
+   *     <li>{@link Character#DIRECTIONALITY_RIGHT_TO_LEFT_ARABIC}</li>
+   *     <li>{@link Character#DIRECTIONALITY_RIGHT_TO_LEFT_EMBEDDING}</li>
+   *     <li>{@link Character#DIRECTIONALITY_RIGHT_TO_LEFT_OVERRIDE}</li>
+   * </ul>
+   *
+   * @param input the input {@link CharSequence} to analyze
+   * @return {@code true} if the input contains at least one RTL character; {@code false} otherwise
+   */
+  private static boolean containsRTL(@Nullable CharSequence input) {
+    if (input == null) {
+      return false;
+    }
+    for (int i = 0; i < input.length(); i++) {
+      char ch = input.charAt(i);
+      byte dir = Character.getDirectionality(ch);
+      if (dir == Character.DIRECTIONALITY_RIGHT_TO_LEFT ||
+          dir == Character.DIRECTIONALITY_RIGHT_TO_LEFT_ARABIC ||
+          dir == Character.DIRECTIONALITY_RIGHT_TO_LEFT_EMBEDDING ||
+          dir == Character.DIRECTIONALITY_RIGHT_TO_LEFT_OVERRIDE) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Applies bidirectional (bidi) Unicode wrapping to each line of the given {@link CharSequence}.
+   * <p>
+   * This method ensures that text containing both left-to-right (LTR) and right-to-left (RTL)
+   * scripts is displayed correctly by wrapping each line using {@link BidiFormatter#unicodeWrap}.
+   * It forces LTR context for wrapping and preserves line breaks.
+   *
+   * @param input the input text as a {@link CharSequence}, possibly containing mixed-direction text
+   * @return a {@link CharSequence} with each line wrapped for proper bidi rendering
+   */
+  public static CharSequence wrapBidiText(CharSequence input) {
+    BidiFormatter bidiFormatter = BidiFormatter.getInstance();
+    String[] lines = input.toString().split("\n");
+    StringBuilder wrapped = new StringBuilder();
+
+    for (int i = 0; i < lines.length; i++) {
+      wrapped.append(bidiFormatter.unicodeWrap(
+          lines[i],
+          TextDirectionHeuristics.LTR,
+          true
+      ));
+      if (i < lines.length - 1) {
+        wrapped.append("\n");
+      }
+    }
+
+    return wrapped;
   }
 }


libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -470,4 +472,67 @@ private static boolean areCharSequencesEqual(
...
+   * @param input the input {@link CharSequence} to analyze
+   * @return {@code true} if the input contains at least one RTL character; {@code false} otherwise
+   */
+  private static boolean containsRTL(@Nullable CharSequence input) {
+    if (input == null) {
+      return false;
+    }
+    for (int i = 0; i < input.length(); i++) {
+      char ch = input.charAt(i);
+      byte dir = Character.getDirectionality(ch);
```
icbaker: I think it would be better/more robust to deal in code points here, so it all works with supplementary plane characters too.

That probably means either:
* Gate this logic on API 24 and use `String.codePoints()`
* Or to support API 21+ use `String.codePointCount()` as your loop control length, with careful incrementing of `i` by 1 or 2 depending on whether you read a surrogate pair with `String.codePointAt` or not.
     * Beware that we are likely moving the whole library to min API 23 relatively soon, so then it just becomes "to support API 23"

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -470,4 +472,67 @@ private static boolean areCharSequencesEqual(
...
+    StringBuilder wrapped = new StringBuilder();
+
+    for (int i = 0; i < lines.length; i++) {
+      wrapped.append(bidiFormatter.unicodeWrap(
+          lines[i],
+          TextDirectionHeuristics.LTR,
+          true
+      ));
+      if (i < lines.length - 1) {
+        wrapped.append("\n");
```
icbaker: I would suggest collecting the lines into a list and using Guava's `Joiner` instead (or if you gate this on API 24 you can use [`Collectors.joining(CharSequence)`](https://developer.android.com/reference/java/util/stream/Collectors#joining(java.lang.CharSequence)) instead)

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -470,4 +472,67 @@ private static boolean areCharSequencesEqual(
...
+   * This method ensures that text containing both left-to-right (LTR) and right-to-left (RTL)
+   * scripts is displayed correctly by wrapping each line using {@link BidiFormatter#unicodeWrap}.
+   * It forces LTR context for wrapping and preserves line breaks.
+   *
+   * @param input the input text as a {@link CharSequence}, possibly containing mixed-direction text
+   * @return a {@link CharSequence} with each line wrapped for proper bidi rendering
+   */
+  public static CharSequence wrapBidiText(CharSequence input) {
+    BidiFormatter bidiFormatter = BidiFormatter.getInstance();
+    String[] lines = input.toString().split("\n");
```
icbaker: This looks like it might destroy any span styling (because you coerce the `CharSequence` (which might be a [`Spanned`](https://developer.android.com/reference/android/text/Spanned) at runtime) to a `String`.

icbaker (commented):
Thanks for the PR! I've added some comments. I was also wondering if you'd be able to provide an example stream or subtitle file that demonstrates the problem you're fixing here? So I can run the demo app without this patch, see the problem, then run it again and see it fixed.

I would also be interested to see what happens with span styling when this code is triggered - because atm I'm worried it might all be lost.

orenskl:
Hi, thanks for the comments.

This is a standard SRT file that can be used to reproduce this issue (just rename to srt if needed): 

[subtitles.txt](https://github.com/user-attachments/files/21109382/subtitles.txt)

Here are some examples:

These lines are displayed as:

Screenshot 2025-07-07 at 22 21 24

should be displayed as:

Screenshot 2025-07-07 at 22 51 06

These lines are displayed as:

Screenshot 2025-07-07 at 22 21 47

Should be displayed as:

Screenshot 2025-07-07 at 22 50 38

You can notice that the punctuation marks are displayed to the right of the sentences and should be displayed to the left (as in all RTL languages). Hope you are good with Hebrew :-)

icbaker:
Thank you for the test file and screenshots, that's really helpful.

I took your SRT file and adapted it to WebVTT so I could add some styling (borrowed from [one of our test files](https://github.com/androidx/media/blob/release/libraries/test_data/src/test/assets/media/webvtt/with_css_styles)), and confirmed that this PR currently strips styling info.

Using this file (just rename to `.vtt` suffix): 
[vtt-2553.txt](https://github.com/user-attachments/files/21120053/vtt-2553.txt)

Compare before your change (styling visible, punctuation in the wrong place):

android-screenshot-2025-07-08T113424

android-screenshot-2025-07-08T113602

With after (styling lost, punctuation in the correct place):

android-screenshot-2025-07-08T113753

android-screenshot-2025-07-08T113803

----

Please take a look and see if you can tweak this PR so it puts the punctuation in the correct place while preserving (or restoring) the span styling.

orenskl:
I was able to reproduce the styling issues with the VTT file on my side, I think the best option here will be to manipulate the cues and then restore the original spans with the indices fixed. I am working on this.

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -470,4 +476,117 @@ private static boolean areCharSequencesEqual(
...
+            (spannedInput.getSpanStart(spans[j]) < lineStart + line.length())) {
+          spanStarts[j] = spanUpdate;
+        }
+        if ((spanEnds[j] < 0) &&
+            ((spannedInput.getSpanEnd(spans[j]) - 1) >= lineStart) &&
+            ((spannedInput.getSpanEnd(spans[j]) - 1) < lineStart + line.length())) {
+          spanEnds[j] = spanUpdate;
+        }
+      }
+      lineStart += line.length() + 1; // +1 for the newline character
```
icbaker: What about `\r\n` newlines?

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -470,4 +476,117 @@ private static boolean areCharSequencesEqual(
...
+        // Each span start or end is updated only once
+        if ((spanStarts[j] < 0) &&
+            (spannedInput.getSpanStart(spans[j]) >= lineStart) &&
+            (spannedInput.getSpanStart(spans[j]) < lineStart + line.length())) {
+          spanStarts[j] = spanUpdate;
+        }
+        if ((spanEnds[j] < 0) &&
+            ((spannedInput.getSpanEnd(spans[j]) - 1) >= lineStart) &&
+            ((spannedInput.getSpanEnd(spans[j]) - 1) < lineStart + line.length())) {
+          spanEnds[j] = spanUpdate;
```
icbaker: I'm afraid I don't fully understand this code.

Is it intended to support spans that start on one line and end on a later line?

If I'm reading the code below on L579 correctly, if the start of span X is in `spanStarts[j]` then the end of the same span must be at the same index: `spanEnds[j]`.

But I'm not sure I understand how that invariant is enforced up here.

Would you consider pulling this `wrapBidiText` method out into a package-private class so we can add unit tests? That would give me more confidence that this works (and will keep working). See for example `SpannedToHtmlConverterTest.java`.

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -470,4 +476,117 @@ private static boolean areCharSequencesEqual(
...
+   * It forces LTR context for wrapping and preserves spans and line breaks.
+   *
+   * @param input the input text as a {@link CharSequence}, possibly containing mixed-direction text
+   * @return a {@link CharSequence} with each line wrapped for proper bidi rendering
+   */
+  public static CharSequence wrapBidiText(CharSequence input) {
+    BidiFormatter bidiFormatter = BidiFormatter.getInstance();
+
+    // Preserve span in the input text.
+    Spanned spannedInput = (Spanned)input;
```
icbaker: This cast should be conditional on an `instanceof` check (i.e. you should also support cues where the text is just a `String` and not a `Spanned`)

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -470,4 +476,117 @@ private static boolean areCharSequencesEqual(
...
+    Object[] spans = spannedInput.getSpans(0, input.length(), Object.class);
+
+    // Create arrays to track the start and end of each span after wrapping.
+    int[] spanStarts = new int[spans.length];
+    int[] spanEnds = new int[spans.length];
+    Arrays.fill(spanStarts, -1);
+    Arrays.fill(spanEnds, -1);
+
+    String[] lines = input.toString().split("\n");
```
icbaker: Please use `Util.split` to avoid the gotchas described here: https://github.com/google/guava/wiki/stringsexplained#splitter

icbaker (changes requested):

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
orenskl: Yes, this is intended to support Spans that start on one line and end on another, the `spanEnd` and `spanStart` array hold each the start and end adjustments needed for each Span. Each `spans[j]` has a corresponding `spanEnd[j]` and `spanStart[j]` adjustments. There is a separate `if` statement for checking the start of a span and the end of a span. The `spanEnd` and `spanStart` arrays are initialized to -1 so that we can remember to update each Span start and end only once.
I hope this makes it more clear I can also add it to the code as a comment. 

Should I add unit tests as well ?

icbaker: Ah cool thanks, now I look at it again I understand how it works.

> Should I add unit tests as well ?

Yes I think it would be very useful to add some tests in this PR.

orenskl:
Hi, is there anything I can do to help you merge this PR ?

icbaker:
Thanks for pinging, I was waiting for the tests and didn't realise they'd already been added (I think just 'resolving' a conversation on GitHub maybe doesn't trigger an email - so I didn't know to look back here) - I'll take another look.

icbaker:

Issue #2542, opened 2025-06-18
Is UAMP's media3 branch still maintained?


nift4:
Hi,

is https://github.com/android/uamp/tree/media3 still maintained (because last commit was quite a while ago)? Or is there another sample as comprehensive (supports phone, tablet, car, etc)?

Thanks!

andrewlewis:
@sigmanzero Nevin, please take a look?

tonihei:
Nevin can correct me, but I think we stopped maintaining UAMP because it's bound to be too easily out of date. The Media3 Session demo is the main demo app for everything related to session support and should showcase all the main features. Is there anything in particular you are missing from the session demo app?

nift4:
My primary concern at this time is PackageValidator: https://github.com/android/uamp/blob/media3/common/src/main/java/com/example/android/uamp/media/PackageValidator.kt
and the associated signature list:
https://github.com/android/uamp/blob/media3/common/src/main/res/xml/allowed_media_browser_callers.xml
because those two are non-trivial to keep up to date, and just being able to copy-paste these from UAMP and having every known client to MediaLibraryService covered was helpful. Of course, something like this integrated into media3 itself would be even better.
The session demo doesn't seem to perform any validation, but I'm not sure if it's appropriate to allow any app access to the user's music through MediaBrowser (I imagine this could be misused for fingerprinting?).

tonihei:
In Media3 there are the utils like `isMediaNotificationController`, `isAutomotiveController` and `isAudioCompanionController` that are also used in the demo app in [`onConnect`](https://github.com/androidx/media/blob/ae9a2339679826f870690092509a135dffc0d7e4/demos/session_service/src/main/java/androidx/media3/demo/session/DemoMediaLibrarySessionCallback.kt#L75) to provide different capabilities than to other random callers. Is this what you are looking for?

As far as I can see, the UAMP filtering has a few more like `isWearCompanionApp` or `isGoogleApp`. Do you think it's useful to add these the Media3 as utils? In the end, it's down to your app to decide whether the browsing is open to everyone or a specific list of use cases. 

[...]

PR #2555, opened 2025-06-23
Trust controllers from apps in current UID

nift4:
i.e. either sharedUid apps, or the current app (though possibly another process)

Issue: https://github.com/androidx/media/issues/2542#issuecomment-2991223070


nift4:
Thank you!

PR #2570, opened 2025-06-29
c2.sec.flac.decoder does not support 32-bit audio on Android 14

nift4:
fixes 32-bit FLAC playback on Galaxy S23 FE (Snapdragon 8 Gen 1 version)

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecInfo.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecInfo.java
@@ -382,7 +382,7 @@ public final class MediaCodecInfo {
   }
 
   private boolean isCompressedAudioBitDepthSupported(Format format) {
-    // MediaCodec doesn't have a way to query decoder bit-depth support.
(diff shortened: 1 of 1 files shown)


microkatz:
@nift4 

Thank you for submitting your PR request! 

If I may ask, do you know or did you test if the c2.sec.flac.decoder actually supports 32-bit audio on Android 15? Or is the purpose of this PR to make devices on 34 utilize the c2.android.flac.decoder? If the latter, then this PR is just a patch and not a full fix.

nift4:
I don't have physical access to a newer OneUI device for testing, but because I didn't get bug reports about newer versions, I assumed it was working.

However your question inspired me to test on Firebase Test Lab, and the newest available device there (S25 Ultra on Android 15) still didn't have working 32-bit FLAC playback.

Should I change the check to return false no matter which SDK if c2.sec.flac.decoder is queried, or do you want to push a commit that does that? (Asking because I don't want to mess up the merge process, not sure if you already started it.)

nift4:
@microkatz 
I have edited the PR. Now, it should be a full fix.

nift4:
Hi @microkatz, I have previously addressed your review comment. Can you please check this PR again? Thanks!

PR #2590, opened 2025-07-05
Bump Robolectric to 4.15.1

utzcoz:
Bump Robolectric to 4.15.1 and adapt test code to flush runnables before internal checking.

### DIFF
--- a/libraries/exoplayer_dash/src/test/java/androidx/media3/exoplayer/dash/e2etest/DashPlaybackTest.java
+++ b/libraries/exoplayer_dash/src/test/java/androidx/media3/exoplayer/dash/e2etest/DashPlaybackTest.java
@@ -88,6 +88,7 @@ import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
 import org.robolectric.annotation.Config;
+import org.robolectric.shadows.ShadowLooper;
 
 /** End-to-end tests using DASH samples. */
 @RunWith(AndroidJUnit4.class)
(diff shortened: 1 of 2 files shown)


libraries/exoplayer_dash/src/test/java/androidx/media3/exoplayer/dash/e2etest/DashPlaybackTest.java
```diff
@@ -357,6 +358,9 @@ public void onLoadCompleted(
         .ignoringNonFatalErrors()
         .untilBackgroundThreadCondition(() -> secondSubtitleFailureCount.get() == 4);
     secondSubtitleResolves.set(true);
+
+    ShadowLooper.runUiThreadTasks();
```
marcbaechinger: Can you explain what issue this is addressing and how it is solved by this line.

In any case I think I'd prefer to separate bumping the version and applying a fix for the dash playback test. When I add this line without bumping the version it appears that this would work as well, so I think these two changes can be separated.

If you can provide me with the reasoning for this change here, I can do a CL internally that bumps the robolectric version and we can then turn this change here to a commit that addresses the problem you are seeing.

utzcoz: This change is safe on the old version, and is used to avoid extra Robolectrie's looper left runnable checking. I can send a new PR for this change separately with old Robolectric version.

marcbaechinger: Would be great to split this, yes!

utzcoz: https://github.com/androidx/media/pull/2628 New pr for test code changes.

icbaker:

utzcoz:
Thanks for your work to make it merged.

PR #2609, opened 2025-07-16
Exclude AC4Profile22 from estimated AC-4 profiles on automotive

dwhea:
Automotive device AC-4 MediaCodec decoders are not expected to support AC4Profile22 (with Level 0), and other legacy AC4 profiles.


microkatz:
@dwhea 

This issue has been addressed in this commit, https://github.com/androidx/media/commit/68f1011da7f64e5c29d26c44ae8638122439648b. Apologies for taking it upon ourselves to implement rather than merge in your PR. It required additional changes for which it was easier to complete ourselves. If the solution does not address the problem then please create an issue ticket! Thanks!

PR #2619, opened 2025-07-17
Add `SphericalGLSurfaceView` support to `PlayerSurface`

MGaetan89:
This change allows `PlayerSurface` to render video using `SphericalGLSurfaceView` when specifying the `SURFACE_TYPE_SPHERICAL_GL_SURFACE_VIEW` type.

The Compose demo has also been updated to include a 360 video example that utilizes this new surface type.

> [!NOTE]
> The 360° video used in the video comes from [Bitmovin](https://bitmovin.com/demos/vr-360/). I've added a link to the page in the demo source code. Let me know if you'd rather use a different video or handle attribution differently.

### DIFF
--- a/demos/compose/src/main/java/androidx/media3/demo/compose/MainActivity.kt
+++ b/demos/compose/src/main/java/androidx/media3/demo/compose/MainActivity.kt
@@ -30,6 +30,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.Button
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
@@ -44,6 +45,7 @@ import androidx.lifecycle.compose.LifecycleResumeEffect
 import androidx.lifecycle.compose.LifecycleStartEffect
 import androidx.media3.common.MediaItem
 import androidx.media3.common.Player
+import androidx.media3.common.listen
 import androidx.media3.demo.compose.buttons.ExtraControls
 import androidx.media3.demo.compose.buttons.MinimalControls
 import androidx.media3.demo.compose.data.videos
@@ -102,7 +104,7 @@ fun ComposeDemoApp(modifier: Modifier = Modifier) {
 
 private fun initializePlayer(context: Context): Player =
   ExoPlayer.Builder(context).build().apply {
-    setMediaItems(videos.map(MediaItem::fromUri))
+    setMediaItems(videos.keys.map(MediaItem::fromUri))
     prepare()
   }
 
@@ -110,10 +112,19 @@ private fun initializePlayer(context: Context): Player =
 private fun MediaPlayerScreen(player: Player, modifier: Modifier = Modifier) {
   var showControls by remember { mutableStateOf(true) }
   var currentContentScaleIndex by remember { mutableIntStateOf(0) }
+  var surfaceType by remember { mutableIntStateOf(SURFACE_TYPE_SURFACE_VIEW) }
   val contentScale = CONTENT_SCALES[currentContentScaleIndex].second
 
   val presentationState = rememberPresentationState(player)
   val scaledModifier = Modifier.resizeWithContentScale(contentScale, presentationState.videoSizeDp)
+
+  LaunchedEffect(player) {
+    player.listen {
+      currentMediaItem?.localConfiguration?.let {
+        surfaceType = videos.getValue(it.uri.toString())
+      }
+    }
+  }
 
   // Only use MediaPlayerScreen's modifier once for the top level Composable
   Box(modifier) {
@@ -122,7 +133,7 @@ private fun MediaPlayerScreen(player: Player, modifier: Modifier = Modifier) {
     // because the Player will not emit the relevant event, e.g. the first frame being ready.
     PlayerSurface(
       player = player,
-      surfaceType = SURFACE_TYPE_SURFACE_VIEW,
+      surfaceType = surfaceType,
       modifier = scaledModifier.noRippleClickable { showControls = !showControls },
     )
 
--- a/demos/compose/src/main/java/androidx/media3/demo/compose/data/videos.kt
+++ b/demos/compose/src/main/java/androidx/media3/demo/compose/data/videos.kt
@@ -15,10 +15,15 @@
  */
 package androidx.media3.demo.compose.data
 
+import androidx.media3.ui.compose.SURFACE_TYPE_SPHERICAL_GL_SURFACE_VIEW
+import androidx.media3.ui.compose.SURFACE_TYPE_SURFACE_VIEW
+
 val videos =
-  listOf(
-    "https://html5demos.com/assets/dizzy.mp4",
-    "https://storage.googleapis.com/exoplayer-test-media-0/shortform_2.mp4",
-    "https://storage.googleapis.com/exoplayer-test-media-1/gen-3/screens/dash-vod-single-segment/video-vp9-360.webm",
-    "https://storage.googleapis.com/exoplayer-test-media-0/shortform_3.mp4",
+  mapOf(
+    "https://html5demos.com/assets/dizzy.mp4" to SURFACE_TYPE_SURFACE_VIEW,
+    "https://storage.googleapis.com/exoplayer-test-media-0/shortform_2.mp4" to SURFACE_TYPE_SURFACE_VIEW,
+    "https://storage.googleapis.com/exoplayer-test-media-1/gen-3/screens/dash-vod-single-segment/video-vp9-360.webm" to SURFACE_TYPE_SURFACE_VIEW,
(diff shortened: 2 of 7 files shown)


demos/compose/src/main/java/androidx/media3/demo/compose/MainActivity.kt
```diff
@@ -102,27 +104,36 @@ fun ComposeDemoApp(modifier: Modifier = Modifier) {
...
 private fun MediaPlayerScreen(player: Player, modifier: Modifier = Modifier) {
   var showControls by remember { mutableStateOf(true) }
   var currentContentScaleIndex by remember { mutableIntStateOf(0) }
+  var surfaceType by remember { mutableIntStateOf(SURFACE_TYPE_SURFACE_VIEW) }
   val contentScale = CONTENT_SCALES[currentContentScaleIndex].second
 
   val presentationState = rememberPresentationState(player)
   val scaledModifier = Modifier.resizeWithContentScale(contentScale, presentationState.videoSizeDp)
 
+  LaunchedEffect(player) {
```
oceanjules, line 121: TBD: this will be triggered on all the events, even when they are not relevant to localConfiguration. We are considering adding a `player.listen(Player.Events, Player.()->Unit)` overload that will let you specify the relevant events for the actionable-trailing-lambda. I'll update this comment once I push that.

demos/compose/src/main/java/androidx/media3/demo/compose/MainActivity.kt
```diff
@@ -102,27 +104,36 @@ fun ComposeDemoApp(modifier: Modifier = Modifier) {
...
+    player.listen {
+      currentMediaItem?.localConfiguration?.let {
```
oceanjules, line 123: Is there any other way to get the surface type out earlier? I would like to keeo MediaPlayerScreen as clean as possible since it will be the skeleton for the future PlayerView equivalent.

So far any listener subscriptions have been encapsulated into State classes. I don't think PresentationState would be the right class to add mediaItem-related info to. But maybe TimelineState? MetadataState?

```
suspend fun observe(): Nothing =
    player.listen { events ->
      // TODO: or other events as well
      if (events.contains(Player.EVENT_MEDIA_ITEM_TRANSITION)) {
         uri = player.currentMediaItem?.localConfiguration?.uri
      }
   }
}
```

Then this Composable will remain as

```
var surfaceType by remember { mutableIntStateOf(SURFACE_TYPE_SURFACE_VIEW) }
val timelineState = rememberTimelineState(player)
surfaceType = videos.getValue(timelineState.uri)
```

demos/compose/src/main/java/androidx/media3/demo/compose/data/videos.kt
```diff
@@ -15,10 +15,15 @@
...
-    "https://html5demos.com/assets/dizzy.mp4",
-    "https://storage.googleapis.com/exoplayer-test-media-0/shortform_2.mp4",
-    "https://storage.googleapis.com/exoplayer-test-media-1/gen-3/screens/dash-vod-single-segment/video-vp9-360.webm",
-    "https://storage.googleapis.com/exoplayer-test-media-0/shortform_3.mp4",
+  mapOf(
+    "https://html5demos.com/assets/dizzy.mp4" to SURFACE_TYPE_SURFACE_VIEW,
+    "https://storage.googleapis.com/exoplayer-test-media-0/shortform_2.mp4" to SURFACE_TYPE_SURFACE_VIEW,
+    "https://storage.googleapis.com/exoplayer-test-media-1/gen-3/screens/dash-vod-single-segment/video-vp9-360.webm" to SURFACE_TYPE_SURFACE_VIEW,
+    "https://storage.googleapis.com/exoplayer-test-media-0/shortform_3.mp4" to SURFACE_TYPE_SURFACE_VIEW,
+    // https://bitmovin.com/demos/vr-360/
```
oceanjules, line 27: nit: leftover uri

libraries/ui_compose/build.gradle
```diff
@@ -49,6 +49,7 @@ android {
 dependencies {
     api project(modulePrefix + 'lib-common')
     api project(modulePrefix + 'lib-common-ktx')
+    api project(modulePrefix + 'lib-exoplayer')
```
oceanjules: this is probably the main blocker atm - we don't want this `lib-ui-compose` module to depend on exoplayer. The same as `lib-ui` didn't (https://github.com/androidx/media/blob/release/libraries/ui/build.gradle#L38)

What this means practically is matching PlayerView's reflection:

https://github.com/androidx/media/blob/b7bbc6e2bc3e45ff3ed99884c114c50f03bba5c9/libraries/ui/src/main/java/androidx/media3/ui/PlayerView.java#L457-L471

libraries/ui_compose/src/main/java/androidx/media3/ui/compose/PlayerSurface.kt
```diff
@@ -130,16 +147,18 @@ private var View.attachedPlayer: Player?
...
 @Target(AnnotationTarget.CLASS, AnnotationTarget.TYPE, AnnotationTarget.TYPE_PARAMETER)
-@IntDef(SURFACE_TYPE_SURFACE_VIEW, SURFACE_TYPE_TEXTURE_VIEW)
+@IntDef(SURFACE_TYPE_SURFACE_VIEW, SURFACE_TYPE_TEXTURE_VIEW, SURFACE_TYPE_SPHERICAL_GL_SURFACE_VIEW)
 annotation class SurfaceType
 
 /** Surface type to create [android.view.SurfaceView]. */
 @UnstableApi const val SURFACE_TYPE_SURFACE_VIEW = 1
 /** Surface type to create [android.view.TextureView]. */
 @UnstableApi const val SURFACE_TYPE_TEXTURE_VIEW = 2
+/** Surface type to create [androidx.media3.exoplayer.video.spherical.SphericalGLSurfaceView]. */
```
oceanjules: reflection during runtime (see comment in `build.gradle`) will probably mean you won't be able to use this fully qualified refernce in the comment.

libraries/ui_compose/src/main/java/androidx/media3/ui/compose/PlayerSurface.kt
```diff
@@ -71,6 +73,16 @@ fun PlayerSurface(
         setVideoView = Player::setVideoTextureView,
         clearVideoView = Player::clearVideoTextureView,
       )
+    SURFACE_TYPE_SPHERICAL_GL_SURFACE_VIEW ->
+      PlayerSurfaceInternal(
```
oceanjules, line 76: Somewhere here, pseudo code:

```
    var clazz: Class<*>?
    try {
      clazz = Class.forName("androidx.media3.exoplayer.video.spherical.SphericalGLSurfaceView")
      val surfaceView = clazz.getConstructor(Context.class)   
      ...
      createView = { surfaceView.newInstance(context) } 
      ...
      }
    } catch (e: ClassNotFoundException) {
      // Expected if ExoPlayer module not available.
      throw IllegalArgumentException("spherical_gl_surface_view requires an ExoPlayer dependency")
    } catch (e: NoSuchMethodException) { 
      throw IllegalArgumentException("... something else went wrong..")
    }
```

demos/compose/src/main/java/androidx/media3/demo/compose/MainActivity.kt
MGaetan89: I'll submit a dedicated PR to add this new state so it will be easier to discuss

demos/compose/src/main/java/androidx/media3/demo/compose/data/videos.kt
MGaetan89: I put this comment to explain where the video comes from.
Should I remove it or clarify it (for example `360 video coming from https://bitmovin.com/demos/vr-360/`)?

libraries/ui_compose/build.gradle
MGaetan89: I've removed the dependency on `lib-exoplayer` in daa4ae578e77efeabe32a86e197b63706e07bf60, and used reflection instead.

libraries/ui_compose/src/main/java/androidx/media3/ui/compose/PlayerSurface.kt
MGaetan89: I've updated the comment to just use the class name instead of a link in daa4ae578e77efeabe32a86e197b63706e07bf60.

libraries/ui_compose/src/main/java/androidx/media3/ui/compose/PlayerSurface.kt
MGaetan89: I've updated my initial logic to use reflection instead in daa4ae578e77efeabe32a86e197b63706e07bf60.

demos/compose/src/main/java/androidx/media3/demo/compose/MainActivity.kt
MGaetan89: I've opened #2700 for this. So I'm marking this comment as resolved, and we can continue the discussion in the other PR.

PR #2625, opened 2025-07-17
Provide context to onForegroundServiceStartNotAllowedException

nift4:
The intended use of this method is to "post a notification from where the user journey can continue", but posting a notification requires a context.


marcbaechinger:

nift4:
thx :)

Issue #2622, opened 2025-07-17
onPlaybackResumption() returning empty list leads to hard-to-debug ForegroundServiceDidNotStartInTimeException


nift4:
### Version

Media3 main branch

### Devices that reproduce the issue

Android 13 and later

### Devices that do not reproduce the issue

older devices

### Reproducible in the demo app?

Yes

### Reproduction steps

1. Press "Play" button on headset
2. ensure onPlaybackResumption() returns empty MediaItemsWithStartPosition

### Expected result

Nothing happens, app crashes with useful crash or atleast with a message in the logcat. Or at least the crash being documented.

I'd suggest (similar to #2591) throwing an exception if we know we are going to die anyway. But this should also be documented.

### Actual result

ForegroundServiceDidNotStartInTimeException

### Media

Not applicable

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

Issue #2591, opened 2025-07-06
onPlaybackResumption() returning exception leads to hard-to-debug ForegroundServiceDidNotStartInTimeException


nift4:
### Version

Media3 main branch

### Devices that reproduce the issue

Android 13 and later

### Devices that do not reproduce the issue

older devices

### Reproducible in the demo app?

Yes

### Reproduction steps

1. Press "Play" button on headset
2. ensure onPlaybackResumption() returns exception (or: just don't override it at all)

### Expected result

useful crash

### Actual result

App crashes with ForegroundServiceDidNotStartInTimeException because media button reciever (by default) always forwards play events, app returns empty MediaItemsWithStartPosition (or some exception) so the play handling just moves player to STATE_ENDED (empty playlist after all), which is [not considered engaging](https://github.com/androidx/media/blob/ae9a2339679826f870690092509a135dffc0d7e4/libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java#L231) (and rightfully so), hence there's no attempt to start into foreground and we crash.

Though these are properly logged to logcat, these crashes are hard to debug when just looking at Crashlytics, which is why I think there's potential for improvement here. If we know for sure we're going to crash anyway, we might as well re-throw the exception in the future as cause for some IllegalStateException explaining the problem.

### Media

Not applicable

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

brahmkshatriya:
+1

marcbaechinger:
Thanks for your report. 

>  just don't override it at all

It's documented that when adding the broadcast receiver, then it should be overridden.

> potential for improvement 

Agreed. Marked as improvements accordingly.

We generally plan to improve the playback resumption API at some point. We can then also look into this. I don't think  we should start on this right now as playback resumption is comparably well documented and there is an exception visible in the logcat already when playback resumption is called (according to the report).

PR #2256, opened 2025-03-22
session: fix a deadlock due to onPlayRequested()

nift4:
Issue: #1197

PR #2626, opened 2025-07-17
session: Avoid ForegroundServiceDidNotStartInTimeException...

nift4:
...and throw more useful exceptions instead.

This will however not throw exceptions where foreground service
start is not absolutely required, and keep logging warnings in
these cases instead.

Issue: https://github.com/androidx/media/issues/2591
Issue: https://github.com/androidx/media/issues/2622

(This PR includes PR #2256 because they conflict. It would probably make sense to merge #2256 first.)

### DIFF
--- a/libraries/session/src/main/java/androidx/media3/session/MediaSession.java
+++ b/libraries/session/src/main/java/androidx/media3/session/MediaSession.java
@@ -1932,6 +1932,14 @@ public class MediaSession {
      * automatically as required. Any additional initial setup like setting playback speed, repeat
      * mode or shuffle mode can be done from within this callback.
      *
+     * <p>If the returned list is empty or an exception is returned, and the request to resume
+     * playback came from {@link MediaButtonReceiver}, an {@link IllegalStateException} will be
+     * thrown. This is because the {@link MediaButtonReceiver} has already requested a foreground
+     * service start, and it's not possible to avoid a {@code
+     * ForegroundServiceDidNotStartInTimeException} anymore. To avoid a crash as result, override
+     * {@link MediaButtonReceiver#shouldStartForegroundService(Context, Intent)} to return {@code
+     * false} if there is nothing to resume. This will avoid requesting a foreground service start.
+     *
      * <p>The method will only be called if the {@link Player} has {@link
      * Player#COMMAND_GET_CURRENT_MEDIA_ITEM} and either {@link Player#COMMAND_SET_MEDIA_ITEM} or
      * {@link Player#COMMAND_CHANGE_MEDIA_ITEMS} available.
--- a/libraries/session/src/main/java/androidx/media3/session/MediaSessionImpl.java
+++ b/libraries/session/src/main/java/androidx/media3/session/MediaSessionImpl.java
@@ -1141,7 +1141,9 @@ import org.checkerframework.checker.initialization.qual.Initialized;
    * @param controller The controller requesting to play.
    */
   /* package */ ListenableFuture<SessionResult> handleMediaControllerPlayRequest(
-      ControllerInfo controller, boolean callOnPlayerInteractionFinished) {
+      ControllerInfo controller,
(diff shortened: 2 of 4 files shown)


nift4:
As requested earlier, a brief reasoning of the design choices:
Throwing and giving up with a clear error message seems to be the best we can do, because:
1. there's no way to start the service from a broadcast reciever without the system enforcing a foreground service will start after we tell it to (bindService hence MediaController isn't supported from a broadcast reciever, and startService without foreground will - if I understand correctly - not transfer the foreground allowlist to our service)
2. if the system enforces the service will start after we tell it to, `shouldStartForegroundService()` should be overwritten for every app that supports logging out / empty playlists / etc, but it's not very discoverable
3. the consequence of not doing that is a cryptic ForegroundServiceDidNotStartInTimeException

1 is a hard fact and we can't do anything about it, 2 is more of a documentation problem and it seems external contributors can't edit those anyway, but 3 is possible to address - and it will give a concrete improvement for everyone suffering from ForegroundServiceDidNotStartInTimeException without understanding the reason, as it makes it practically impossible for the library to generate one now. I myself had ForegroundServiceDidNotStartInTimeException in Play Console because I didn't realize shouldStartForegroundService() is required, but with this patch there'd be a useful explainer instead.

I also don't see any disadvantage of this patch for SDK>=31, and I consider the added sanity check on <31 an advantage because developers that don't always test on Android 12+ will still get notified about the fact their code is wrong.

jackyhieu1211:
@marcbaechinger 

Hi mr @marcbaechinger 

I’m currently facing the `ForegroundServiceDidNotStartInTimeException` crash in production, and this PR seems to address it by handling the exception and reporting it to Firebase without crashing the app.

This issue is still happening frequently on Firebase Crashlytics.

I would like to ask:

1. Is this PR considered ready from your side?
2. Is there any plan or timeline to merge it into the main branch?
3. If any additional changes are required before merge, could you share the current status?

At minimum, what I need is that when this exception occurs, it should **not be reported as a crash on Firebase Crashlytics**. This affects real users, so any update would be very helpful.

Thank you!

nift4:
> this PR seems to address it by handling the exception and reporting it to Firebase without crashing the app

No, that is sadly not possible. The root cause would have to be addressed for this, and it is often different for each individual app, there's no general library bug that causes this in 1.9.x AFAIK (previous versions did have bug #2768). What this PR does is instead make the crash clearly tell what the immediate reason for the crash (ie, no media items returned) is, so it is easier to investigate.

marcbaechinger:
Thanks for the insight Nick!

@jackyhieu1211

> this PR seems to address it by handling the exception and reporting it to Firebase without crashing the app

May I ask you a file a bug and explain in a bit more detail what you are seeing? LIke a stack trace, bug report or similar and what your idea is how it's caused?

If this is about a crash in `onPlaybackResumption()` or when adding an empty list from there (as mentioned in the description of this change), then I think we can (want) address this with a less intrusive change. Need to be evaluated, but I think it would make sense to file a bug and add a bit more of information.

jackyhieu1211:
@nift4 @marcbaechinger 

Hi, thanks for the suggestion!

1. I cannot reproduce the issue locally. I only receive this crash from Firebase Crashlytics, so I don’t have a reliable way to trigger it manually.


2. To explain more: recently Google Play has introduced stricter App Quality checks. For apps in the “Play Video” category, the maximum allowed crash rate is around 1.09%.
However, this specific crash alone accounts for 2–5% of all crashes, which means the app is flagged as a low-quality app on Play Store.
Because of that, my intention is simply: if the user hits this issue, the app should return an error instead of crashing completely.


3. From my perspective, this PR solves exactly that: it prevents the hard crash. The logic still reports an error, but the app stays alive.


4. I’ll continue investigating. If I discover anything more—such as steps to reproduce or additional insights—I’ll file a more detailed bug report.



Thanks again!

PR #2633, opened 2025-07-19
Remove the inexistant `lib-datasource-httpengine` module

MGaetan89:
This module doesn't seem to exist, so this commit removes it from the list of loaded modules.


icbaker:
Thanks for the PR - for completeness, this is a clean-up that was missed in https://github.com/androidx/media/commit/250fc80419b4b8ef1b753824e53ebdac176968f5.

Issue #2577, opened 2025-06-30
setMetadata causing ANR across many devices


nift4:
### Version

Media3 pre-release (alpha, beta or RC not in this list)

### More version details

2ca41b03a383b8f0e30c917e9221dbdbf0f31114 (somewhat close to 1.8.0-alpha1)

### Devices that reproduce the issue

According to Play Console:
Pixel 6a running Android 16
Fairphone 4 running Android 13
Samsung a52sxq running Android 14
Redmi Note 10 (sunny) running Android 12

around 0.10% of all users are affected, but my app has a rather small install base of 4850 users, make of that what you will (but I want to avoid reaching the bad behaviour threshold)
### Devices that do not reproduce the issue

any of the devices I have :)

### Reproducible in the demo app?

I'm not sure

### Reproduction steps

I'm not sure, I can't repro this locally. In the code I just call setMediaItems, nothing special.

### Expected result

No ANR

### Actual result

ANR 1: Pixel 6a running Android 16

```
"main" tid=1 Native
  #00  pc 0x00000000000dc848  /apex/com.android.runtime/lib64/bionic/libc.so (__ioctl+8)
  #01  pc 0x000000000007fd7c  /apex/com.android.runtime/lib64/bionic/libc.so (ioctl+156)
  #02  pc 0x000000000006d484  /system/lib64/libbinder.so (android::IPCThreadState::transact+1572)
  #03  pc 0x0000000000066088  /system/lib64/libbinder.so (android::BpBinder::transact+312)
  #04  pc 0x00000000001ed708  /system/lib64/libandroid_runtime.so (android_os_BinderProxy_transact+152)
  at android.os.BinderProxy.transactNative (Native method)
  at android.os.BinderProxy.transact (BinderProxy.java:592)
  at android.media.session.ISession$Stub$Proxy.setMetadata (ISession.java:526)
  at android.media.session.MediaSession.setMetadata (MediaSession.java:523)
  at androidx.media3.session.legacy.MediaSessionCompat$MediaSessionImplApi21.setMetadata (MediaSessionCompat.java:2069)
  at androidx.media3.session.legacy.MediaSessionCompat.setMetadata (MediaSessionCompat.java:538)
  at androidx.media3.session.MediaSessionLegacyStub.setMetadata (MediaSessionLegacyStub.java:976)
  at androidx.media3.session.MediaSessionLegacyStub.access$1100 (MediaSessionLegacyStub.java:110)
[...]

PR #2648, opened 2025-07-24
Move legacy session calls to a new thread

nift4:
Issue: https://github.com/androidx/media/issues/2577
Test: add Thread.sleep(5000) to setMetadata(), app stays responsive

### DIFF
--- a/libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
+++ b/libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
@@ -471,20 +471,6 @@ import java.util.concurrent.TimeoutException;
     }
 
     @Override
-    public void onMediaButtonPreferencesChanged(
-        MediaController controller, List<CommandButton> mediaButtonPreferences) {
-      mediaSessionService.onUpdateNotificationInternal(
-          session, /* startInForegroundWhenPaused= */ false);
-    }
-
-    @Override
-    public void onAvailableSessionCommandsChanged(
-        MediaController controller, SessionCommands commands) {
-      mediaSessionService.onUpdateNotificationInternal(
-          session, /* startInForegroundWhenPaused= */ false);
-    }
-
-    @Override
     public ListenableFuture<SessionResult> onCustomCommand(
         MediaController controller, SessionCommand command, Bundle args) {
       @SessionResult.Code int resultCode = SessionError.ERROR_NOT_SUPPORTED;
@@ -503,20 +489,6 @@ import java.util.concurrent.TimeoutException;
       // We may need to hide the notification.
       mediaSessionService.onUpdateNotificationInternal(
           session, /* startInForegroundWhenPaused= */ false);
-    }
-
-    @Override
-    public void onEvents(Player player, Player.Events events) {
-      // We must limit the frequency of notification updates, otherwise the system may suppress
-      // them.
-      if (events.containsAny(
-          Player.EVENT_PLAYBACK_STATE_CHANGED,
-          Player.EVENT_PLAY_WHEN_READY_CHANGED,
-          Player.EVENT_MEDIA_METADATA_CHANGED,
-          Player.EVENT_TIMELINE_CHANGED)) {
-        mediaSessionService.onUpdateNotificationInternal(
-            session, /* startInForegroundWhenPaused= */ false);
-      }
     }
   }
 
--- a/libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
+++ b/libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
@@ -760,6 +760,7 @@ import java.util.concurrent.atomic.AtomicInteger;
         Log.e(TAG, "Failed to start foreground", e);
         return false;
       }
+      throw e;
     }
     if (hasOtherForegroundSession) {
       // startForeground() normally does this, but as we passed another notification to it, we have
--- a/libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
+++ b/libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
@@ -840,7 +840,7 @@ import java.util.concurrent.atomic.AtomicInteger;
   }
 
   @RequiresApi(31)
-  /* package */ static final class Api31 {
+  private static final class Api31 {
     public static boolean instanceOfForegroundServiceStartNotAllowedException(
         IllegalStateException e) {
       return e instanceof ForegroundServiceStartNotAllowedException;
--- a/libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
+++ b/libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
@@ -79,11 +79,17 @@ import java.util.concurrent.TimeoutException;
   private final Intent startSelfIntent;
   private final String startSelfIntentUid;
   private final Map<MediaSession, ControllerInfo> controllerMap;
+  private final Map<MediaSession, MediaNotification> mediaNotifications;
 
   private MediaNotification.Provider mediaNotificationProvider;
-  private int totalNotificationCount;
-  @Nullable private MediaNotification mediaNotification;
-  private boolean startedInForeground;
+
+  /**
+   * The session that provides the notification the foreground service is started with, or null if
+   * the service is not in foreground at the moment. Each service can only have one foreground
+   * notification, so we have to switch this around as needed if we have multiple sessions.
+   */
+  private @Nullable MediaSession foregroundSession;
+
   private boolean isUserEngaged;
   private boolean isUserEngagedTimeoutEnabled;
   private long userEngagedTimeoutMs;
@@ -106,7 +112,7 @@ import java.util.concurrent.TimeoutException;
     startSelfIntentUid = UUID.randomUUID().toString();
     startSelfIntent.putExtra(SELF_INTENT_UID_KEY, startSelfIntentUid);
     controllerMap = new HashMap<>();
-    startedInForeground = false;
+    mediaNotifications = new HashMap<>();
     isUserEngagedTimeoutEnabled = true;
     userEngagedTimeoutMs = MediaSessionService.DEFAULT_FOREGROUND_SERVICE_TIMEOUT_MS;
     showNotificationForIdlePlayerMode =
@@ -155,6 +161,9 @@ import java.util.concurrent.TimeoutException;
   public void removeSession(MediaSession session) {
     @Nullable ControllerInfo controllerInfo = controllerMap.remove(session);
     if (controllerInfo != null) {
+      // Update the notification count so that if a pending notification callback arrives (e.g., a
+      // bitmap is loaded), we don't show a stale notification.
+      controllerInfo.notificationSequence++;
       MediaController.releaseFuture(controllerInfo.controllerFuture);
     }
   }
@@ -195,18 +204,21 @@ import java.util.concurrent.TimeoutException;
     return CallbackToFutureAdapter.getFuture(
         completer -> {
           if (!mediaSessionService.isSessionAdded(session) || !shouldShowNotification(session)) {
-            removeNotification();
+            removeNotification(session);
             completer.set(null);
             return "notificationRemoved";
           }
-          int notificationSequence = ++totalNotificationCount;
+
+          ControllerInfo controllerInfo = checkNotNull(controllerMap.get(session));
+          int notificationSequence = ++controllerInfo.notificationSequence;
           ImmutableList<CommandButton> mediaButtonPreferences =
               checkNotNull(getConnectedControllerForSession(session)).getMediaButtonPreferences();
           MediaNotification.Provider.Callback callback =
               notification ->
                   mainExecutor.execute(
-                      () -> onNotificationUpdated(notificationSequence, session, notification));
-
+                      () ->
+                          onNotificationUpdated(
+                              controllerInfo, notificationSequence, session, notification));
           Util.postOrRun(
               session.getImpl().getApplicationHandler(),
               () -> {
@@ -242,7 +254,7 @@ import java.util.concurrent.TimeoutException;
   }
 
   public boolean isStartedInForeground() {
-    return startedInForeground;
+    return foregroundSession != null;
   }
 
   public void setUserEngagedTimeoutMs(long userEngagedTimeoutMs) {
@@ -316,8 +328,11 @@ import java.util.concurrent.TimeoutException;
   }
 
   private void onNotificationUpdated(
-      int notificationSequence, MediaSession session, MediaNotification mediaNotification) {
-    if (notificationSequence == totalNotificationCount) {
+      ControllerInfo controllerInfo,
+      int notificationSequence,
+      MediaSession session,
+      MediaNotification mediaNotification) {
+    if (controllerInfo.notificationSequence == notificationSequence) {
       boolean startInForegroundRequired =
           shouldRunInForeground(/* startInForegroundWhenPaused= */ false);
       try {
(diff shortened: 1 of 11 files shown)


tonihei:
Thanks for providing the PR - I probably need a bit of time to review it in more detail. 

Some comments from a quick glance:
 - There is blocking call in `onMediaButtonEvent` which we shouldn't keep if the goal is to remove all potentially blocking calls
 - Introducing the new thread on this level in `MediaSessionLegacyStub` makes it harder to reason about the flow as we suddenly need to account for the possibility of more than one thread in the same class. I wonder if it would be much cleaner to handle the threading in its own dedicated place (e.g. inside `MediaSessionCompat`?). 
 - Eventually, all of the binder calls triggered from the media session module (including the media3-only ones!) should be done from a shared background thread. We can  start with more isolated cases like the calls to the platform MediaSession, but that also means it's fine to create the background thread high-up in the chain if needed.
 - One potential concern is about delicate timing issues  - setting the platform media session state and updating the notification for example may now happen in a slightly different order. The underlying risk is that some other logic may accidentally rely on a certain execution order at the moment and break. This kind of issue is really hard to predict or avoid, but just noting that this is one of the reasons I haven't touched this part yet.

nift4:
Hi,

> There is blocking call in onMediaButtonEvent which we shouldn't keep if the goal is to remove all potentially blocking calls

It blocks the media session compat interaction thread, not the main thread, so it shouldn't be too much of a problem (may delay media button clicks but ig we're processing an earlier event, we have to delay anyway).

> Introducing the new thread on this level in MediaSessionLegacyStub makes it harder to reason about the flow as we suddenly need to account for the possibility of more than one thread in the same class. I wonder if it would be much cleaner to handle the threading in its own dedicated place (e.g. inside MediaSessionCompat?).

Yes, the callbacks from legacy session side (onMediaButtonEvent, onPlay, etc) are called on the legacy session interaction thread, while the ControllerLegacyCbForBroadcast callbacks are called on session application thread. But we need to know when setMetadata is done in order to make sure notification update is only posted after that.

> Eventually, all of the binder calls triggered from the media session module (including the media3-only ones!) should be done from a shared background thread. We can start with more isolated cases like the calls to the platform MediaSession, but that also means it's fine to create the background thread high-up in the chain if needed.

I think we should start with individual threads for now and revisit later.

> One potential concern is about delicate timing issues - setting the platform media session state and updating the notification for example may now happen in a slightly different order. The underlying risk is that some other logic may accidentally rely on a certain execution order at the moment and break. This kind of issue is really hard to predict or avoid, but just noting that this is one of the reasons I haven't touched this part yet.

I sidestepped this by not reacting to events in the media notification controller and instead call onNotificationRequestRequired() when calls like setMetadata are actually done.

I experimented with delaying media3 controller calls until platform session is done but it added 500 lines to the diff and caused hard-to-reproduce crashes, so I dropped it.

nift4:
rebased and pushed a small patch to be less spammy with notification updates if multiple events happen in one Looper iteration, similar to how onEvents() worked

nift4:
Hi @tonihei, I noticed I had not really addressed any of the review comments properly before, sorry about that. I now uploaded a refactored version, could you please take a look again? I'm trying to get my huge stack of patches upstreamed (or otherwise the need resolved), because it takes quite some time to maintain. Thanks a lot!

> There is blocking call in onMediaButtonEvent which we shouldn't keep if the goal is to remove all potentially blocking calls

This is now gone because the callback thread changed to session thread.

> Introducing the new thread on this level in MediaSessionLegacyStub makes it harder to reason about the flow as we suddenly need to account for the possibility of more than one thread in the same class. I wonder if it would be much cleaner to handle the threading in its own dedicated place (e.g. inside MediaSessionCompat?).

It's moved to MediaSessionCompat with ListenableFuture in the few places where we need to wait for it to be complete, and you're right, it looks a lot better.

> Eventually, all of the binder calls triggered from the media session module (including the media3-only ones!) should be done from a shared background thread. We can start with more isolated cases like the calls to the platform MediaSession, but that also means it's fine to create the background thread high-up in the chain if needed.

Introduced a backgroundThread in MediaSessionImpl for that purpose.

> One potential concern is about delicate timing issues - setting the platform media session state and updating the notification for example may now happen in a slightly different order. The underlying risk is that some other logic may accidentally rely on a certain execution order at the moment and break. This kind of issue is really hard to predict or avoid, but just noting that this is one of the reasons I haven't touched this part yet.

I had shipped the previous version of this patch in beta channel for months without issues now. :) At least the subset of media3 that my app uses appears to be fine.

libraries/session/src/main/java/androidx/media3/session/MediaLibrarySessionImpl.java
```diff
@@ -360,19 +359,6 @@ protected MediaSessionServiceLegacyStub createLegacyBrowserService(
     return stub;
   }
 
-  @Override
-  protected void dispatchRemoteControllerTaskWithoutReturn(RemoteControllerTask task) {
```
tonihei: There seem to be some unrelated changes in here, like this one. I agree that this method is pointless as far as I can see, but it has nothing to do with the threading update. I'll remove this one separately, no need to send a PR.

libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
```diff
@@ -405,20 +405,6 @@ public void onConnected(boolean shouldShowNotification) {
       }
     }
 
-    @Override
-    public void onMediaButtonPreferencesChanged(
```
tonihei, line 483: This is likely unrelated or a merge conflict. Could you restore the removed methods in this class?

libraries/session/src/main/java/androidx/media3/session/MediaSessionLegacyStub.java
```diff
@@ -869,65 +874,60 @@ private void dispatchSessionTaskWithPlayerCommand(
       Log.d(TAG, "RemoteUserInfo is null, ignoring command=" + command);
       return;
     }
-    postOrRun(
```
tonihei: Maybe I'm missing something, but why did you remove the postOrRun call in this method? It seems to me that the call sites are the platform callbacks that are now explicitly called on the internal background thread. So shouldn't the callback then post back to the application thread where we can safely access the player state and run updates?

libraries/session/src/main/java/androidx/media3/session/MediaSessionLegacyStub.java
```diff
@@ -1222,30 +1222,33 @@ private static void sendCustomCommandResultWhenReady(
...
-  private static void setQueue(MediaSessionCompat sessionCompat, @Nullable List<QueueItem> queue) {
-    sessionCompat.setQueue(queue);
+  private static <T> void ignoreFuture(Future<T> unused) {
+    // no-op
   }
 
-  @SuppressWarnings("nullness:argument") // MediaSessionCompat didn't annotate @Nullable.
-  private void setQueueTitle(MediaSessionCompat sessionCompat, @Nullable CharSequence title) {
-    sessionCompat.setQueueTitle(isQueueEnabled() ? title : null);
+  private <T> void requestNotificationRefreshWhenDone(
```
tonihei: I'm slightly unsure why this method is needed in so many places. As far as I can see, we only required it previously once the lazily loaded Bitmap is set on the platform session (to tell SysUI or others to refresh their state). Now it seems this request is triggered from lots of other places as well.

libraries/session/src/main/java/androidx/media3/session/legacy/MediaSessionCompat.java
```diff
@@ -422,8 +438,8 @@ public void setFlags(@SessionFlags int flags) {
    *
    * @param stream The {@link AudioManager} stream this session is playing on.
    */
-  public void setPlaybackToLocal(int stream) {
-    impl.setPlaybackToLocal(stream);
+  public ListenableFuture<Object> setPlaybackToLocal(int stream) {
```
tonihei: Potentially related to my other  question about why the notification update is needed in so many places: I'm unsure why many of the methods with a `null` return value suddenly need a `ListenableFuture`. I guess the only reason is to know when the call has been made so subsequent actions like the request for a new notification can be triggered? If so, this may only be needed for the setMetadata call as mentioned in the other comment. 

For the ones that are still needed, it's best to return a ListenableFuture<Void> explicitly since we are not expecting any actual returned Object besides null.

tonihei (commented):
Thanks for the update - starts looking really great! I left a few comments about things I'm unsure about, but the general shape makes sense.

libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
nift4: Hi Toni, no this wasn't unrelated nor a merge conflict. It's actually related to the `requestNotificationRefreshWhenDone`, previously, platform session was updated in player thread before notification manager got any event. But now platform session is updated in background thread so MediaNotificationManager would actually trigger the notification update too early, which causes out of sync data, and caused actual buggy notification. This is a bit counterintuitive, but SystemUI requests new data from platform session once new notification is posted. So the ordering must be first update platform session, wait until the binder call is done, then post notification. That is the reason why media notification manager no longer has this responsibility and it's moved to legacy stub with `requestNotificationRefreshWhenDone`. Hope that clears it up.

libraries/session/src/main/java/androidx/media3/session/MediaSessionLegacyStub.java
nift4: Sorry, I thought I had added it in the first place in the pre-refactor version. I went through all the call sites and postOrRun was unneeded, so I removed it.

> It seems to me that the call sites are the platform callbacks that are now explicitly called on the internal background thread.

No, the platform callbacks are still on player thread directly. That helped a lot for simplyfying threading and doesn't cause any blocking calls (I've checked in platform source code), so it seems to be fine.

libraries/session/src/main/java/androidx/media3/session/MediaSessionLegacyStub.java
nift4: See above, it's because this class has some of responsibility that MediaNotificationManager previously had.

libraries/session/src/main/java/androidx/media3/session/legacy/MediaSessionCompat.java
nift4: > I guess the only reason is to know when the call has been made so subsequent actions like the request for a new notification can be triggered?

Yes, to avoid race condition between notification and platform session data being set. We need to wait for blocking binder call into platform session to finish before posting notification.

> If so, this may only be needed for the setMetadata call as mentioned in the other comment.

See above on why it's all of these.

> it's best to return a ListenableFuture explicitly since we are not expecting any actual returned Object besides null.

You mean `ListenableFuture<?>` or without generic so really just `ListenableFuture`? (I _think_ it was not recommended to use generic with erased type, so I just chose a dummy object for now, but I'm happy to change to something else)

libraries/session/src/main/java/androidx/media3/session/MediaLibrarySessionImpl.java
nift4: I was going through all the dispatch methods to ensure I didn't miss any blocking call or have something on wrong thread, and stumbled onto this, so I removed it. Intended to mention during review but it seems forgot. I'll rebase the PR once the removal is pushed so that the unrelated change is out of the diff. :)

tonihei: Removed in https://github.com/androidx/media/commit/97feef2c816856ae559ae7add40ee5b120152d4b

libraries/session/src/main/java/androidx/media3/session/legacy/MediaSessionCompat.java
tonihei: You can use `ListenableFuture<Void>`  to only allow `null` as a return value. This somehow got lost in the markdown formatting of my original reply :)

libraries/session/src/main/java/androidx/media3/session/MediaNotificationManager.java
tonihei: Thanks for the explanation! 

That's the kind of thing I was referring to when I noted the potential for "delicate timing issues"...  Returning a Future on some of these operations under the assumption that a controller uses a notification listener to listen to changes in this class is not ideal either as it bakes in a lot of implementation detail of System UI unfortunately.  It also means we almost send twice as many notification updates now I believe (once the state changes and then again when the Future returns).

I'm actually surprised this always works at the moment given the notification update and the media session update are handled independently, but I guess because both are sent to the system server and System UI only requests the MediaSession state update after it received yet another callback, it just works. 

I wonder if the most consistent way to solve this is to move the `MediaController` used by `MediaNotificationManager` to the same background thread. Then the state updates that trigger the notification are sent to the same thread that updates the platform MediaSession, ensuring consistent ordering again and no duplicate updates. It probably also helps to move these notification calls off the main thread anyway. Would you be willing to look into this as well? If not, we can try to implement this on our side, but need to see when we can look into this exactly.

nift4: > Would you be willing to look into this as well?

I'm working on it, and I'll finish it over the holidays, but I assume you'll be on leave for a while. So, happy holidays :)

nift4:
Hi @tonihei, notification handling is now entirely moved to session's background thread.

While doing that, I had a few obstacles, like MediaNotificationManager having some bugs related to multiple session. In order to write proper code that does not add new bugs I had to fix those, so diff got a bit large. (I'm happy to iterate on this, or go back to the previous approach, whatever you prefer.) Also note that unlike before moving from main thread, now setAvailableCommands() still need the future based callback because processing requires both player and background thread.

Please review again and thanks in advance!

nift4:
Hi @tonihei, I uploaded the changes you had requested before and would appreciate the next review pass :) Thanks!

tonihei:
Just to confirm, I've seen your comment, but haven't got around to look into this in more detail again.

nift4:
> Just to confirm, I've seen your comment, but haven't got around to look into this in more detail again.

Thanks for the heads-up!

(Pushed to fix the conflicts.)

nift4:
Hi @tonihei,

there is a time sensitive chance right now: In https://github.com/androidx/media/commit/c8d9a096e83ee3424b5e8cfd89ec6270c39e3bf0 we got a new API specifically "to enable the default implementation of the library to provide multi thread scenarios".

Currently, this method must also be run in the main thread, but it can return a future.

https://github.com/androidx/media/blob/c6779dd3aff9c085158d68f9ff5a402f215bb1b3/libraries/session/src/main/java/androidx/media3/session/MediaSessionService.java#L843-L860

This is already great for the purpose of moving notification to background thread as we discussed before. Right now in this PR I have a very similar method which also returns a future, but I also have the method allowed to be called from any thread by the library, because it would be called in response to media notification controller which is on session background thread, and also runs the logic on the same session background thread in my design.

However, if we would use the method definition from main branch, there would still be useless call to main thread just to post to background thread from there.

Right now, though, there is the chance, because this method was not yet in any release, to change the definition to allow the library to call it in any thread. We then can independently merge this PR to actually make use of that capability, or we can during review decide it should not go on the background thread after all and change the definition from "may be called on any thread" to "will be called on main thread" (as that direction is, after all, a backwards compatible change).

We can do this as long as there is no new release including this commit, I believe.

What do you think? Is this a good idea?

(Btw, I will take care of the conflicts in this PR soon.)

nift4:
Hello @tonihei. Once again, I would like to bring above proposal to change onUpdateNotificationAsync() to your attention, as I believe the next release branch is not that far off, which would make it too late to look into changing the API definition seamlessly. Handling notifications on a background thread (as you proposed earlier in this thread) requires running onNotificationUpdateAsync() on a background thread, and it's a good opportunity to change this now as it has been introduced recently. Thank you!

nift4:
Hi @tonihei, should I keep rebasing the PR once in a while, or is it unlikely that you'll have time for it soon?

PR #2655, opened 2025-07-25
Ensures errors are correctly attributed to ad or content

stevemayhew:
This change is part of a fix to the issue "Player Errors during Ad playback not attributed correctly"

The change introduced in commit 7cf2fd9 was a partial fix to allow ad playback errors to be reported. It delayed removing the listener when an error released the AdsMediaSource.

What remained was the larger issue: When an error occurs **during content preparation while an ad is still playing**, ExoPlayer cannot correctly attribute the error to the content source. Instead, the error bubbles up through `ExoPlayerImplInternal.stopInternal()` and causes the **entire CompositeMediaSource (AdsMediaSource + content)** to be torn down.

This results in the `ImaAdsLoader` being stopped prematurely — even when the ad is not at fault — and breaks proper ad lifecycle handling.

**User-Facing Issues**

1. **The ad is incorrectly marked as failed** A content preparation error is misattributed to the currently playing ad, resulting in an invalid `AD_STATE_ERROR`.
2. **The ad is never marked as completed** Even though `ALL_ADS_COMPLETED` is dispatched by IMA, the player is torn down before ExoPlayer updates `AdPlaybackState` to reflect completion.
3. **A single failing ad in a pod is not skipped** If one ad in a pod fails (e.g., due to a 404 or DRM error), the player stops entirely instead of skipping to the next ad in the group.
4. **A failing ad does not fall back to content** When an ad fails, the player should skip to content. Instead, the entire playback pipeline is torn down via `stopInternal()`.

This change fixes the first two User-Facing issues.  When combined with error recovery logic it also fixes the 4th issue. The 3rd issue may require some work on the IMA SDK side, the mid-pod failure aborts all ad playback with this fix, rather than skipping just the failing ad.

### DIFF
--- a/libraries/exoplayer_ima/src/main/java/androidx/media3/exoplayer/ima/AdTagLoader.java
+++ b/libraries/exoplayer_ima/src/main/java/androidx/media3/exoplayer/ima/AdTagLoader.java
@@ -391,7 +391,23 @@ import java.util.Objects;
   /** Deactivates playback. */
   public void deactivate() {
     Player player = checkNotNull(this.player);
-    if (!AdPlaybackState.NONE.equals(adPlaybackState) && imaPausedContent) {
+    // Post release of listener behind any already queued Player.Listener events to ensure that
+    // any pending events are processed before the player is deferred.
+    handler.post(
+        () -> {
+          deactivateInternal(player);
+        });
+  }
+
+  /**
+   * Deactivates playback internally, after the Listener.onEvents() cycle completes so the complete
+   * state change picture is clear. For example, if an error caused the deactivation, need to
+   * determine whether the error was for an ad or content.
+   */
+  private void deactivateInternal(Player player) {
+    if (!AdPlaybackState.NONE.equals(adPlaybackState)
+        && imaPausedContent
+        && player.getPlayerError() == null) {
       if (adsManager != null) {
         adsManager.pause();
       }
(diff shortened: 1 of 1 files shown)


stevemayhew:
Note a branch with extensive logging and a partial solution for User Facing Issue #4 is in my branch from this change: https://github.com/stevemayhew/media/tree/p-fixes-ad-error-handling-logging

tonihei:
Thanks for providing the fix, only now found the time to have a closer look. The fix looks correct to me, I'm going to merge it internally.

libraries/exoplayer_ima/src/main/java/androidx/media3/exoplayer/ima/AdTagLoader.java
```diff
@@ -391,7 +391,20 @@ public void activate(Player player) {
...
+
+  /**
+   * Deactivates playback internally, after the Listener.onEvents() cycle completes so the complete
+   * state change picture is clear. For example, if an error caused the deactivation, the error
+   * callback can be handled first.
+   */
+  private void deactivateInternal(Player player) {
+    if (!adPlaybackState.equals(AdPlaybackState.NONE)
+        && imaPausedContent
+        && player.getPlayerError() == null) {
```
tonihei, line 407: @stevemayhew Could you explain why this additional check is needed again? What would go wrong with marking the adsmanager as paused and storing the ad resume position in case of an error in the player?

Issue #770, opened 2023-10-26
Playback resumption - Unexpected dead silent notification with stop() method invoked on MediaController


TomasValenta:
### Version

Media3 1.1.1

### Devices that reproduce the issue

* Pixel 7 running Android 13
* Samsung Galaxy S21+ running Android 13

### Reproducible in the demo app?

Yes

### Reproduction steps

1. Checkout [my demo-session fork](https://github.com/TomasValenta/media) with the **MediaButtonReceiver** declared in the manifest and simple stop button added to the PlayableFolderActivity ([full diff here](https://github.com/androidx/media/compare/release...TomasValenta:media:release)).
2. Build and run the **demo-session** sample.
3. Play any song (or songs) from the example playlist.
4. Go back to the **PlayableFolderActivity** and hit **STOP** button.
5. A **dead silent notification** will appear.

### Expected result

**Dead silent notification** should not be displayed (as it behaves when no MediaButtonReceiver is declared).

I know, I can handle the play event from the dead silent notification in the onPlaybackResumption() method (and that's how I do it too), but if I understand it correctly, this notification shouldn't show at all after calling stop() method on MediaController or not? (as it behaves when no MediaButtonReceiver is declared)

### Actual result

A **dead silent notification** will appear.

### Media

 My fork - https://github.com/TomasValenta/media

Dead silent notification
<img width="602" alt="image" src="https://github.com/androidx/media/assets/4901251/b9025345-29af-41c8-991c-cfa1876cf34d">

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

marcbaechinger:
Thanks for the detailed report and the text fixture.

The notification in the screenshot is not posted by Media3. This is the playback resumption notification posted by system UI (see for instance the ExoPlayer logo as the app icon at the top left, the demo app uses a different icon). After calling `stop`, the player transitions into `STATE_IDLE` which is mapped to the platform session that system UI reads.

When we do a `adb shell dumpsys media_session` in this state,  then we see the following state of the platform session of the demo app:

```
androidx.media3.session.id. androidx.media3.demo.session/androidx.media3.session.id. (userId=0)
[...]
      state=PlaybackState {state=0, position=0, buffered position=0, speed=0.0, updated=479973, actions=7339655, custom actions=[Action:mName='Disable shuffle mode, mIcon=2131230888, mExtras=Bundle[EMPTY_PARCEL]], active item id=0, error=null}
[...]

PR #2670, opened 2025-07-27
Add notification for empty player setting

nift4:


tonihei:
Thanks for the idea. How would a notification for the empty player look like? It seems to me that basically all relevant and required information is missing at this point unless there is a custom `MediaNotification.Provider`?

nift4:
> How would a notification for the empty player look like? It seems to me that basically all relevant and required information is missing at this point unless there is a custom MediaNotification.Provider?

In its most basic form, it looks like this:
grafik
Notably, the title "Gramophone is running" is supplied by the system because the app doesn't send any. On older versions, it'd display empty text instead.

Indeed, rather useless/irrelevant on its own.

The reason I am interested in it at the moment is stupid hacks, see https://github.com/androidx/media/issues/770#issuecomment-3124666502. But that is not the only reason I want this.

On my not-too-far-future roadmap, there's an item about more support for session-based cast protocols (it's not what I'm actually working with, but you can think of Miracast as example). Etablishing a connection takes considerable effort (confirmation on both devices) and time, hence it was decided a user should generally manually terminate a session (for example, by pressing the custom close command button in the media notification) instead of the app ending the session when there is no media currently set for whatever reason. To remind the user there still is casting session going on, there needs to be some sort of notification. Making the media notification disappear when playlist is empty just to show a casting session notification which isn't MediaStyle, which then disappears again and having a media notification come up when playlist stops being empty, is poor UX. Instead, the media notification should just signify that it is empty. This will be done with a ForwardingSimpleBasePlayer that modifies getMediaMetadata() to show a relevant message. (A custom MediaNotification.Provider is not necessary in this case.) For example:
```kotlin
    override fun getState(): State {
        if (player.currentTimeline.isEmpty) {
            val superState = super.state
            return superState.buildUpon().setPlaylist(superState.timeline, superState.currentTracks,
                MediaMetadata.Builder().setTitle("Nothing here...").build()).build()
        }
        return super.getState()
    }
```
The play button can also be added/removed as needed by changing available commands.

Issue #2675, opened 2025-07-27
ForwardingPlayer removeListener() doesn't work if Player uses object sameness instead of equality


nift4:
### Version

Media3 main branch

### More version details

bd2b1a4b8d0aae8f60ba38e00e00bf4ca6a156f5

### Devices that reproduce the issue

Android 16 emulator

### Reproducible in the demo app?

Yes

### Reproduction steps

If a Player implements removeListeners() by using == operator, it will never be possible to remove a listener using ForwardingPlayer.

### Expected result

ForwardingPlayer doing bookkeeping to be able to remove the listener

### Actual result

Listener never gets removed because a new object is created:
```
public void removeListener(Listener listener) {
    player.removeListener(new ForwardingListener(this, listener));
  }
```

### Media

Not applicable

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

icbaker:
Thanks for raising this - I agree it's a bug. I think it's probably wrong that `ForwardingPlayer` tries to rely on `Object.equals()` at all, and we should assume that all `Player` implementations do listener registration/unregistration by reference equality (and arguably by extension: That `Player.Listener` impls shouldn't implement `equals()`).

icbaker:
This is fixed by https://github.com/androidx/media/commit/55eaa00d98c708dd3014fec54bda824d448fe175

Issue #2674, opened 2025-07-27
ExoPlayer in ForwardingSimpleBasePlayer triggers multiple asserts


nift4:
### Version

Media3 main branch

### More version details

bd2b1a4b8d0aae8f60ba38e00e00bf4ca6a156f5

### Devices that reproduce the issue

Android 16 emulator

### Devices that do not reproduce the issue

N/A

### Reproducible in the demo app?

Not tested

### Reproduction steps

1. Wrap ExoPlayer in ForwardingSimpleBasePlayer
2. Play a file ExoPlayer cannot process (i.e. SID file [Rapid_Fire.sid.zip](https://github.com/user-attachments/files/21455467/Rapid_Fire.sid.zip))
3. Play another file ExoPlayer does support

### Expected result

it works

### Actual result

```
Caused by: java.lang.IllegalArgumentException: Player error only allowed in STATE_IDLE
                                                                                                     at androidx.media3.common.util.Assertions.checkArgument(Assertions.java:55)
                                                                                                     at androidx.media3.common.SimpleBasePlayer$State.<init>(SimpleBasePlayer.java:1048)
                                                                                                     at androidx.media3.common.SimpleBasePlayer$State.<init>(SimpleBasePlayer.java:99)
                                                                                                     at androidx.media3.common.SimpleBasePlayer$State$Builder.build(SimpleBasePlayer.java:827)
                                                                                                     at androidx.media3.common.ForwardingSimpleBasePlayer.getState(ForwardingSimpleBasePlayer.java:187)
                                                                                                     at org.akanework.gramophone.logic.utils.exoplayer.EndedWorkaroundPlayer.getState(EndedWorkaroundPlayer.kt:66)
                                                                                                     at androidx.media3.common.SimpleBasePlayer.updateStateForPendingOperation(SimpleBasePlayer.java:3878)
                                                                                                     at androidx.media3.common.SimpleBasePlayer.updateStateForPendingOperation(SimpleBasePlayer.java:3864)
                                                                                                     at androidx.media3.common.SimpleBasePlayer.setMediaItemsInternal(SimpleBasePlayer.java:2313)
                                                                                                     at androidx.media3.common.SimpleBasePlayer.setMediaItems(SimpleBasePlayer.java:2300)
                                                                                                     at androidx.media3.common.ForwardingPlayer.setMediaItems(ForwardingPlayer.java:111)
[...]

PR #2676, opened 2025-07-27
Add listener bookkeeping to ForwardingPlayer

nift4:
Issue: https://github.com/androidx/media/issues/2675

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
+++ b/libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
@@ -20,6 +20,7 @@ import android.view.Surface;
 import android.view.SurfaceHolder;
 import android.view.SurfaceView;
 import android.view.TextureView;
+import androidx.annotation.GuardedBy;
 import androidx.annotation.Nullable;
 import androidx.media3.common.text.Cue;
 import androidx.media3.common.text.CueGroup;
@@ -55,7 +56,9 @@ import java.util.List;
 public class ForwardingPlayer implements Player {
 
   private final Player player;
-  private final IdentityHashMap<Listener, Listener> listeners = new IdentityHashMap<>();
+
+  @GuardedBy("listeners")
+  private final IdentityHashMap<Listener, ForwardingListener> listeners = new IdentityHashMap<>();
 
   /** Creates a new instance that forwards all operations to {@code player}. */
   public ForwardingPlayer(Player player) {
@@ -81,7 +84,7 @@ public class ForwardingPlayer implements Player {
   @Override
   public void addListener(Listener listener) {
     synchronized (listeners) {
-      Listener forwardingListener = listeners.get(listener);
+      ForwardingListener forwardingListener = listeners.get(listener);
       if (forwardingListener == null) {
         forwardingListener = new ForwardingListener(this, listener);
       }
--- a/libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
+++ b/libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
@@ -25,6 +25,7 @@ import androidx.media3.common.text.Cue;
 import androidx.media3.common.text.CueGroup;
 import androidx.media3.common.util.Size;
 import androidx.media3.common.util.UnstableApi;
+import java.util.IdentityHashMap;
 import java.util.List;
 
 /**
@@ -54,6 +55,7 @@ import java.util.List;
 public class ForwardingPlayer implements Player {
 
   private final Player player;
+  private final IdentityHashMap<Listener, Listener> listeners = new IdentityHashMap<>();
 
   /** Creates a new instance that forwards all operations to {@code player}. */
   public ForwardingPlayer(Player player) {
@@ -78,7 +80,14 @@ public class ForwardingPlayer implements Player {
    */
   @Override
   public void addListener(Listener listener) {
-    player.addListener(new ForwardingListener(this, listener));
+    synchronized (listeners) {
+      Listener forwardingListener = listeners.get(listener);
+      if (forwardingListener == null) {
+        forwardingListener = new ForwardingListener(this, listener);
+      }
+      player.addListener(forwardingListener);
+      listeners.put(listener, forwardingListener);
+    }
   }
 
   /**
@@ -90,7 +99,12 @@ public class ForwardingPlayer implements Player {
    */
   @Override
   public void removeListener(Listener listener) {
-    player.removeListener(new ForwardingListener(this, listener));
+    synchronized (listeners) {
+      Listener forwardingListener = listeners.remove(listener);
+      // If forwardingListener is null, we don't know this listener. Just pass in the real listener
+      // as the underlying player would not know it either. It can decide to throw or ignore.
+      player.removeListener(forwardingListener != null ? forwardingListener : listener);
+    }
   }
 
   /** Calls {@link Player#setMediaItems(List)} on the delegate. */
@@ -1069,28 +1083,6 @@ public class ForwardingPlayer implements Player {
     @Override
     public void onDeviceVolumeChanged(int volume, boolean muted) {
       listener.onDeviceVolumeChanged(volume, muted);
-    }
-
-    @Override
-    public boolean equals(@Nullable Object o) {
-      if (this == o) {
-        return true;
-      }
-      if (!(o instanceof ForwardingListener)) {
-        return false;
-      }
-      ForwardingListener that = (ForwardingListener) o;
-      if (!forwardingPlayer.equals(that.forwardingPlayer)) {
-        return false;
-      }
-      return listener.equals(that.listener);
-    }
-
-    @Override
-    public int hashCode() {
-      int result = forwardingPlayer.hashCode();
-      result = 31 * result + listener.hashCode();
-      return result;
     }
   }
 }
--- a/libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
+++ b/libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
@@ -201,7 +201,7 @@ public class ForwardingPlayerTest {
     public void addListener(Listener listener) {
       for (Listener listener1 : listeners) {
         if (listener == listener1) {
-          throw new IllegalArgumentException("Trying to add duplicate listener added");
+          throw new IllegalArgumentException("Trying to add duplicate listener");
         }
       }
       listeners.add(listener);
--- a/libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
+++ b/libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
@@ -23,14 +23,18 @@ import static androidx.media3.test.utils.TestUtil.assertForwardingClassForwardsA
 import static androidx.media3.test.utils.TestUtil.assertSubclassOverridesAllMethods;
 import static androidx.media3.test.utils.TestUtil.getInnerClass;
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertThrows;
 import static org.mockito.ArgumentMatchers.same;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 
+import androidx.annotation.Nullable;
 import androidx.media3.test.utils.StubPlayer;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.common.collect.ImmutableSet;
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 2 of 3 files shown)


libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
```diff
@@ -78,7 +81,9 @@ public Looper getApplicationLooper() {
    */
   @Override
   public void addListener(Listener listener) {
-    player.addListener(new ForwardingListener(this, listener));
+    ForwardingListener forwardingListener = new ForwardingListener(this, listener);
+    listeners.put(listener, forwardingListener);
```
icbaker: `Player.addListener()` can be called on any thread, so you need some concurrency protection here.

icbaker: You probably want to check whether the map already contains `key = listener` or not - otherwise the following operations go wrong:

```java
// Adds listener -> forwardingListener1 mapping & passes
// forwardingListener1 to underlying player.
forwardingPlayer.addListener(listener);

// Replaces listener -> forwardingListener1 mapping with
// listener -> forwardingListener2 and passes
// forwardingListener1 to underlying player.
forwardingPlayer.addListener(listener);

// Removes listener -> forwardingListener2 mapping
// and removes forwardingListener2 from underlying player.
forwardingPlayer.removeListener(listener);

// No way to remove forwardingListener1 from the underlying player...
```

libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
```diff
@@ -54,6 +56,7 @@
 public class ForwardingPlayer implements Player {
 
   private final Player player;
+  protected final Map<Listener, Listener> listeners = new HashMap<>();
```
icbaker: Can this be private?

Consider also using [`IdentityHashMap`](https://developer.android.com/reference/java/util/IdentityHashMap) to reinforce the idea of identity equality.

libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
```diff
@@ -90,7 +95,11 @@ public void addListener(Listener listener) {
    */
   @Override
   public void removeListener(Listener listener) {
-    player.removeListener(new ForwardingListener(this, listener));
+    Listener forwardingListener = listeners.remove(listener);
+    if (forwardingListener == null) {
+      throw new IllegalArgumentException("Trying to remove listener that never was registered");
```
icbaker: I wonder if we can avoid **always** throwing an exception in this case - atm `ExoPlayer` doesn't throw if you remove an unrecognized listener (since it's assumed that the goal state of "this listener isn't registered" is met by doing nothing, so the method can do nothing).

I don't think the `Player` interface mandates what happens in this case. We could delegate the decision of "what happens when an unrecognized listener is removed?" to the underlying impl by **deliberately** passing in a known-unrecognized impl in here:

```java
if (forwardingListener == null) {
  // Try and remove a listener instance that is guaranteed not to be
  // registered, allowing the underlying impl to either throw or ignore.
  player.removeListener(new Listener() {});
}
```

libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
icbaker: I wonder if this addition to the bookkeeping set should go after the call to the underlying player (in case it fails for some reason)?

nift4:
Thanks for the thoughful review. :)

> You can probably also remove the ForwardingListener.equals/hashCode implementations in this same PR?

Done.

> Please can you also add some tests to ForwardingPlayerTest for this change.

Done.

> Player.addListener() can be called on any thread, so you need some concurrency protection here.

Done.

> You probably want to check whether the map already contains key = listener or not - otherwise the following operations go wrong:

Done, and this is tested in the test now.

> Can this be private?

Yeah, it could, but I think it's better to make it protected. While it's unrelated to the bug that prompted this PR, having access to all registered listeners to be able to send them events is quite useful in the context of a ForwardingPlayer (as it generally has to dispatch events for everything if it changes anything at all).

> Consider also using [IdentityHashMap](https://developer.android.com/reference/java/util/IdentityHashMap) to reinforce the idea of identity equality.

Done. It's also tested in the test that this works even if player does it differently.

> I wonder if we can avoid always throwing an exception in this case - atm ExoPlayer doesn't throw if you remove an unrecognized listener (since it's assumed that the goal state of "this listener isn't registered" is met by doing nothing, so the method can do nothing).
>
> I don't think the Player interface mandates what happens in this case. We could delegate the decision of "what happens when an unrecognized listener is removed?" to the underlying impl by deliberately passing in a known-unrecognized impl in here:

Although not in the way you proposed, this is done as well, and tested as well.

> I wonder if this addition to the bookkeeping set should go after the call to the underlying player (in case it fails for some reason)?

Done.

libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
```diff
@@ -78,7 +80,14 @@ public Looper getApplicationLooper() {
    */
   @Override
   public void addListener(Listener listener) {
-    player.addListener(new ForwardingListener(this, listener));
+    synchronized (listeners) {
+      Listener forwardingListener = listeners.get(listener);
+      if (forwardingListener == null) {
+        forwardingListener = new ForwardingListener(this, listener);
+      }
+      player.addListener(forwardingListener);
```
icbaker, line 91: Shouldn't this be inside the `if (forwardingListener == null)` block? For symmetry with your `removeListener` impl, which will only ever call `removeListener()` once if called repeatedly with the same `listener` (so we should also only call `addListener` if we're called repeatedly with the same `listener`).

At which point, it might make sense to flip the contents check into a guard clause:

```java
if (listeners.containsKey(listener)) {
  return;
}
ForwardingListener forwardingListener = new ForwardingListener(this, listener);
player.addListener(forwardingListener);
listeners.put(listener, forwardingListener);
```

libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
```diff
@@ -41,11 +45,23 @@
...
+    assertThat(listener2).isEqualTo(listener1);
+    assertThat(listener1.hashCode()).isEqualTo(2);
+    assertThat(listener2.hashCode()).isEqualTo(2);
 
+    player.addListener(listener1);
+    // Add listener1 again.
+    player.addListener(listener1);
+    player.addListener(listener2);
+    assertThat(player.listeners).hasSize(1);
+    player.listeners.clear();
```
icbaker: Up to here it looks like you are basically testing your test/fake implementations, and not the actual prod code.

I think the implementation of `AllIsEqualPlayerListener` and `EqualityBasedFakePlayer` is trivial enough that you don't need to check it every time (so you can remove those assertions here and below).

libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
```diff
@@ -139,4 +241,46 @@ public void removeListener(Listener listener) {
...
+
+  private static class IdentityBasedFakePlayer extends StubPlayer {
+
+    private final List<Listener> listeners = new ArrayList<>();
+
+    @Override
+    public void addListener(Listener listener) {
+      for (Listener listener1 : listeners) {
+        if (listener == listener1) {
+          throw new IllegalArgumentException("Trying to add duplicate listener added");
```
icbaker: Why does this `Player` impl throw an exception on a duplicate listener vs silently ignoring it like `EqualityBasedFakePlayer` does?

Based on the type names I assumed the only difference between these impls was the way they kept track of listeners (reference equality vs equals()).

I don't think the `Player` interface mandates one behaviour over the other, so I can see an argument that it's reasonable to test delegate impls that do both, but I wonder how much extra coverage it actually offers? (i.e. can you think of a reasonable bug that might be hidden in your impl if you removed this exception being thrown from here?)

libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
```diff
@@ -73,9 +106,78 @@ public void removeListener_removesForwardingListener() {
...
+    assertThat(player.listeners).hasSize(2);
+    player.listeners.clear();
+
+    // The listeners are equal, but the Player handles that, and ForwardingPlayer should, too.
+    ForwardingPlayer forwardingPlayer = new ForwardingPlayer(player);
+    forwardingPlayer.addListener(listener1);
+    // Add listener1 again.
+    assertThrows(IllegalArgumentException.class, () -> forwardingPlayer.addListener(listener1));
+    assertThat(player.listeners).hasSize(1);
+    assertThrows(IllegalArgumentException.class, () -> forwardingPlayer.removeListener(listener2));
```
icbaker: I think this case of removing an un-registered listener is covered by the assertion below on L159, so it can likely be removed from here.

icbaker:
> > Can this be private?
>
> Yeah, it could, but I think it's better to make it protected. While it's unrelated to the bug that prompted this PR, having access to all registered listeners to be able to send them events is quite useful in the context of a ForwardingPlayer (as it generally has to dispatch events for everything if it changes anything at all).

I understand this desire, but I think we should keep it `private` in this PR and consider whether we want to open up the visibility as a follow-up (since as you say, it's pretty unrelated to this fix). Arguments against opening it up:
1. An implementation of `ForwardingPlayer` that is getting complex enough to need to fire its own listener events should probably be using `ForwardingSimpleBasePlayer` instead.
2. Subclasses would have to be careful to add their own `synchronized` block on every access too.
3. Simply making it protected allows subclasses to also add/remove listeners, which gets increasingly hard to reason about
   * This and (2) could be mitigated by explicit 'fire event X' methods instead of exposing the whole collection of listeners, but I think (1) is probably the most compelling argument that this is trying to make `ForwardingPlayer` "too clever".

libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
nift4: > can you think of a reasonable bug that might be hidden in your impl if you removed this exception being thrown from here?

ForwardingPlayer returning early and hence never triggering the exception being passed to the caller.

libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
nift4: > For symmetry with your removeListener impl, which will only ever call removeListener() once if called repeatedly with the same listener

I don't think that's the case. I changed removeListener() to call down into the player every time as well. Sure, not with the ForwardingListener as keeping it would be a memory leak, but with the passes in listener. Add and remove are already symmetric, no matter if listener is registered or not we call down into the actual player and it decides whether to throw or not.

libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
nift4: Point taken, will remove.

libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
nift4: Done

libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
nift4: > Based on the type names I assumed the only difference between these impls was the way they kept track of listeners (reference equality vs equals()).

The basic idea was that the `EqualityBasedFakePlayer` behaves like `ExoPlayer`, using equality and never throwing, while `IdentityBasedFakePlayer` is the polar opposite ignoring equality and throwing when nonsensical things are done, both of which seem to be valid interpretations of the Player interface contract.

edit: While ForwardingPlayer by necessity has to choose whether to allow equality or not, the choice of throwing or ignoring can (and like you said, should) be left to the underlying Player.

nift4:
> I understand this desire, but I think we should keep it private in this PR

Point taken, I admit I'm only using ForwardingPlayer for this because ForwardingSimpleBasePlayer didn't work out as ExoPlayer keeps triggering asserts (see #2674). The long term solution is to fix ExoPlayer to not trigger asserts and use ForwardingSimpleBasePlayer.

libraries/common/src/main/java/androidx/media3/common/ForwardingPlayer.java
icbaker: ack - makes sense, thanks

libraries/common/src/test/java/androidx/media3/common/ForwardingPlayerTest.java
icbaker: makes sense, thanks for explaining.

icbaker (approved):

icbaker:

nift4:
Thanks! :D

PR #2685, opened 2025-07-30
Remove outdated qualifiers on resources folders

MGaetan89:
Some resources were still in the `-v21`/`-v22`/`-v23` resources folders.
This PR moves them to the main folder, updating the original files when needed.


nift4:
shouldn't this also remove .png files for example in ui library? while the .png files [would still be used](https://android.googlesource.com/platform/frameworks/base/+/31245b4f06003f1c8cd44c31b387c96ab4e282f9%5E%21/) on devices that exactly match the standard dpi qualifiers, anydpi .xml files will be used for any other dpi, and frankly there is no reason to keep the app size bloat that are these .png files anymore as anydpi vector drawables work just fine for these.

MGaetan89:
Good point! I'll look into removing the PNG files when there is a vector equivalent

tonihei:
> shouldn't this also remove .png files for example in ui library

Yes, it should, thanks for noting! These pngs only existed to support API<21 devices, so can now be removed entirely.

MGaetan89:
Done. I've removed PNG files that already have a vector equivalent in https://github.com/androidx/media/pull/2685/commits/a0d7ba3cae95748c4bba123949a7691d679626e6, and converted `exo_icon_circular_play` to vector in https://github.com/androidx/media/pull/2685/commits/c00d28f05421026ca45f6bec53fb45bd3a899bba.

> [!NOTE]
> There's still `exo_icon_vr.png` in `:lib-ui` that is not a vector, but I didn't find the icon on https://fonts.google.com/icons

rohitjoins:

MGaetan89:
@rohitjoins with e42940c71885aa98b95f57c1434241dccea7056c, the content of https://github.com/androidx/media/blob/main/libraries/session/src/main/res/values-v21/styles.xml is no longer applied. Is it ok?
If it is, maybe the [`values-v24/styles.xml`](https://github.com/androidx/media/blob/main/libraries/session/src/main/res/values-v24/styles.xml) file can be removed too? Its content is similar to the one in [`values/styles.xml`](https://github.com/androidx/media/blob/main/libraries/session/src/main/res/values/styles.xml) now.

rohitjoins:
@MGaetan89 yes, that change is intentional. This is because the values are now provided by the base `androidx.media`.

> If it is, maybe the [values-v24/styles.xml](https://github.com/androidx/media/blob/main/libraries/session/src/main/res/values-v24/styles.xml) file can be removed too?

That's a great observation. Will remove that file too! Thanks!

oceanjules:
> There's still `exo_icon_vr.png` in `:lib-ui` that is not a vector, but I didn't find the icon on https://fonts.google.com/icons

How about 
* [head_mounted_device ](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:head_mounted_device:FILL@0;wght@0;GRAD@0;opsz@24&icon.size=24&icon.color=%235f6368) most matches the glasses we have
* [vrpano](https://fonts.google.com/icons?selected=Material+Icons+Outlined:vrpano:&icon.size=24&icon.color=%235f6368&icon.query=realit&icon.set=Material+Icons) doesn't really look like VR

MGaetan89:
> * [head_mounted_device ](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:head_mounted_device:FILL@0;wght@0;GRAD@0;opsz@24&icon.size=24&icon.color=%235f6368) most matches the glasses we have

I personally would have gone with that one, but I wasn't sure if the visual change was ok (especially since it is used in the player itself, and not just the demo).

rohitjoins:
Had to rollback this change due to internal Gradle build failure, will merge it again after we can find the cause.

MGaetan89:
Thanks for the info. Let me know if I can help in fixing the issue.

rohitjoins:
Merged in https://github.com/androidx/media/commit/306c2d0e92444cdc0f5bb168afeb1ba9a11f0955. 

Will clean up remaining PNGs in follow up changes.

PR #2694, opened 2025-08-03
Fix text color in `:demo:session`

MGaetan89:
The text color of the `Play` button in `:demo:session` was using the default text color, while the `Shuffle` button has a white text color.
This aligns the two buttons to use white in both cases.

-----

Here's a before/after of this change:

| Description | `main` | This PR |
|--------|--------|--------|
| Dark mode | Screenshot_20250804_105838 | Screenshot_20250804_105933 |
| Light mode | Screenshot_20250804_105849 | Screenshot_20250804_105939 |


rohitjoins:

PR #2700, opened 2025-08-06
Add `MetadataState` to `:ui:compose`

MGaetan89:
This commit introduces `MetadataState`, a Compose state that exposes metadata information about the current `MediaItem`. At the moment, it only provides the media uri.

### DIFF
--- a/libraries/ui_compose/src/main/java/androidx/media3/ui/compose/state/MetadataState.kt
+++ b/libraries/ui_compose/src/main/java/androidx/media3/ui/compose/state/MetadataState.kt
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.media3.ui.compose.state
+
+import android.net.Uri
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.media3.common.Player
+import androidx.media3.common.listen
+import androidx.media3.common.util.UnstableApi
+
+/**
+ * Remembers the value of a [MetadataState] created based on the passed [Player] and launches a
+ * coroutine to listen to the [Player's][Player] changes. If the [Player] instance changes between
+ * compositions, this produces and remembers a new [MetadataState].
+ */
+@UnstableApi
+@Composable
+fun rememberMetadataState(player: Player): MetadataState {
+  val metadataState = remember(player) { MetadataState(player) }
+  LaunchedEffect(player) { metadataState.observe() }
+  return metadataState
+}
+
+/**
+ * State that holds information to correctly deal with UI components related to the current
+ * [MediaItem][androidx.media3.common.MediaItem] metadata.
+ *
+ * @property[uri] The URI of the current media item, if available.
+ */
+@UnstableApi
+class MetadataState(private val player: Player) {
+  var uri by mutableStateOf(player.getMediaItemUri())
+    private set
+
+  suspend fun observe(): Nothing {
+    player.listen { events ->
+      if (events.contains(Player.EVENT_MEDIA_ITEM_TRANSITION)) {
+        uri = getMediaItemUri()
+      }
+    }
+  }
+
+  private fun Player.getMediaItemUri(): Uri? {
+    return currentMediaItem?.localConfiguration?.uri
+  }
+}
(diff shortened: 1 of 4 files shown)


libraries/ui_compose/src/main/java/androidx/media3/ui/compose/state/MetadataState.kt
```diff
@@ -0,0 +1,65 @@
...
+  suspend fun observe(): Nothing {
+    player.listen { events ->
+      if (events.contains(Player.EVENT_MEDIA_ITEM_TRANSITION)) {
+        uri = getMediaItemUri()
+      }
+    }
+  }
+
+  private fun Player.getMediaItemUri(): Uri? {
+    return currentMediaItem?.localConfiguration?.uri
```
oceanjules: Should be protected by `player.isCommandAvailable(COMMAND_GET_CURRENT_MEDIA_ITEM)`

Something like
```
player.listen { events ->
      if (
        events.containsAny(
          Player.EVENT_MEDIA_ITEM_TRANSITION,
          Player.EVENT_AVAILABLE_COMMANDS_CHANGED,
        )
      ) {
          uri = getMediaItemUriWithCommandCheck()
        }
      }
    }

  private fun Player.getMediaItemUriWithCommandCheck(): Uri? {
        if (player.isCommandAvailable(Player.COMMAND_GET_CURRENT_MEDIA_ITEM)) {
          return currentMediaItem?.localConfiguration?.uri
        } else { null }
   }
  ```

File: libraries/ui_compose/src/test/java/androidx/media3/ui/compose/state/MetadataStateTest.kt [hunk 2]
```diff
@@ -0,0 +1,108 @@
...
+    player.seekToNext()
+    composeTestRule.waitForIdle()
+
+    assertThat(state.uri).isNull()
+
+    player.seekToNext()
+    composeTestRule.waitForIdle()
+
+    assertThat(state.uri).isEqualTo(uri2)
+  }
```
oceanjules, line 101: Great tests! Could you add one when the command is not available, then becomes available (TestPlayer has addCommand/removeCommand methods)

libraries/ui_compose/src/test/java/androidx/media3/ui/compose/utils/TestPlayer.kt
```diff
@@ -172,9 +172,7 @@ internal class TestPlayer(
 
   fun setDuration(uid: String, durationMs: Long) {
     val index = state.playlist.indexOfFirst { it.uid == uid }
-    if (index == -1) {
-      throw IllegalArgumentException("Playlist does not contain item with uid: $uid")
-    }
+    require(index > -1) { "Playlist does not contain item with uid: $uid" }
```
oceanjules: nit, but maybe index >= 0 will look slightly better

libraries/ui_compose/src/main/java/androidx/media3/ui/compose/state/MetadataState.kt
MGaetan89: Good catch!
I've added the check for `COMMAND_GET_CURRENT_MEDIA_ITEM` and a dedicated test in https://github.com/androidx/media/pull/2700/commits/944314a8bbc7bc62f09575a91483225e56ecd500.

oceanjules:

MGaetan89:
@oceanjules I see that you have pushed the new `listenTo` extension. Is it fine if I update this PR to use it?

oceanjules:
@MGaetan89 yep, go for it! Although the CL in its current shape and form is still under internal scrutiny at the moment, so a `listenTo` fix will not necessarily help push it forward, sorry. Multiple problems here, but largely broken down into: 1) shouldn't flatten MediaItem into primary fields (like `localConfiguration.uri` to simply `uri`) and 2) should group logical `Player` data into larger state holders (maybe one that will hold information beyond MediaItem and MediaMetadata). 

I understand that this CL was not even on the original plan (GL surface type brainstorm is also happening in internally) and now it's blowing up into a whole design discussion, but please bear with us while we design maintainable and reusable Compose blocks.

MGaetan89:
> yep, go for it!

Done 👍🏻 

> Although the CL in its current shape and form is still under internal scrutiny at the moment, so a `listenTo` fix will not necessarily help push it forward, sorry.

No worries, I am not in a hurry to have this merged 🙂 

> Multiple problems here, but largely broken down into: 1) shouldn't flatten MediaItem into primary fields (like `localConfiguration.uri` to simply `uri`) and 2) should group logical `Player` data into larger state holders (maybe one that will hold information beyond MediaItem and MediaMetadata).
> 
> I understand that this CL was not even on the original plan (GL surface type brainstorm is also happening in internally) and now it's blowing up into a whole design discussion, but please bear with us while we design maintainable and reusable Compose blocks.

Thanks for the insights! I wasn't expecting this change to go through easily. This new state probably needs to have a broader purpose, rather than just exposing a URI as a shortcut.
Anyway, let me know if I need to make any change here once the design decision has moved forward, or if I can be of assistance in any way.

bubenheimer:
>  2) should group logical `Player` data into larger state holders (maybe one that will hold information beyond MediaItem and MediaMetadata).

Ultimately it should be a single (snapshot) Player state to stop fighting the (Compose) framework for eternity.

oceanjules:
I will close this PR due to the recent addition of https://github.com/androidx/media/commit/a2f071d1217614aa1d4750fa70fb862982a27bd6. In a nutshell, we decided that "only metadata" state holder was probably too narrow of a scope, whereas "a whole timeline" state holder was too broad (although we will soon publish something like a `PlaylistState`). Also, we decided not to flatten the mediaitem's nested structure (i.e. bringing the uri out of localConfiguration onto the top level of some UI state holder class).

Hopefully, you can get all the necessery information out of `CurrentMediaItemState.mediaItem`. I know this whole change was only motivated by https://github.com/androidx/media/pull/2619, so if you rebase that on the latest main, we could have a discussion about using reflection and exoplayer in our new modules.

PR #2709, opened 2025-08-10
Align demo apps title

MGaetan89:
This PR aligns the demo apps title, by using the `Media3` prefix. With this change, all demo apps are grouped together in the app launcher, making them easier to find.

### DIFF
--- a/demos/cast/src/main/res/values/strings.xml
+++ b/demos/cast/src/main/res/values/strings.xml
@@ -16,7 +16,7 @@
 
 <resources>
 
-  <string name="application_name">Exo Cast Demo</string>
+  <string name="application_name">Media3 Cast Demo</string>
 
   <string name="media_route_menu_title">Cast</string>
 
--- a/demos/composition/src/main/res/values/strings.xml
+++ b/demos/composition/src/main/res/values/strings.xml
@@ -13,7 +13,7 @@
      limitations under the License.
 -->
 <resources>
-  <string name="app_name">Composition Demo</string>
+  <string name="app_name">Media3 Composition Demo</string>
   <string name="edit">Edit</string>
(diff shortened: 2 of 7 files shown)


oceanjules:
* Pros: all the apps will be grouped together when sorted alphabetically 
* Cons: hard to distinguish them because "Media3" takes up too much space. Compose and Composition Demo will probably look the same.

image

image

----------

What is nice is that Cast demo has those casting waves in the png, maybe we could distinguish by that

* https://github.com/androidx/media/blob/main/demos/cast/src/main/res/mipmap-hdpi/ic_launcher.png
* https://github.com/androidx/media/blob/main/demos/main/src/main/res/mipmap-hdpi/ic_launcher.png

MGaetan89:
> * hard to distinguish them because "Media3" takes up too much space. Compose and Composition Demo will probably look the same.

Oh, that's a good point. Maybe remove every prefix? Every demo has a Media3 logo, so it's easy to know where they come from, even if they are just named `Session Demo` for example.

Or should we just drop this?

oceanjules:
> Maybe remove every prefix? 

But then they are not nicely sorted. I was thinking of M3 as a prefix, but @marcbaechinger said it was car-related :) 

If you have the energy to create different mipmap quality icons, then maybe icon+M3 prefix will be enough for distinguishing+sorting
Screenshot 2025-08-13 at 15 06 12

MGaetan89:
> @marcbaechinger said it was car-related :)

Or Material 3 :)

> If you have the energy to create different mipmap quality icons, then maybe icon+M3 prefix will be enough for distinguishing+sorting

I'll be on holiday next week, and I don't know if I'll have time to look into that before. But I can give it a try once I'm back.

Issue #2717, opened 2025-08-15
Remove redundant appcompat line in shortform demo build.gradle


gramound:
https://github.com/androidx/media/blob/b7bbc6e2bc3e45ff3ed99884c114c50f03bba5c9/demos/shortform/build.gradle#L76

These lines are redundant: 
```
    implementation 'androidx.appcompat:appcompat:1.6.1'
    implementation 'androidx.appcompat:appcompat:' + androidxAppCompatVersion
```
`androidxAppCompatVersion` is already set to 1.6.1 in constants.gradle

marcbaechinger:
Thanks for your PR!

I'm closing this. We can follow up on the PR if required.

PR #2721, opened 2025-08-15
Removed redundant dependency in shortform demo

gramound:
Fixes: Issue: #2717


marcbaechinger:

PR #2730, opened 2025-08-17
Fix missing link in ExoPlayer API reference.

amanc3765:
- This minor change fixes the missing link to `Builder` object in `ExoPlayer` API reference.
- Although the link exists, but due to a bug (as per TW team), the link is not rendered since it is split into two lines.

**PS:** This change also allows me to get familiar with the Developer workflow of this Github repo. Thanks for your time!

Screenshot 2025-08-17 at 3 31 51 PM


icbaker:
As discussed offline, unfortunately this won't work to resolve the issue because the line-break will be re-introduced by google-java-format which runs on all our files at presubmit.

The right way to solve this is by fixing the doc generation pipeline, rather than trying to manually introduce line-breaks into the source.

amanc3765:
Thanks Ian for the heads up!

PR #2738, opened 2025-08-20
Parse ALAC bit depth from initialization data

nift4:


rohitjoins:

nift4:
Thanks for the speedy review :D

Issue #2742, opened 2025-08-25
Composition Player gets stuck during sequence transition when using MultipleInputVideoGraph


AradiPatrik:
### Version

Media3 main branch

### More version details

Top of main branch as of Aug 25 2025

### Devices that reproduce the issue

All devices

### Devices that do not reproduce the issue

none

### Reproducible in the demo app?

Yes

### Reproduction steps

# DefaultVideoCompositor deadlock causes multi-sequence compositions to freeze during media item transitions

## Summary
When using `MultipleInputVideoGraph` with sequences containing multiple media items, the `DefaultVideoCompositor` deadlocks when one sequence transitions between items, causing the entire composition to freeze.

1. Open demo-composition app
2. Select "2x2 grid" layout
3. Delete all items from video sequence
4. Add the first sample video 8 times (10-second video)
5. Start playback

https://github.com/user-attachments/assets/fd1c7ac6-8f9a-4d0c-9363-31a0e1e4b781

**Expected:** All 4 sequences play their 2 videos (20 seconds total) smoothly  
**Actual:** Video freezes after first 10 seconds, audio continues, timer keeps counting indefinitely, seeking breaks

## Root Cause
`DefaultVideoCompositor.getFramesToComposite()` blocks ALL compositing when ANY input has empty frame queue:

```java
for (int i = 0; i < inputSources.size(); i++) {
  if (inputSources.valueAt(i).frameInfos.isEmpty()) {
    return ImmutableList.of();  // Blocks all compositing
  }
}
```

During media item transitions, the transitioning sequence's frame queue becomes temporarily empty while the decoder reinitializes. This prevents the compositor from producing any output frames, causing other sequences to never reach their end-of-stream signal.

**Example timing from debugging:**
- Sequence transitions at 10s (decoder release → empty queue → reinitialization)
- Compositor stops producing output during this ~200ms gap
- Other sequences need their final ~120ms of frames to reach end-of-stream
- These final frames are never composited due to the blocking condition
- Sequences cannot signal end-of-stream (lastOutputFrame < finalFrame)
- Permanent deadlock

### Early End-of-Stream Signal
Modify `PlaybackVideoGraphWrapper` to signal end-of-stream slightly before the actual end to avoid the deadlock:
```java
// In onOutputFrameAvailableForRendering()
boolean isLastFrame = finalFramePresentationTimeUs != C.TIME_UNSET
[...]

PR #2744, opened 2025-08-25
Fix DefaultVideoCompositor deadlock during media item transitions

AradiPatrik:
When using MultipleInputVideoGraph with sequences containing multiple
media items, the compositor would deadlock when transitioning between
items within a sequence. This prevented other sequences from reaching
their end-of-stream signal, causing the entire composition to freeze.

The root cause was getFramesToComposite() returning an empty list
whenever ANY input source had an empty frame queue. During media item
transitions, the transitioning sequence's queue becomes temporarily
empty while the decoder reinitializes (~200ms). This blocked all
composition, preventing other sequences from compositing their final
frames needed to signal end-of-stream.

This change introduces frame caching to maintain composition continuity:
- Added lastAvailableFrames cache to store the most recent frame from
  each input source
- Modified getFramesToComposite() to use cached frames when an input's
  queue is temporarily empty
- Updated selectSecondaryFrame() to use single available frames instead
  of waiting for buffering
- Ensured primary frame is only released after being cached

With these changes, composition continues using the last available frame
from transitioning sequences, allowing other sequences to properly reach
end-of-stream while maintaining visual continuity (no black frames).

Fixes: https://github.com/androidx/media/issues/2742

### DIFF
--- a/libraries/effect/src/main/java/androidx/media3/effect/DefaultVideoCompositor.java
+++ b/libraries/effect/src/main/java/androidx/media3/effect/DefaultVideoCompositor.java
@@ -88,6 +88,9 @@ public final class DefaultVideoCompositor implements VideoCompositor {
   private final SparseArray<InputSource> inputSources;
 
   @GuardedBy("this")
+  private final SparseArray<InputFrameInfo> lastAvailableFrames = new SparseArray<>();
+
+  @GuardedBy("this")
   private boolean allInputsEnded; // Whether all inputSources have signaled end of input.
 
   private final TexturePool outputTexturePool;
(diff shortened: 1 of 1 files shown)


AradiPatrik:
Here's the manual test result. The sequences transition without deadlock in demo-composition.

https://github.com/user-attachments/assets/b15c2ff4-893d-45bb-ae52-07aebdb93f32

tonihei:
@ychaparov Could you take a look or reassign as needed?

claincly:
Hello @AradiPatrik, thanks for sending the PR. I have a few naive questions:

- This change does seem to change some of the exising composition logic, have you verified that the existing tests, like `DefaultVideoCompositorPixelTest` pass?
- Can you add some tests to make sure that your change is tested?

Issue #2418, opened 2025-05-14
Sessions demo in compose


pawaom:
Sessions demo in compose

We already have the exoplayer demo in compose, 
Recently in one of the [commits ](https://github.com/androidx/media/commit/d1642ec29c018d32c74701f4ce90ae9be3a0d707)I saw even composition has been updated to compose,

It would be better to have the sessions demo in compose as well.

PR #2787, opened 2025-09-18
Add LazyColumn shortform demo with single-player preloading (addresses #1853)

### LINKED ISSUE #1853: Blank/Slow Frame Render for videos
Unfortunately we can't answer all questions. Unclear questions or questions with
insufficient information may not get attention.

Before filing a question:
-------------------------

- Ask general Android development questions on Stack Overflow
- Search existing issues, including issues that are closed
  https://github.com/androidx/media/issues?q=is%3Aissue
- For ExoPlayer-related questions, please also check for existing questions on
  the ExoPlayer tracker:
  https://github.com/google/ExoPlayer/issues?q=is%3Aissue

When filing a question:
-------------------------

Describe your question in detail.
I've been trying to implement functionality similar to the Shortform app, but encountered an issue during video transitions. While my implementation is based on a single player with a "Next" button, I noticed two main issues:

1. The first video takes extra time to render its initial frame. Additionally, when moving to the next video, intermittent delays appear on some videos (typically the 0th, 3rd, or 4th in the sequence). These delays occur regardless of how I order URLs from MediaItemDatabase. If I go through videos from 0 -> 5 and then back from 5 -> 0, the initial frame delay reappears, especially for the 0th video.
2. Although I use a single player in my implementation, switching between videos is occasionally slow, causing a visible lag for the initial frame of some videos. In the Shortform demo app, the transition is smoother, though I sometimes notice similar initial frame lags on the first and a few middle videos.

I'm attaching recordings of both the Shortform app issue and my implementation for reference.

Could you help me identify possible causes for the render delays in my implementation, and why it performs slower compared to the demo app? Additionally, I’d appreciate insights into the primary issues faced in the Shortform demo app as well.

| Issue in Shortform App | Issue Faced in my internal demo app(Similar to shortform ) |
|--------|--------|
|   |  |

In case your question refers to a problem you are seeing in your app:

- Output of running `$ adb bugreport` in the console

In case your question is related to a piece of media:

- URI to test content

- For protected content:
  - DRM scheme and license server URL
  - Authentication HTTP headers

Don't forget to check ExoPlayer's supported formats and devices, if applicable
(https://developer.android.com/guide/topics/media/exoplayer/supported-formats).

If there's something you don't want to post publicly, please submit the issue,
then email the link/bug report to android-media-github@google.com using a
subject in the format "Issue #1234", where #1234 is your issue number (we don't
reply to emails).

Amud:
https://github.com/user-attachments/assets/c9d82c21-5b97-4ebf-a6f2-e3dd2777506f

Adds a LazyColumn-based shortform demo that optimizes a single-ExoPlayer setup with preloading, addressing performance concerns raised in #1853.

### Related issue
Fixes #1853.

### Background
- Initial frame delays with single player + `PreloadManager`
- Lag on first video and during transitions
- Performance gap vs a multi-player pool approach

### Solution
- **Window-based media management**
  - Sliding window (~10 items) with dynamic load/unload based on scroll
  - Reduces memory pressure while maintaining smooth playback
- **Optimized preloading**
  - `TargetPreloadStatusControl` preloads 3–5s for adjacent items
  - Preloads items at distance ±1 and ±2
- **Reactive cache tracking**
  - `mutableStateMapOf` tracks cache completion
  - Thumbnails load only after content is cached
- **Single player benefits**
  - Lower memory use than a 3‑player pool
  - Simpler state management with smooth transitions

### Performance (observed)
- Reduced initial frame delay via targeted preloading
- Windowing avoids memory exhaustion
- Cache-aware thumbnails improve perceived responsiveness
- Consistent transition smoothness across indices

### Implementation highlights
- `LazyColumnPlayerManager`: sliding window + preload management
- `LazyColumnTargetPreloadStatusControl`: 3–5s buffer strategy
- `SinglePlayerSetupHelper`: optimized load control and cache
- Compose UI reacts to cache status changes

### Testing
- Transitions across all items
- Smooth playback without noticeable frame drops
- Stable memory during extended scrolling
- No rendering delays after initial cache warm-up

### Entry point
Adds a “LazyColumn” button in `MainActivity` alongside the existing ViewPager demo.

### DIFF
--- a/demos/shortform/build.gradle
+++ b/demos/shortform/build.gradle
@@ -14,6 +14,7 @@
 apply from: '../../constants.gradle'
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
+apply plugin: 'org.jetbrains.kotlin.plugin.compose'
 
 android {
     namespace 'androidx.media3.demo.shortform'
@@ -57,6 +58,7 @@ android {
     }
     buildFeatures {
         viewBinding true
+        compose true
     }
     sourceSets {
         main {
@@ -80,4 +82,14 @@ dependencies {
     implementation project(modulePrefix + 'lib-exoplayer-dash')
     implementation project(modulePrefix + 'lib-exoplayer-hls')
     implementation project(modulePrefix + 'lib-ui')
+
+    // Compose dependencies
+    def composeBom = platform('androidx.compose:compose-bom:2024.12.01')
+    implementation composeBom
+    implementation 'androidx.activity:activity-compose:1.9.0'
+    implementation 'androidx.compose.foundation:foundation'
+    implementation 'androidx.compose.material3:material3'
+    implementation 'androidx.compose.ui:ui-tooling-preview'
+    implementation 'androidx.compose.runtime:runtime'
+    debugImplementation 'androidx.compose.ui:ui-tooling'
 }
--- a/demos/shortform/src/main/AndroidManifest.xml
+++ b/demos/shortform/src/main/AndroidManifest.xml
@@ -34,6 +34,10 @@
       android:exported="false"
       android:label="@string/title_activity_view_pager"
       android:name=".viewpager.ViewPagerActivity"/>
+    <activity
+      android:exported="false"
+      android:label="@string/title_activity_lazy_column"
+      android:name=".lazycolumn.LazyColumnActivity"/>
   </application>
 
   <uses-permission android:name="android.permission.INTERNET" />
--- a/demos/shortform/src/main/java/androidx/media3/demo/shortform/MainActivity.kt
+++ b/demos/shortform/src/main/java/androidx/media3/demo/shortform/MainActivity.kt
@@ -23,6 +23,7 @@ import android.view.View
 import android.widget.EditText
 import androidx.appcompat.app.AppCompatActivity
 import androidx.media3.common.util.UnstableApi
+import androidx.media3.demo.shortform.lazycolumn.LazyColumnActivity
 import androidx.media3.demo.shortform.viewpager.ViewPagerActivity
 import java.lang.Integer.max
 import java.lang.Integer.min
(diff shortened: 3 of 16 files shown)


Amud:
hi @oceanjules , Please let me know if there are any feedbacks on this PR.

oceanjules:
Hi Amud, sorry for a delay, thank you for waiting! This is a great contribution and we appreciate the effort you've put into addressing the performance issues with the short-form demo.

However, this pull request has grown quite large and mixes several distinct concepts. You PR adds production-like features which is a great example of how to build a short-form video feeds, but not a good example for a developer who wants to understand a single feature of the Media3 library. In a real-world app, this is exactly what would be expected from a developer to make a smooth experience for a user. However, our demo apps are not made to be standalone apps, but instead showcase particular features of the library, grouped by domain.

There are users who might want a purely java+xml layouts example of using PreloadManager without having to untangle it from Compose UI. There are users who would want the Session Demo to be in Compose (https://github.com/androidx/media/issues/2418). Someone interested in using Media3 with Compose would be overwhelmed by the custom caching and preloading implementation.

At the same time, we are not ready to productionise parts of this PR either because we do not have strong guidance about player management logic and how to keep a pool of players, what algorithm to use to evict old players and how to initialize more. We often use our own demo apps to ensure the feature will land well which is why we want them to be as lightweight as possible. When people report bugs, they should be able to quickly build the app and give us short reproduction steps.

We can leave this PR for any developers who stumble upon it and want to study your code to get some inspiration for building a performant, short-form video app - although for better visibility I would advise you to make a Medium article :)

oceanjules:
I will close this PR because we now have a short-form screen in the [compose demo](https://github.com/androidx/media/tree/main/demos/compose/src/main/java/androidx/media3/demo/compose). See: https://github.com/androidx/media/commit/390315218351e16bddb3aee7f3a1ce2e367406cb, https://github.com/androidx/media/commit/bc5c18cd09ebd11421765909b893c33679da9963,

PR #2798, opened 2025-09-23
Allow gains to be larger than 1.0 in gainProcessor

prabhavsharma:


tonihei:
Could you check that the CLA is signed (see comment above) so that we can take a look? Thanks!

PR #2816, opened 2025-10-01
 Open Fix missing KEYCODE_HEADSETHOOK caseMain

### LINKED ISSUE #2768: ForegroundServiceDidNotStartInTimeException due to dropped KeyCode. KEYCODE_HEADSETHOOK
### Version

Media3 main branch

### More version details

_No response_

Samsung devices are the vast majority of the crashes.

_No response_

No

### Reproduction steps

I can reproduce the issue via unit tests

### Expected result

The app doesn't crash.

### Actual result

          Fatal Exception: android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{f5d3566 u0 xxxx/xxxxService}
       at android.app.ActivityThread.generateForegroundServiceDidNotStartInTimeException(ActivityThread.java:2118)
       at android.app.ActivityThread.throwRemoteServiceException(ActivityThread.java:2089)
       at android.app.ActivityThread.-$$Nest$mthrowRemoteServiceException()
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2354)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:211)
       at android.os.Looper.loop(Looper.java:300)
       at android.app.ActivityThread.main(ActivityThread.java:8227)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1028)
        

### Media

I've had an ongoing issue with ForegroundServiceDidNotStartInTimeExceptions.   I implement onPlaybackResumption, and I have a MediaButtonReceiver that implements shouldStartForegroundService to verify I have something to play.  My service implements MediaLibraryService and I have a custom player.  Over the past few months, I've been adding Google Analytics events at various points.   When the crashes occur, onPlaybackResumption isn't called.  I just overrode onMediaButtonEvent, it just logs an event and returns super.   onMediaButtonEvent is called.  It's called from onMediaButtonEvent in MediaSessionImpl.  I'm also logging the intent.   The intent is always something like this:
 `Intent { act=android.intent.action.MEDIA_BUTTON flg=0x10000010 cmp=xxxx/xxxxrService (has extras) } Bundle[android.intent.extra.PACKAGE_NAME=android, android.intent.extra.KEY_EVENT=KeyEvent { action=ACTION_DOWN, keyCode=KEYCODE_HEADSETHOOK, scanCode=0, metaState=0, flags=0x8, repeatCount=0, eventTime=617362637000000, downTime=617362634000000, deviceId=-1, source=0x101, displayId=-1 }]` 
Its always KEYCODE_HEADSETHOOK, always repeatCount=0.  I've traced through the code in MediaSessionImpl, I see right after callback.onMediaButtonEvent, it does a swicth with a case for KEYCODE_HEADSETHOOK.  With the info I have, it's hard to know which branch its hits there.  One branch tries to play/pause without calling onPlaybackResumption (so I haven't loaded anything to play). I see doubleTapCompleted is true, then the KEYCODE_HEADSETHOOK will be changed to KEYCODE_MEDIA_NEXT in applyMediaButtonKeyEvent (called at the bottom MediaSessionImpl.onMediaButtonEvent).   Otherwise KEYCODE_HEADSETHOOK doesn't appear to get handled.  I may well have missed something in my walk through on the code, but regardless, the bottom line is onPlaybackResumption is never called, so we don't even attempt to play anything.  Thus the timeout.

### Bug Report

johngray1965:
Fixes #2768

### DIFF
--- a/libraries/test_session_current/src/androidTest/java/androidx/media3/session/MediaSessionTest.java
+++ b/libraries/test_session_current/src/androidTest/java/androidx/media3/session/MediaSessionTest.java
@@ -652,6 +652,8 @@ public class MediaSessionTest {
     player.awaitMethodCalled(MockPlayer.METHOD_SEEK_TO_NEXT, TIMEOUT_MS);
     player.awaitMethodCalled(MockPlayer.METHOD_SEEK_TO_PREVIOUS, TIMEOUT_MS);
     player.awaitMethodCalled(MockPlayer.METHOD_STOP, TIMEOUT_MS);
+    player.awaitMethodCalled(MockPlayer.METHOD_PLAY, TIMEOUT_MS);
+    player.awaitMethodCalled(MockPlayer.METHOD_PLAY, TIMEOUT_MS);
     assertThat(callerCollectorPlayer.callingControllers).hasSize(9);
     for (ControllerInfo controllerInfo : callerCollectorPlayer.callingControllers) {
       assertThat(session.get().isMediaNotificationController(controllerInfo)).isTrue();
--- a/libraries/test_session_current/src/androidTest/java/androidx/media3/session/MediaSessionTest.java
+++ b/libraries/test_session_current/src/androidTest/java/androidx/media3/session/MediaSessionTest.java
@@ -16,6 +16,7 @@
 package androidx.media3.session;
 
 import static android.os.Build.VERSION.SDK_INT;
+import static android.view.KeyEvent.KEYCODE_HEADSETHOOK;
 import static android.view.KeyEvent.KEYCODE_MEDIA_FAST_FORWARD;
 import static android.view.KeyEvent.KEYCODE_MEDIA_NEXT;
 import static android.view.KeyEvent.KEYCODE_MEDIA_PAUSE;
@@ -634,6 +635,14 @@ public class MediaSessionTest {
                       impl.onMediaButtonEvent(
                           controllerInfo, getMediaButtonIntent(KEYCODE_MEDIA_STOP)))
                   .isTrue();
+              assertThat(
+                      impl.onMediaButtonEvent(
+                          controllerInfo, getMediaButtonIntent(KEYCODE_MEDIA_PLAY_PAUSE)))
+                  .isTrue();
+              assertThat(
+                      impl.onMediaButtonEvent(
+                          controllerInfo, getMediaButtonIntent(KEYCODE_HEADSETHOOK)))
+                  .isTrue();
             });
 
     player.awaitMethodCalled(MockPlayer.METHOD_PLAY, TIMEOUT_MS);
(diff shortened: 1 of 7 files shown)


johngray1965:
@marcbaechinger Here you go

marcbaechinger:
Many thanks!

johngray1965:
@marcbaechinger Thank you.  I won't change anything unless I'm asked to.

libraries/test_session_current/src/androidTest/java/androidx/media3/session/MediaSessionTest.java
```diff
@@ -652,8 +652,6 @@ controllerInfo, getMediaButtonIntent(KEYCODE_HEADSETHOOK)))
     player.awaitMethodCalled(MockPlayer.METHOD_SEEK_TO_NEXT, TIMEOUT_MS);
     player.awaitMethodCalled(MockPlayer.METHOD_SEEK_TO_PREVIOUS, TIMEOUT_MS);
     player.awaitMethodCalled(MockPlayer.METHOD_STOP, TIMEOUT_MS);
-    player.awaitMethodCalled(MockPlayer.METHOD_PLAY, TIMEOUT_MS);
```
marcbaechinger: just FYI: This two are redundant as they are asserted on L648 already. It does only assert whether the method was called once. Said this, it's apparent that the test isn't a very good test, Or at least it is misleading. 

Removing the last two calls:

```
assertThat(
                      impl.onMediaButtonEvent(
                          controllerInfo, getMediaButtonIntent(KEYCODE_MEDIA_PLAY_PAUSE)))
                  .isTrue();
              assertThat(
                      impl.onMediaButtonEvent(
                          controllerInfo, getMediaButtonIntent(KEYCODE_HEADSETHOOK)))
                  .isTrue();
```

only fails L657 which asserts that 9 controllers where calling. There isn't a way to assert that these two calls have called play a second and third time.

marcbaechinger:
Thanks John. I added a couple of test cases to make it obvious what your fix actually fixed.

The test that your change makes pass is the unit test I added in `MediaSessionServiceTest`. With your change, an `Intent` that is sent to `onStartCommand()` of the service is now handled just like a `PLAY_PAUSE` command. As you already explained, if playback resumption is triggered with a `KEY_HEADSETHOOK` instead of `PLAY_PAUSE` or `PLAY` this didn't trigger `play()` on the player. In such a case the service wouldn't be started into the foreground which may well be the reason for the remaining crashes we were seeing.

I've added the test cases and sent that for review. The PR will be merged as soon as this is done. Please continue not adding further changes. :)

Thanks a lot for that contribution. Very much appreciated!

johngray1965:
@marcbaechinger Thank you.  I was tempted to add more tests myself, but I was trying to keep the PR as small as possible.

johngray1965:
@marcbaechinger the new tests look great, one suggestion, MediaButtonReceiver checks that the keyCode is KeyEvent.KEYCODE_MEDIA_PLAY, KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE, or KeyEvent.KEYCODE_HEADSETHOOK.  Why not have the parameterized test do all three (I believe its not doing KeyEvent.KEYCODE_MEDIA_PLAY).  I imagine that covered elsewhere, but why not cover it here too for completeness.

marcbaechinger:
Yeah, can do. I'll do that in a follow up CL if I get the approval. It's a bit easier for me then routing everything through GitHub and back to our internal repo. Will do that just after I have submitted internally. How does that sound?

johngray1965:
@marcbaechinger sounds good.  I feel like the tests not exercising all the options is how this originally slipped through the cracks.

marcbaechinger:
Found another warning, so here we go. :)

PR #2819, opened 2025-10-03
Ability to pass a FactoryInstance to WaveformAudioBufferSink which will create custom WaveformBar instances

SathyaTadepalli:
If a consumer app wants to override the behavior of WaveformBar it is not possible, because the creation is happening internally in WaveformAudioBufferSink.
One option is to have a override constructor for WaveformAudioBufferSink, which takes an Factory that will be responsible for creating WaveformBar. 
Consumer app can pass this factory instance and create custom WaveformBar instances.
The current constructor which is present will be still available and and the current WaveformBar will be created.
There will be no impact in current behavior

Added unit test for the custom factory.


tonihei:
Could you check that the CLA is signed (see comment above) so that we can take a look? Thanks?

@ivanbuper Also assigning to you since to have a look once the CLA is there.

PR #2823, opened 2025-10-06
fix(uri): handle '@' correctly

0xtun4:
### Bug

`UriUtil.removeUserInfo()` uses regex split based on `userInfo + "@"`, which fails when the user info contains special characters such as `@`, `+`, or `.`.  
This can lead to `ArrayIndexOutOfBoundsException` or incorrect authority parsing.

Examples that crash:
rtsp://admin:pa%40%40@example.net

Here, the `%40` sequences are decoded into `@` **before** the split occurs, so the code ends up trying to split on a string that no longer matches the original encoded authority. This breaks when user info itself contains `@`.

### Fix

Replaced the manual split with `uri.getHost()` and `uri.getPort()` to rebuild the authority.  
Also added validation for cases where `userInfo` is non-null but `@` is missing.

This approach is RFC-compliant, works with multiple `@` or IPv6 hosts, and avoids regex issues.

### DIFF
--- a/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMessageUtil.java
+++ b/libraries/exoplayer_rtsp/src/main/java/androidx/media3/exoplayer/rtsp/RtspMessageUtil.java
@@ -192,9 +192,14 @@ import java.util.regex.Pattern;
     }
 
     // The Uri must include a "@" if the user info is non-null.
-    String authorityWithUserInfo = checkNotNull(uri.getEncodedAuthority());
-    checkArgument(authorityWithUserInfo.contains("@"));
-    String authority = Util.split(authorityWithUserInfo, "@")[1];
+    String authorityWithUserInfo = uri.getAuthority();
+    checkArgument(authorityWithUserInfo != null && authorityWithUserInfo.contains("@"),
+        "Invalid URI: userInfo is non-null but '@' delimiter is missing");
+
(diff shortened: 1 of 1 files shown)


icbaker:
> Here, the `%40` sequences are decoded into `@` **before** the split occurs

Is this true even after https://github.com/androidx/media/pull/1138 was merged? (note there's also a brief discussion in there about switching to `getHost()` and `getPort()`)

We have a test case with `%40` that passes:

https://github.com/androidx/media/blob/bfe5930f7f29c6492d60e3d01a90abd3c138b615/libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/RtspMessageUtilTest.java#L453-L456

0xtun4:
> > Here, the `%40` sequences are decoded into `@` **before** the split occurs
> 
> Is this true even after #1138 was merged? (note there's also a brief discussion in there about switching to `getHost()` and `getPort()`)
> 
> We have a test case with `%40` that passes:
> 
> https://github.com/androidx/media/blob/bfe5930f7f29c6492d60e3d01a90abd3c138b615/libraries/exoplayer_rtsp/src/test/java/androidx/media3/exoplayer/rtsp/RtspMessageUtilTest.java#L453-L456

Ah, that explains it — I'm still on 1.4.1, so the `%40` case is still decoded early in that version. Thanks for the reference to #1138 and the test, I'll upgrade to the latest to get the fix.

Issue #2484, opened 2025-05-29
HLS Interstitial support for live sliding window


TomVarga:
### HLS Interstitial support for live sliding window
HLS Interstitial should work for live sliding window content too.

### Proposed solution
There are probably many things that need to change to make this happen, probably the most difficult is the timeline handling as the start position shifts with every playlist update.

tonihei:
I believe this is already supported except for a few special tags (on the main branch, to be released with 1.8.0). @marcbaechinger will be able to confirm.

marcbaechinger:
In the main branch, there is support for live with `HlsInterstitialsAdsLoader`.

I'm not sure if this is what you are looking for. From past post my understanding is that you are seeking to support HLS interstitials on your own by reading the manifest/playlist.

So if you want to use the implementation for `HlsInterstitialsAdLoader` then you'd need to look into the main branch as this isn't released yet.

If your request is only about having interstitials parsed in the playlist this should work already with 1.6 because this is done by core ExoPlayer, not the ads loader.

> the timeline handling as the start position shifts with every playlist update.

This only affect the window position and not the period position. The period position is monotonically increasing. Roughly, the window position is then calculated by using the playlist start position that can be subtracted from the period position.

> should work for live sliding window content too

Can you elaborate your report a bit around what exactly isn't working?

TomVarga:
Please bare with me, I need a bit of time to do a setup in the demo app to show/double check that my issue is still their on main. (I am also going on vacation next week, so this might take longer)

marcbaechinger:
> oing on vacation next week

Me too. :)

TomVarga:
Hi @marcbaechinger 

Hope you had a great vacation.

I had a bit of time to look into this and it looks like playback at least works but I am not sure about a few things:

I'd appreciate some guidance on a couple of things:

1) I am not immediately seeing how can I skip an ad? 
Looked like as if getAdsResumptionStates would be where I get the adPlaybackState, but that looks like it is ignored for live.

2) Looks like when the stream is seeked past an interstitial (for example to start at live position) then an interstitial is always played first. I am not immediately seeing how to prevent this besides something like keeping state myself and returning empty asset list for the given interstitial.

TomVarga:
There seems to be an issue with live streams when more than one interstitial is present.

I have a couple of test streams that allowed me to narrow down the potential issu:

This is an AOD (has END tag) stream that works:
[...]

PR #2826, opened 2025-10-07
strict matching of segment boundary

TomVarga:
This issue is spawned from https://github.com/androidx/media/issues/2484#issuecomment-3324175439

summary:

at

https://github.com/androidx/media/blob/d0e41339467a523e58d64e297f9b2b28924ce6cd/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/HlsInterstitialsAdsLoader.java#L1524-L1527

using `mediaPlaylist.targetDurationUs` can lead to identifying the wrong index as it's not guranteed that all segments will match the target duration. Especially around interstitial boundaries it's happens often that the actual segment duration is vastly different. This happening a lot of time in a stream can lead to identifying the wrong target segment.

The proposed solution instead looks at the actual segment duration to find the best match.

### DIFF
--- a/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/HlsInterstitialsAdsLoader.java
+++ b/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/HlsInterstitialsAdsLoader.java
@@ -38,13 +38,13 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 import static java.lang.Math.abs;
 import static java.lang.Math.max;
-import static java.lang.Math.min;
 
 import android.content.Context;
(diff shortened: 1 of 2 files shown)


TomVarga:
@marcbaechinger do please let me know if you think this needs further improvement.

marcbaechinger:
Thanks Tamás! Yup, will take that asap.

marcbaechinger:
Looks nice, thank you! 

marcbaechinger:
Many thanks for the pull. It's merged now. 

I've added some unit test that actually didn't test your new code, but it was convenient now that the method is visible for testing. I also think the test cases may not happen right now for valid playlist, as the data range need to be within the available playlist time range. However, we consider adding programmatic Interstitial insertion through an API on the ads loader. For such cases, these checks may be useful

Thanks again!

PR #2832, opened 2025-10-09
Migrate everything to media3's Log

nift4:
To consistently allow apps to intercept logs and just be generally consistent in the project.


icbaker:
The demo apps are designed to be copy-paste starting points for people building their own apps, and we don't want those devs to use media3's `Log` class in their apps (they should just use the framework `Log` class).

nift4:
Makes sense, undid the demo app changes.

icbaker:

nift4:
Thanks :)

PR #2840, opened 2025-10-10
Parse ReplayGain data from LAME Xing/Info header

nift4:
see http://gabriel.mp3-tech.org/mp3infotag.html#replaygain

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3InfoReplayGain.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3InfoReplayGain.java
@@ -15,7 +15,6 @@
  */
 package androidx.media3.extractor.mp3;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static java.lang.annotation.ElementType.TYPE_USE;
 
 import androidx.annotation.IntDef;
@@ -31,17 +30,6 @@ import java.util.Objects;
 /** Representation of the ReplayGain data stored in a LAME Xing or Info frame. */
 @UnstableApi
 public final class Mp3InfoReplayGain implements Metadata.Entry {
-  /**
-   * 32 bit floating point "Peak signal amplitude".
-   *
-   * <p>1.0 is maximal signal amplitude store-able in decoding format. 0.8 is 80% of maximal signal
-   * amplitude store-able in decoding format. 1.5 is 150% of maximal signal amplitude store-able in
-   * decoding format.
-   *
-   * <p>A value above 1.0 can occur for example due to "true peak" measurement. A value of 0.0 means
-   * the peak signal amplitude is unknown.
-   */
-  public final float peak;
 
   /** A gain field can store one gain adjustment with name and originator metadata. */
   public static final class GainField {
@@ -191,6 +179,18 @@ public final class Mp3InfoReplayGain implements Metadata.Entry {
   }
 
   /**
+   * 32 bit floating point "Peak signal amplitude".
+   *
+   * <p>1.0 is maximal signal amplitude store-able in decoding format. 0.8 is 80% of maximal signal
+   * amplitude store-able in decoding format. 1.5 is 150% of maximal signal amplitude store-able in
+   * decoding format.
+   *
+   * <p>A value above 1.0 can occur for example due to "true peak" measurement. A value of 0.0 means
+   * the peak signal amplitude is unknown.
+   */
+  public final float peak;
+
+  /**
    * The first of two gain fields in the LAME MP3 Info header, or null if the field is invalid or
    * unset.
    */
@@ -203,7 +203,6 @@ public final class Mp3InfoReplayGain implements Metadata.Entry {
   @Nullable public GainField field2;
 
   private Mp3InfoReplayGain(float peak, @Nullable GainField field1, @Nullable GainField field2) {
-    checkArgument(field1 != null || field2 != null);
     this.peak = peak;
     this.field1 = field1;
     this.field2 = field2;


libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3InfoReplayGain.java
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * NAME of Gain adjustment in the first field, also called "Radio Replay Gain" field:
+   *
+   * <p>b000 = not set
+   *
+   * <p>b001 = radio
+   *
+   * <p>b010 = audiophile
+   */
+  public final byte field1Name;
```
icbaker: We don't need to perfectly match the spec here - it might be a clearer API to have an IntDef that reflects the three possible values of unset, radio & audiophile.

Similarly, maybe a nested type structure would be clearer than `name1A`, `name1B`, `name2A` etc.

Something like:

```java

public final class Mp3InfoReplayGain {

  public static final class GainField {
    public final @Name int name;
    public final @Originator int originator;
    public final int gain;
  }
  public final float peak;
  public final GainField field1;
  public final GainField field2;
}
```

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/XingFrame.java
```diff
@@ -35,6 +36,44 @@
    */
   public final long dataSize;
 
+  /** Whether this frame is the LAME variant of a Xing frame, and hence has ReplayGain data. */
+  public final boolean hasReplayGain;
```
icbaker: Similar comment here, instead of a boolean controlling the presence of all the other fields, move them into a type that can then be nullable.

Maybe you can even re-use the `Metadata.Entry` subclass here? Instead of duplicating all the fields.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3InfoReplayGain.java
nift4: Thanks for the feedback, that makes a lot of sense and I implemented that.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/XingFrame.java
nift4: Similar to above, that works great and is now implemented.

nift4:
Hi @icbaker, I had pushed a new commit to address the review comments, could you please review again? Thanks! :)

icbaker:
The change looks generally good.

I ran all the tests in `Mp3ExtractorTest` and `Mp3PlaybackTest` with the change, and nothing failed (i.e. it doesn't find any replaygain data in our existing MP3 test assets).

Can you include a very small MP3 test asset in this PR that we can use to test the extraction?

Ideally by re-encoding or re-muxing one of the existing test assets in our library.

nift4:
Hi @icbaker,

That is odd, I assume the additional metadata log line should've triggered a difference in any case. With EventLogger I see that 9 files (one of them is `libraries/test_data/src/test/assets/media/mp3/bear-vbr-xing-header.mp3`) have empty data (common for non-LAME encoders writing LAME headers)

```
D    Metadata [
D      TSSE: description=null: values=[Lavf58.29.100]
D      ReplayGain Xing/Info: peak=0.0, f1 name=0, f1 orig=0, f1 val=0.0, f2 name=0, f2 orig=0, f2 val=0.0
D    ]
```

What indeed seems missing is an file with actual data, so I have re-encoded one of the samples with `lame <file>` (with the default replaygain-fast mode) and with `lame --replaygain-accurate <file>` in the new patchset using `LAME 64bits version 3.100 (http://lame.sf.net)`.

icbaker:
Ah sorry, I ran the tests in the wrong IDE - so not pointing at the branch with your pending changes... When I run them in the correct place, some do fail (but yes, as you say with 'zero' values).

But it's still good to have a test for the non-zero case, so thanks for providing that.

I've updated the test dump files, and added new ones for the two new files.

I've also made a few tweaks to the implementation (mostly re-ordering).

nift4:
@icbaker Regarding isValid: the spec explicitly advises against interpreting unknown field names, and isValid() doesn't prevent apps from manually performing a similar check if someone really does make a new field name. But as only these two field names are in use for the better part of two and a half decades, I don't think it's very likely, which is why I think isValid() provides some value by suggesting users to check the name is not 0 (in which case all data in the gain field is junk) nor invalid.

If you still think it shouldn't be there, fine with me, but I wanted to mention this extra context.

icbaker:
I think we should either parse the data that's there and expose it to the developer to choose what to do with (current state with `isValid` removed).

Or confidently say that only radio & audiophile will ever be seen, and everything else should be ignored, in which case we just skip any other values and don't expose them at all. In this case there's no need for `isValid` we just have null values either for one of the `GainField` fields or we just don't emit the `Mp3InfoReplayGain` entry at all.

If ExoPlayer was also implementing the handling of this data, then I'd lean more strongly towards the second approach (don't parse what we don't understand / can't use).

In the current code, the parsing & emitting is the **only** thing that ExoPlayer is doing, and so emitting whatever we find seems reasonable.

nift4:
Makes sense.

> If ExoPlayer was also implementing the handling of this data

That would be nice, but for it to work in offload we'd need an implementation based on DynamicsProcessing and for devices below Android 9 there'd need to be an AudioProcessor based one (if implementing the full spec, it would need a dynamic range compression software implementation, which is hard to do in a performant way in Java). In addition there's 7 ways I know of ReplayGain information can be stored in just MP3 files alone simply due to the long history of both (ID3 TXXX REPLAYGAIN_* frames (modern software), ID3 RVA2/XRVA frames (ID3v2.4), ID3 RVAD/RVA (ID3v2.3), ID3 comment frame (mpg123), ID3 RGAD frame (2002 ReplayGain spec), Xing header (LAME) - this PR, and APE tags (mp3gain)), it's an incredibly complex thing to do and probably won't ever happen (if there is interest besides all that, I'm happy to work on it though).

Today there aren't even all the pieces for an app side implementation alone, as LAME parsing (this PR), APE tag parsing (plan to code and upstream later), and disabling gapless if required (discussed in #2855) are required, so I'm focusing on that for now.

icbaker:

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3InfoReplayGain.java
```diff
@@ -0,0 +1,256 @@
...
+   * Parses an instance from the packed representation.
+   *
+   * <p>Returns null if both {@code field1} and {@code field2} are invalid or should be ignored.
+   */
+  @Nullable
+  public static Mp3InfoReplayGain parse(float peak, int field1, int field2) {
+    GainField parsedField1 = GainField.parse(field1);
+    GainField parsedField2 = GainField.parse(field2);
+    if (parsedField1 == null && parsedField2 == null) {
+      return null;
```
nift4, line 221: I think this should also check for peak == 0f. I am not sure if there are files out there that have peak set but no fields, but it would be more correct I suppose.

icbaker: Done, thanks

I included a negative check too since it looks like that's possible in the binary representation but meaningless based on my reading of the peak field docs in this PR - does that look right to you as well?

nift4: Yeah, I think you're right, if it's negative it's entirely nonsense.

nift4: Oh, I intended to say I think it should be `if (peak <= 0 && parsedField1 == null && parsedField2 == null) {`. There are valid files with no peak but with gain field, for example the replaygain-fast sample

icbaker: Ah right, sorry - thanks, fixed now

nift4:
Thanks :)

PR #2861, opened 2025-10-23
Create scala.yml

miansargodhian786-arch:


icbaker:
We don't use scala in this project, this doesn't look relevant

PR #2868, opened 2025-10-28
Allow space as a date-time separator in `Util.parseXsDateTime`

JaroslavHerber:
The `XS_DATE_TIME_PATTERN` in `Util.java` was previously limited to parsing ISO 8601 date-times that use a literal 'T' or 't' as the separator between the date and time components.

However, some content providers and metadata sources use a space character as the separator (e.g., "2025-10-28 12:09:41.913000+00:00"). This caused `Util.parseXsDateTime` to fail when parsing these valid, albeit non-standard, date-time strings, leading to potential parsing errors for certain media streams.

This commit updates the regular expression by adding a space to the character class `[Tt]`, making it `[Tt ]`. This change extends parsing support to include date-times with a space separator while maintaining full compatibility with the standard 'T'/'t' separators.

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/util/Util.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/Util.java
@@ -190,7 +190,7 @@ public final class Util {
   private static final String TAG = "Util";
   private static final Pattern XS_DATE_TIME_PATTERN =
       Pattern.compile(
-          "(\\d\\d\\d\\d)\\-(\\d\\d)\\-(\\d\\d)[Tt]"
+          "(\\d\\d\\d\\d)\\-(\\d\\d)\\-(\\d\\d)[Tt ]"
               + "(\\d\\d):(\\d\\d):(\\d\\d)([\\.,](\\d+))?"
               + "([Zz]|((\\+|\\-)(\\d?\\d):?(\\d\\d)))?");
   private static final Pattern XS_DURATION_PATTERN =


JaroslavHerber:
Example stream: https://sdn-global-live-streaming-packager-cache.3qsdn.com/64733/64733_264_live.m3u8

icbaker:
The HLS spec defines `date-time-msec` as
> date-time-msec is an ISO/IEC 8601:2004 [[ISO_8601](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-16#ref-ISO_8601)] date/time representation, such as YYYY-MM-DDThh:mm:ss.SSSZ.

https://en.wikipedia.org/wiki/ISO_8601 claims the space separator is permitted in RFC 3339 but **not** ISO 8601:
> Separating date and time parts with other characters such as space is not allowed in ISO 8601, but allowed in its profile RFC 3339.

That said, it seems likely that other content producers will make the same mistake and use an RFC 3339 space-separated date-time instead of the `T`-separated format mandated by the HLS spec (via ISO 8601). So although I'd argue these streams are invalid per the HLS spec, we can accept this PR to parse them without too much risk of unexpected consequences.

icbaker:
Please can you re-send this PR against the `main` branch - we cannot merge it into the `release` branch.

JaroslavHerber:
Will resend in main branch

PR #2875, opened 2025-10-31
Add ELF alignment option for iamf target

thonsi:
more details https://github.com/AOMediaCodec/libiamf/pull/147

### DIFF
--- a/libraries/decoder_iamf/src/main/jni/CMakeLists.txt
+++ b/libraries/decoder_iamf/src/main/jni/CMakeLists.txt
@@ -48,3 +48,5 @@ target_link_libraries(iamfJNI
 # Enable 16 KB ELF alignment.
 target_link_options(iamfJNI
                     PRIVATE "-Wl,-z,max-page-size=16384")
+target_link_options(iamf
+                    PRIVATE "-Wl,-z,max-page-size=16384")


rohitjoins:
Hi @thonsi,

Thank you for raising this pull request. Based on the version of [AGP](https://github.com/androidx/media/blob/main/build.gradle#L20) used, we should not need to add these compiler options to align to 16 KB, please see the [doc](https://developer.android.com/guide/practices/page-sizes#agp_version_851_or_higher).

Based on the screenshot in https://github.com/AOMediaCodec/libiamf/pull/147#issuecomment-3474064556, it seems you're using an older AGP version to build this as the shared libraries are compressed. Can you please upgrade the version and try again?

We would also remove the compiler options passed for the JNI, which is now redundant.

thonsi:
I'm using this. 

Android Studio Otter = 2025.2.1
agp = "8.13.0"
ndkVersion = "29.0.14206865"
buildToolsVersion = "36.0.0"
distributionUrl=https\://services.gradle.org/distributions/gradle-9.2.0-bin.zip

rohitjoins:
I have AGP version `8.12.3` and after building via Android Studio I can see them 16 KB aligned.

5b6PfARgNWQuSSn

thonsi:
x86 doesn't have that problem. Kindly check other architectures.

rohitjoins:
I have verified the same on `x86_64` and `arm64-v8a`. Please see the screenshots.

Screenshot 2025-11-04 at 11 42 27
Screenshot 2025-11-04 at 11 44 26

PR #2878, opened 2025-11-04
Allow space as a date-time separator in Util.parseXsDateTime

JaroslavHerber:
The `XS_DATE_TIME_PATTERN` in `Util.java` was previously limited to parsing ISO 8601 date-times that use a literal 'T' or 't' as the separator between the date and time components.

However, some content providers and metadata sources use a space character as the separator (e.g., "2025-10-28 12:09:41.913000+00:00"). This caused `Util.parseXsDateTime` to fail when parsing these valid, albeit non-standard, date-time strings, leading to potential parsing errors for certain media streams.

This commit updates the regular expression by adding a space to the character class `[Tt]`, making it `[Tt ]`. This change extends parsing support to include date-times with a space separator while maintaining full compatibility with the standard 'T'/'t' separators.

Example stream: https://sdn-global-live-streaming-packager-cache.3qsdn.com/64733/64733_264_live.m3u8

Re-send into main from this:
https://github.com/androidx/media/pull/2868#issue-3561376744

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -34,6 +34,7 @@
     *   Add `HlsInterstitialsAdsLoader.Listener.onAdStarted` to report the start
         of an ad period
         ([#2859](https://github.com/androidx/media/issues/2859)).
+    *   Accept space as a date/time separator in ISO 8601 date-time strings.
 *   DASH extension:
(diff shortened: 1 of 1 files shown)


icbaker:

PR #2883, opened 2025-11-08
Fix compilation error with android.nonTransitiveRClass=true

h6ah4i:
This change fixes a compilation error that occurs when `android.nonTransitiveRClass=true` is enabled in `gradle.properties`.

### DIFF
--- a/libraries/session/src/main/java/androidx/media3/session/MediaSessionImpl.java
+++ b/libraries/session/src/main/java/androidx/media3/session/MediaSessionImpl.java
@@ -2243,11 +2243,11 @@ import org.checkerframework.checker.initialization.qual.Initialized;
         int maxWidth =
             context
                 .getResources()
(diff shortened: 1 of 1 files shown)


icbaker:
Can I just check that this only causes problems if you are [building the library locally](https://github.com/androidx/media/blob/release/README.md#locally)? 

I tried to reproduce the problem with a project depending on `implementation("androidx.media3:media3-session:1.9.0-alpha01")` and didn't see an issue.

icbaker:
Actually after looking more closely at this code, these are private resource IDs we shouldn't be referencing - I removed the reference in https://github.com/androidx/media/commit/420caf5ea054a2ee6a3c9ca43993d1bb019e49a5.

So I'll close this PR as obsolete.

icbaker:
I have also removed this setting from our `gradle.properties` file so we shouldn't introduce any similar problems in future: https://github.com/androidx/media/commit/b770edfb6e91325151fed7ae996affb85ad84cc9

PR #2886, opened 2025-11-11
Use streaminfo block for FLAC files.

Tolriq:
Some files do not have the sample size in each frame headers, in that case, ffmpeg needs the streaminfo data to properly decode them.

Issue: #2887


rohitjoins:

Issue #2910, opened 2025-11-24
GlUtil.java checks for GL errors as part of creating EGL resources even when no valid GL context can exist.


k3wlbuddy:
### Version

Media3 main branch

### More version details

main

### Devices that reproduce the issue

Not device specific and should reproduce on any Android device.

### Devices that do not reproduce the issue

N/A. 

### Reproducible in the demo app?

Not tested

### Reproduction steps

Any application that depends on media3 utilities that invoke the helper functions in `GlUtil.java`.

The repro step we've been using on our side is running the `CtsMediaEditingTestCases.TranscodeQualityTest` part of the Android CTS which ends up using `media3.transformer`

Steps to run the Android `CtsMediaEditingTestCases.TranscodeQualityTest` CTS:

- Get Android CTS from `https://dl.google.com/dl/android/cts/android-cts-16_r3-linux_x86-arm.zip`
- Unzip the CTS.
- Get the required media files for the CTS from `https://dl.google.com/android/xts/cts/tests/tests/mediaediting/CtsMediaEditingTestCases-1.1.zip`
- Unzip the media files and push to device.
- Install the `CtsMediaEditingTestCases.apk` from the downloaded CTS
- Run the CTS as `adb shell am instrument -w -e class android.media.mediaediting.cts.TranscodeQualityTest android.media.mediaediting.cts/androidx.test.runner.AndroidJUnitRunner -log-level FINEST`
- Run `logcat | grep libEGL` to see the print ` libEGL  : call to OpenGL ES API with no current context (logged once per thread)`

### Expected result

The log `libEGL  : call to OpenGL ES API with no current context (logged once per thread)` should ideally not appear in logcat and `GlUtil.java` should not be throwing GlExceptions for it.

For the above to happen,  within `GlUtil.java`, invocations to `checkGlError` should not be invoked before a valid GL context is present/made current.

A valid GL context is considered to be made current by AOSP's `libEGL.so` when the application makes a EGLContext current by invoking `eglMakeCurrent`. But `GlUtil.java` is invoking `checkGlError` as part of  `createEglContext` and `getDefaultEglDisplay`, both APIs that can be and almost always are invoked without a GLContext present.

The fix would be to not check for GL errors as part of `createEglContext` and `getDefaultEglDisplay`

### Actual result

logcat should not contain the print `libEGL  : call to OpenGL ES API with no current context (logged once per thread)` and GlUtil.java should not be invoking checkGlErrors untill a valid GLContext is present and made current.

[...]

PR #2912, opened 2025-11-24
media3: utils: Do not check for GL error without GL context/

k3wlbuddy:
GlUtil.java's helper functions createEglContext and getDefaultEglDisplay unconditionally invoke checkGlError for checking any GL errors.

Unfortunately, since a valid EGLContext would not have been made current before executing the aformentioned helper functions, the function pointer for the GL API glGetError would be pointing to a default "gl_no_context" function inside AOSP's libEGL.so that will log a "call to OpenGL ES API with no current context" message in logcat.

Only as part of eglMakeCurrent does libEGL.so replace all GL API function pointers to "gl_no_context" with pointers to the actual functions implementing the APIs.

This PR removes the invocation of checkGlError as part of createEglContext and getDefaultEglDisplay to avoid prints of the aforementioned message from libEGL.so.

Fixes Issue #2910.

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/util/GlUtil.java
+++ b/libraries/common/src/main/java/androidx/media3/common/util/GlUtil.java
@@ -276,7 +276,6 @@ public final class GlUtil {
             /* unusedMinor */ new int[1],
             /* minorOffset= */ 0),
         "Error in eglInitialize.");
-    checkGlError();
     return eglDisplay;
   }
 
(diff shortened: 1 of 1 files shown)


k3wlbuddy:
@ychaparov Do you know who can help review this PR? Thanks!

ychaparov:
Sorry about that, I already had https://github.com/androidx/media/commit/f68b9dd9333f8fe1f6e1ea0fdc4c0d9240388a53 under review when you filed https://github.com/androidx/media/issues/2910 and before you sent this PR

The issue should be fixed on the main branch already

PR #2926, opened 2025-12-03
Remove @Nullable from MediaButtonReceiver.onReceive()

kelmer44:
The [docs](https://developer.android.com/reference/android/content/BroadcastReceiver#onReceive(android.content.Context,%20android.content.Intent)) for `onReceive` state:

> This method is called when the BroadcastReceiver is receiving an Intent broadcast.

So I am proposing to remove this qualifier. I wonder if it's possible to get this method called with a null parameter, other than clients calling this method directly, in which case I think enforcing a non-nullable parameter is all the better.

This prevents all extending kotlin classes from having to nullsafe the intent parameter before continuing.

### DIFF
--- a/libraries/session/src/main/java/androidx/media3/session/MediaButtonReceiver.java
+++ b/libraries/session/src/main/java/androidx/media3/session/MediaButtonReceiver.java
@@ -124,6 +124,7 @@ public class MediaButtonReceiver extends BroadcastReceiver {
    */
   @Override
   public void onReceive(Context context, Intent intent) {
+    checkNotNull(intent);
     handleIntentAndMaybeStartTheService(context, intent);
   }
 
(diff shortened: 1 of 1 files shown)


kelmer44:
I think i confused CI by pointing to the wrong branch 👀

libraries/session/src/main/java/androidx/media3/session/MediaButtonReceiver.java
```diff
@@ -123,7 +123,7 @@ public class MediaButtonReceiver extends BroadcastReceiver {
    * strongly recommended way to handle the intent is using the default implementation.
    */
   @Override
-  public void onReceive(Context context, @Nullable Intent intent) {
+  public void onReceive(Context context, Intent intent) {
```
marcbaechinger, line 126: As far as I can tell, if there is no @Nullable or @NotNull annotation on Android framework code, then there is no nullability information available and we can't assume this is non-null. 

Agreed the doc is clear about it and we can't do something sensible without an `Intent`. Mind adding `checkNotNull(intent)` when passing it further down. Sorry for being overly picky, I see we would have a NPE right after but I think I'd prefer this being consistently where we get the garbage input. Shouldn't hurt as we agree this doesn't happen.

kelmer44: done

PR #2946, opened 2025-12-16
Treat hev2/hvc2 as a valid HEVC base for Dolby Vision Profile 20

ybai001:


ybai001:
Dolby Vision Profile 20 is for HMD device. It is based on Multiview HEVC. You may consider DV profile 20 = Multiview HEVC (base layer) + Dolby Vision metadata.

ybai001:
Hi, @microkatz, could you help to have a look on this PR? Thanks.

Issue #2968, opened 2025-12-26
TS Container with Multiple Audio Tracks


osscv:
Problem: it was able to detect the DVB Subtittle in the TS Container after some effrot, but when select
the tracks, it was not able to show the DVB Subtittle properly (nothing shows).

Did anyone knows how to optimize or able to play it properly? (As MPV Player [Other Player} was able)

osscv:
I do understand that the DVB Subtitles are bitmap-based subtitles.
At present, Media3 does not include a built-in renderer for DVB (bitmap) subtitles, therefore native subtitle rendering is not supported. Is this correct?

FongMi:
Please refer to the template and provide a test url.

oceanjules:
What do you mean by built-in renderer? We have https://github.com/androidx/media/blob/7cc1056f840ce226598d3b990d4a6f7cd17e2831/libraries/extractor/src/main/java/androidx/media3/extractor/text/dvb/DvbParser.java#L47

Cues will have parsed bitmaps and `SubtitleView` inside `PlayerView` will render those.

https://github.com/androidx/media/blob/7cc1056f840ce226598d3b990d4a6f7cd17e2831/libraries/extractor/src/main/java/androidx/media3/extractor/text/dvb/DvbParser.java#L256-L264

We moved subtitle _parsing_ from the renderer "side" to extractor "side".  Have you disabled this flag:
https://github.com/androidx/media/blob/7cc1056f840ce226598d3b990d4a6f7cd17e2831/libraries/extractor/src/main/java/androidx/media3/extractor/ts/TsExtractor.java#L111

.. or are you ising any custom extractor setup? Ideally you would try it out on our `demos/main` with your media URL and get a logcat.

osscv:
@oceanjules Hi, Julia, here is a sample:
LIVE 1: https://o11.163189.xyz/stream/live/aod/ 
LIVE 2: https://o11.163189.xyz/stream/live/qj/
for you to test (please try it as soon as possible, as this stream was founded online, which is a sample of TS containing DVB Subtitles Tracks & Multiple Audio Tracks.


As a side of notes:
In my current implementations on Media3, i was able to detect the subtitle tracks but was unable to show the DVB sub properly for the TS container type (I tested with the MPV player it was able to play & detect the tracks as well.)


Disclaimer: I am not responsible for the content of the streams, as they were obtained from publicly available online sources.

osscv:
[...]

PR #2973, opened 2025-12-31
Enable detection of DVB subtitles and audio tracks in TS container (.ts)

osscv:
Removed early return to enable detection of DVB subtitles and multiple audio tracks in TS container. 

Sloutions for my Issue: https://github.com/androidx/media/issues/2968

@Override
  public void packetStarted(long pesTimeUs, @TsPayloadReader.Flags int flags) {
    if ((flags & FLAG_DATA_ALIGNMENT_INDICATOR) == 0) {
      //return; [Remove this able to Detect DVB Subtittles & Mutiple Audio Tracks in TS Container (.ts)
    }
    writingSample = true;
    sampleTimeUs = pesTimeUs;
    sampleBytesWritten = 0;
    bytesToCheck = 2;
  }


osscv:
image

icbaker:
The code you have commented out is designed to catch & skip unexpected/malformed packets, I'm afraid we are not going to remove it.

PR #2978, opened 2026-01-02
Fix onAudioSessionIdChanged() handling in (Forwarding)SimpleBasePlayer

nift4:
Issue: #3140

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -6,6 +6,8 @@
     *   Add `Format.channelMask` to explicitly represent the audio channel mask,
         and a new `Util.getAudioTrackChannelConfig(Format)` overload to safely
         resolve it.
+    *   Fix handling of `onAudioSessionIdChanged` in `SimpleBasePlayer` and
+        `ForwardingSimpleBasePlayer`.
 *   ExoPlayer:
     *   Mark tracks with a well-formed but unrecognized codec profile or level
         as `supported=NO_EXCEEDS_CAPABILITIES` instead of `supported=YES` (which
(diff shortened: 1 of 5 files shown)


nift4:
Hi @marcbaechinger, can you please check this PR? Thanks!

icbaker:
Please can you add some tests for this

nift4:
@icbaker Thanks for the reminder, I added tests.

Weirdly, ForwardingSimpleBasePlayerTest.getterMethods_noOtherMethodCalls_returnCurrentStateFromWrappedPlayer() was testing ForwardingPlayer instead of ForwardingSimpleBasePlayer, which I think might've been a mistake, so I fixed that. And then a lot of asserts started failing due to wrong available commands so I fixed that too. Please take a look whether that was correct, as it feels a bit odd that it was testing the wrong class?

Edit: also, to be clear, there is already MediaControllerListenerTest.onAudioSessionIdChanged_isCalledAndUpdatesGetter() which covers MediaController, it seems controller itself never was affected and it just happens because both demo and my prod app use ForwardingSimpleBasePlayer.

icbaker:
> Weirdly, ForwardingSimpleBasePlayerTest.getterMethods_noOtherMethodCalls_returnCurrentStateFromWrappedPlayer() was testing ForwardingPlayer instead of ForwardingSimpleBasePlayer, which I think might've been a mistake, so I fixed that. And then a lot of asserts started failing due to wrong available commands so I fixed that too. Please take a look whether that was correct, as it feels a bit odd that it was testing the wrong class?

Thanks for flagging this! I agree it looks wrong. It seemed a bit messy to change the class-under-test in this PR (since it's not really related), so I submitted a separate change to make that fix: https://github.com/androidx/media/commit/4a32661d56fe21a282a0a14146522ce1d046c7e1

I've rebased & re-pushed this PR on top.

icbaker:

nift4:
Thanks for picking up this pull request ^^

Issue #2992, opened 2026-01-08
[Feature Request] Add preferredAudioChannelCount to TrackSelectionParameters


KanghyunJeon:
### Feature Description
Add a new parameter `preferredAudioChannelCount` to `TrackSelectionParameters` that allows users to specify a preferred number of audio channels when selecting audio tracks.

### Use case description
When playing content with multiple audio tracks of varying channel counts (e.g., stereo and 5.1 surround), users may want to explicitly prefer a specific channel configuration. 
For example:
  - A user with a stereo speaker setup may prefer 2-channel audio
  - A user with a surround sound system may prefer 6-channel (5.1) audio
Currently, there is no way to express this preference through `TrackSelectionParameters`. Users can only constrain the maximum channel count via `maxAudioChannelCount`, but cannot specify a preferred channel count.

### Proposed solution
Add `preferredAudioChannelCount` parameter similar to how `preferredAudioLanguage` and `preferredAudioRoleFlags` work:
  1. Add `preferredAudioChannelCount` field to `TrackSelectionParameters`
  2. Add corresponding builder method `setPreferredAudioChannelCount(int channelCount)`
  3. Update `DefaultTrackSelector.AudioTrackScore` to score tracks based on channel count preference
  4. When set, preferred channel count should take priority over role flags (similar to how language takes priority over role flags)

### Priority Order
  The proposed comparison order in `AudioTrackScore`:
  1. Preferred language (highest priority)
  2. Preferred channel count
  3. Preferred role flags
  4. Others...

KanghyunJeon:
Hi, I've created a PR to implement this feature request.

tonihei:
Would you also work to have a `minAudioChannelCount` setting? This would be more in line with other settings like video resolution that provide a min and max to express user preferences.

> A user with a stereo speaker setup may prefer 2-channel audio. A user with a surround sound system may prefer 6-channel (5.1) audio

This should already happen automatically. If the speakers used only support stereo (including Bluetooth), ExoPlayer prefers a stereo track by default assuming it has a better quality optimized for stereo. If the audio output has more channels, ExoPlayer prefers the higher channel tracks.

microkatz:
@KanghyunJeon 

Thank you for creating a ticket to better discuss the enhancement request. I think we require some clarity on what you are requesting. 

[...]

PR #2996, opened 2026-01-08
[Feature] Add preferredAudioChannelCount for audio track selection

KanghyunJeon:
## Summary
  This PR implements the feature requested in #2992.

  Adds a new `TrackSelectionParameters.preferredAudioChannelCount` API that allows apps to prefer audio tracks with higher channel counts (e.g., 5.1, 7.1 surround) over stereo tracks, regardless of role flags (main vs alternate).

  ## Changes
  - Add `preferredAudioChannelCount` field to `TrackSelectionParameters`
  - Add `setPreferredAudioChannelCount(int)` builder method
  - Modify `AudioTrackInfo.compareTo()` to prioritize higher channel count when preference is set
  - Reuse existing `parameters` and `channelCount` fields 

 ## Behavior
  - When `preferredAudioChannelCount > 0`, tracks are compared by channel count before role flags
  - Higher channel count tracks are preferred
  - Falls back to lower channel count if higher is not supported by the renderer
  - Set to `0` (default) to use original behavior based on role flags

  ## Issue
 #2992

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -19,6 +19,9 @@
 *   CompositionPlayer:
 *   Transformer:
 *   Track selection:
+    *   Add `TrackSelectionParameters.preferredAudioChannelCount` to prefer
+        audio tracks with a specific channel count
+        ([#2996](https://github.com/androidx/media/pull/2996)).
 *   Extractors:
 *   Inspector:
 *   Audio:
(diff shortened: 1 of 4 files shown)


microkatz:

microkatz:
@KanghyunJeon 

In reviewing your case further we have additional questions before we can move your PR forward. If you could please answer the questions in the connected Ticket you filed here, https://github.com/androidx/media/issues/2992#issuecomment-4030524997, that would be great!

Thanks!

PR #3014, opened 2026-01-16
Update DefaultActionFactory.java

AndresBC-123:


marcbaechinger:
If you want this pull to be processed please make a pull against the `main` branch rather than the release branch.

For the exact change in that file, can you give us some background on what you are trying to achieve with this public method that gives you the `PendingIntent`? Without any further reasoning I don't think just making this method public on the `DefaultActionFactory`  makes sense.

I'm happy to support you to achieve your goals with our code if you provide more information.

PR #3038, opened 2026-01-30
Ability to pass a FactoryInstance to WaveformAudioBufferSink which will create custom WaveformBar instances

SathyaTadepalli:
If a consumer app wants to override the behavior of WaveformBar it is not possible, because the creation is happening internally in WaveformAudioBufferSink.
One option is to have a override constructor for WaveformAudioBufferSink, which takes an Factory that will be responsible for creating WaveformBar.
Consumer app can pass this factory instance and create custom WaveformBar instances.
The current constructor which is present will be still available and the current WaveformBar will be created.
There will be no impact in current behavior

Added unit test for the custom factory.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/WaveformAudioBufferSink.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/WaveformAudioBufferSink.java
@@ -89,6 +89,13 @@ public class WaveformAudioBufferSink implements TeeAudioProcessor.AudioBufferSin
     void onNewWaveformBar(int channelIndex, WaveformBar waveformBar);
   }
 
+  public static class WaveformBarFactory {
+    /** Called when a new waveform bar has to be created */
+    WaveformBar createWaveformBar() {
+      return new WaveformBar();
+    }
+  }
+
   private final int barsPerSecond;
   private final Listener listener;
   private final SparseArray<WaveformBar> outputChannels;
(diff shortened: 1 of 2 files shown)


icbaker:
This is a duplicate of https://github.com/androidx/media/pull/2998 which was closed because the CLA was not signed after 3 weeks. The CLA still appears to be unsigned. Please sign the CLA.

SathyaTadepalli:
> * cla/google

I have signed it yesterday.

icbaker:
Thanks, I've re-run it and it's passing now.

kim-vde:
Hi - would it be possible to open a new PR from an individual-owned fork? We can't push changes to organization-owned forks like this one unless we have collaborator access. If that's not possible then we can still merge this PR but it will result in an ['evil' merge](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefevilmergeaevilmerge). See more info here: https://github.com/androidx/media/blob/release/CONTRIBUTING.md#push-access-to-pr-branches

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/WaveformAudioBufferSink.java
```diff
@@ -98,6 +105,13 @@ public interface Listener {
   private @MonotonicNonNull ChannelMixingMatrix channelMixingMatrix;
   private int samplesPerBar;
 
+  private WaveformBarFactory waveformBarFactory;
+
+  /**
+   * Default factory instance that creates standard WaveformBar objects.
+   */
+  private static final WaveformBarFactory DEFAULT_WAVEFORM_BAR_FACTORY = new WaveformBarFactory();
```
kim-vde, line 113: We usually use interfaces to allow customization. Could you make WaveformBar an interface?

Issue #2855, opened 2025-10-17
How to prevent offload gapless playback based on metadata?


nift4:
I'm working on a ReplayGain implementation for my music player. To ensure battery usage is low, I want to make offload work with it as well.
ReplayGain handles files with low volume by increasing the volume above 100%. Clipping is considered, so this is only done if a file has a low peak volume and silence in upper part of headroom. Increasing volume above 100% without compression (not needed since after increasing the volume, files still won't clip) can only be achieved using DynamicsProcessing in Android, not using AudioTrack.setVolume() which is capped at 100%. So I need to update DynamicsProcessing parameter synchronized with song change.
However when using gapless offload, it's hard to time the change correctly as it already processes start of next file before current one is done playing. Hence I need to prevent offload gapless if gain changes. However offload gapless should still work if gain is the same (to allow gapless albums to still be played gapless). The gain is stored in Format.Metadata tags.
I am implementing this in a decorating AudioSink to ensure I can update effect settings before frames are written to AudioTrack. Is there any way to prevent offload gapless this way if gain is different?

andrewlewis:
Just to check on this point:

> it's hard to time the change correctly as it already processes start of next file before current one is done playing

I thought the audiofx implementations (including [DynamicsProcessing](https://developer.android.com/reference/android/media/audiofx/DynamicsProcessing)) were implemented in the HAL so any changes take effect almost immediately even when using offload. So you just need to time the reconfiguration of the audio effect to the moment at which the new track starts playing? (Synchronization/timing still won't be perfect, but if it's a seamless, gapless transition then metadata shouldn't change drastically between the pieces of the content.)

To put it another way, I think you need to time the changes based on the player position, not based on the audio writing position (which can be way ahead in media time when using offload).

nift4:
> so any changes take effect almost immediately even when using offload

I am not sure if HAL guarantees that, I would assume HAL uses large-ish buffers internally to save power and hence changes are applied with some latency. My assumption is that if I have not given it any frames, then it cannot have already processed them, so this would allow perfect synchronization as long as there's a gap. But I did not test this properly so far (even if I did, I'd need to test all major SoC vendors seperately for the results to be meaningful, I suppose.)

> (Synchronization/timing still won't be perfect, but if it's a seamless, gapless transition then metadata shouldn't change drastically between the pieces of the content.)

[...]

PR #3039, opened 2026-01-30
Ask the audio output whether it wants to be reused

nift4:
as discussed with @andrewlewis, this is required for ReplayGain in offload mode to sometimes prevent gapless mode to avoid music being too loud for a short amount of time (as there is no way to change effect settings during gapless track transition). To detect the ReplayGain data, Format.metadata is required which is available through FormatConfig parameter.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioOutput.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioOutput.java
@@ -22,6 +22,8 @@ import androidx.media3.common.C;
 import androidx.media3.common.PlaybackParameters;
 import androidx.media3.common.util.UnstableApi;
 import androidx.media3.exoplayer.analytics.PlayerId;
+import androidx.media3.exoplayer.audio.AudioOutputProvider.FormatConfig;
+import androidx.media3.exoplayer.audio.AudioOutputProvider.OutputConfig;
 import java.nio.ByteBuffer;
 
 /** An interface to wrap an object that can play audio, like an {@link AudioTrack}. */
@@ -169,6 +171,13 @@ public interface AudioOutput {
   /** Sets the {@link PlayerId} on the audio output. */
   @UnstableApi
   default void setPlayerId(PlayerId playerId) {}
+
+  /** Returns if the configurations are sufficiently compatible to reuse this audio output. */
+  @UnstableApi
+  default boolean canReuseAudioOutput(
+      OutputConfig currentConfig, FormatConfig newFormat, OutputConfig newConfig) {
+    return newConfig.equals(currentConfig);
+  }
 
   /** Attaches an auxiliary effect to the output. */
   void attachAuxEffect(int effectId);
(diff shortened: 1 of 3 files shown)


tonihei:
I'm not entirely certain this is the right place for this customization. If the `OutputConfig` changes, then it seems there is no chance we can reuse the existing `AudioOutput`, so the only remaining decision point is whether the `AudioOutput` should be recreated despite the exact same config. At this point, there isn't really an obvious signal at the `AudioOutput` abstraction layer that would prevent it from being reused. 

Sounds like you discussed this with @andrewlewis before, but could you describe in more detail what you want to achieve and which API calls you'd to make and then? For example, I'm not sure how'd apply effects to offload playback because they data is still encoded (and PCM offload not yet supported).

nift4:
Hi @tonihei,

> If the OutputConfig changes, then it seems there is no chance we can reuse the existing AudioOutput

This doesn't apply if the AudioOutput doesn't use AudioTrack (such as using libusb and android USB host API for USB DAC), as then some concepts of AudioTrack (such as AudioAttributes) are completely irrelevant, so even if they change, AudioOutput can be reused as long as nothing more significant (sample rate, etc) changed.

That is irrelevant to the primary purpose of the PR (which is described below), but I had this in mind when creating this change as, if a new API is added to ExoPlayer, it makes sense to fulfill all reasonable use cases applicable.

> but could you describe in more detail what you want to achieve and which API calls you'd to make and then? For example, I'm not sure how'd apply effects to offload playback because they data is still encoded (and PCM offload not yet supported).

Platform AudioFX API supports applying effect to compressed offload playback, by describing the modifications that have to be done to the audio in a common structure, and DSP code provided by platform/device will apply it after decoding.

For example: https://developer.android.com/reference/android/media/audiofx/DynamicsProcessing
with the https://developer.android.com/reference/android/media/audiofx/DynamicsProcessing#setPostEqAllChannelsTo(android.media.audiofx.DynamicsProcessing.Eq) API

Not every device supports this, but those that do (such as modern Pixel devices) allow more audio usecases that require DSP to still use compressed offload.

Now the reason why audio output reuse need to be prevented is that this AudioFX API has no way to fence any changes to a specific audio timestamp, they will always be applied "as soon as possible, but best effort basis". 

ReplayGain adds a loudness change value that has to be applied by player to ID3/Vorbis comment metadata. But it is critical to only apply this to correct song. So if there is offload gapless transition between song A and B, but they have very different gain value, and accidentally gain value of B (let's say +10dB) would be applied to song A (normally could have -8dB for example) which is already very loud in the compressed data, we would play back the last few frames at +18dB above the expected volume and that could cause user's ears to bleed.

To sidestep the issue, we don't reuse audio output and hence prevent offload gapless transition (but ONLY if the gain value is different. for songs in same album, it would be the same, so gapless album playback can still function as intended). Instead, we wait until old audio output is done playing all data, then create new one. And during creating of new one, the customized audio output will first call AudioFX API to apply new gain, which will apply from the very first frame of the new audio output, and then create the audio output. This works well for the simple reason that frames that aren't yet written into AudioOutput cannot be buffered by AudioFX system.

andrewlewis:
I think the context for this was https://github.com/androidx/media/issues/2855.

tonihei:
Thanks for the additional details!

> This doesn't apply if the AudioOutput doesn't use AudioTrack (such as using libusb and android USB host API for USB DAC), as then some concepts of AudioTrack (such as AudioAttributes) are completely irrelevant, so even if they change, AudioOutput can be reused as long as nothing more significant (sample rate, etc) changed.

That's a good motivating example for why this makes sense in general.

> Platform AudioFX API supports applying effect to compressed offload playback

That makes sense too, I didn't consider effects applied at this level. 

I'll start the merge process for this PR.

nift4:
Thanks for the quick review, again!

Issue #3045, opened 2026-02-03
Some subtitle formats are not recognized


mobd:
### Version

Media3 main branch

### Devices that reproduce the issue

Xiaomi 14 running Android 16

### Reproducible in the demo app?

Yes

### Reproduction steps

1. Save the following text in an .srt file
```
1
0:00:00.00 --> 0:00:20.00
test
```
2. Play some media
3. Select the .srt file

### Expected result

Subtitles are displayed

### Actual result

No subtitles are shown.

### Media

Not applicable

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

mobd:
This format is used by some online players. It is handled correctly by VLC. The only difference between this and a normal .srt file is the formatting of timestamps:
`0:00:00.00` instead of `00:00:00,000`

According to [this comment ](https://github.com/recloudstream/cloudstream/issues/2476#issuecomment-3837939675) the issue is with the regex used:
```
\s*((?:(\d+):)?(\d+):(\d+)(?:,(\d{3}))?)\s*-->\s*((?:(\d+):)?(\d+):(\d+)(?:,(\d{3}))?)\s*
```
This will not match "0:00:00.00 --> 0:00:20.00" and gives "Skipping invalid timing".

icbaker:
Do you have a spec reference for the SubRip format that allows a `.` decimal separator and only 2 digits of fractional seconds?

On [wikipedia](https://en.wikipedia.org/wiki/SubRip#Format) it seems to be strictly defined as `,` and 3 digits:

> The timecode format used is hours:minutes:seconds,milliseconds with time units fixed to two zero-padded digits and fractions fixed to three zero-padded digits (00:00:00,000). The comma (,) is used for fractional separator.

This [Library of Congress page](https://loc.gov/preservation/digital/formats//fdd/fdd000569.shtml) looks similar:

> Timecode format: hours:minutes:seconds,milliseconds (00:00:00,000)

It seems to me that ExoPlayer is following the spec, and the media you are trying to play is not - and the solution is to fix the media to be compliant to the spec, not force every player in the ecosystem to support all the possible ways in which media producers ignore the spec. That's why specs exist - to make parsing easier, and avoid players having to try and guess what the content producer meant and increase their complexity to handle all the possibilities.

mobd:
I think you are right it's not according to spec. It's also not exactly like the WebVTT spec (that does use a full stop)

[...]

PR #3050, opened 2026-02-03
Fix: SUBRIP_TIMECODE #3045

fire-light42:
This fixes the subrip regex to allow more valid time encodings. According to the comment above the regex you already allow "non-spec" srt subtitles as: "Some SRT files don't include hours or milliseconds in the timecode, so we use optional groups."

So this pull request is only expands upon this reasoning by allowing "," and "." and more or less than 3 decimals because real world srt files use that sometimes. 

I do not have a google account, so I am unable to accept the CLA, but this pull request is free for anyone to use or create another pull request from. It is mainly to show how simple #3045 is to fix.

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/text/subrip/SubripParser.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/text/subrip/SubripParser.java
@@ -60,7 +60,7 @@ public final class SubripParser implements SubtitleParser {
   private static final String TAG = "SubripParser";
 
   // Some SRT files don't include hours or milliseconds in the timecode, so we use optional groups.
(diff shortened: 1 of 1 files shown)


libraries/extractor/src/main/java/androidx/media3/extractor/text/subrip/SubripParser.java
icbaker, line 289: Did you test this? I don't believe it will work correctly with a two digit fraction format described in #3045 - because here we assume the fractional part matched by the regex is in milliseconds (i.e. 3 fractional digits of a second). If there are only two digits, the value will be in centiseconds and so needs to be multiplied by 10 here. And since you amended the regex to `+` you also need to handle a single digit, and more than 3 digits.

icbaker:
Closing since it doesn't sound like the CLA will be signed.

PR #3055, opened 2026-02-04
Fix SurfaceView overflow in RESIZE_MODE_ZOOM

### LINKED ISSUE #1107: A few video in a list overlap each other
### Version

Media3 1.2.1

### More version details

We are designing a social media feed for our app and we are facing a problem with with our compose implementation 
in a feed we are using `androidx.media3.exoplayer.ExoPlayer` and via 
`AndroidView(
            modifier = Modifier.align(Alignment.BottomCenter),
            factory =PlayerView(content)
`
where I set a player per an item like this `this.player = player`
and to fit the screen size I use `AspectRatioFrameLayout.RESIZE_MODE_ZOOM`  <br>
it looks like a player takes all of the player Views to show a video

as a result, you can see the same video in several list items

it looks like the same issue was reported for old [exo-player](https://stackoverflow.com/questions/77766103/playerview-is-overlapping-to-each-other-in-exoplayer-media3-while-using-in-viewp), however without a working solution unfortunately 

Please help me to provide the best experience to our users

Pixel 5

### Reproduction steps

1. create a lazy column list 
2. add several videos in a row
3. start playing them one by one

### Expected result

The video should take a frame size and don't interact with a neighbour item

### Actual result

Current video plays in several list items

### Media

https://github.com/androidx/media/assets/2707599/62880d83-fee2-4e4c-b599-c2fcd60e1d47

### Bug Report

yemirhan:
## Summary

- Fixes `SurfaceView` overflow in `AspectRatioFrameLayout` when using `RESIZE_MODE_ZOOM`, which caused videos to overlap neighboring items in `RecyclerView`/`LazyColumn`
- Keeps the frame at parent-provided size while measuring children at zoomed size, centers oversized children, and clips overflow via canvas and RenderNode clipping

## Details

When using `RESIZE_MODE_ZOOM`, `AspectRatioFrameLayout` previously expanded its own measured dimensions beyond parent bounds. `SurfaceView`'s separate hardware surface extended past the frame, causing videos to visually overlap neighboring list items. `TextureView` was unaffected since it renders through the normal canvas pipeline.

### Changes to `AspectRatioFrameLayout.java`:

1. **`onMeasure()`** — After measuring children at zoomed size, resets the frame's measured dimensions back to the original parent-constrained size via `setMeasuredDimension()`
2. **`onLayout()`** — Centers children whose measured size exceeds the frame for a symmetric center-crop effect. `wrap_content` children (buffering spinner, error text) are left untouched
3. **`dispatchDraw()`** — Clips the canvas to the frame's bounds. Combined with `FrameLayout`'s default `clipChildren=true`, this also enables RenderNode-based clipping for `SurfaceView` on API 29+

Other resize modes (FIT, FILL, FIXED_WIDTH, FIXED_HEIGHT) are unaffected — all new logic is guarded by `resizeMode == RESIZE_MODE_ZOOM`.

## Test plan

- [x] Verified with two `PlayerView`s (both using `RESIZE_MODE_ZOOM` + `SurfaceView`) stacked vertically with a colored divider — videos no longer bleed past their bounds
- [ ] Verify other resize modes (FIT, FILL, FIXED_WIDTH, FIXED_HEIGHT) still work correctly
- [ ] Test on API < 29 devices for canvas-only clipping path

Fixes #1107

### DIFF
--- a/libraries/ui/src/main/java/androidx/media3/ui/AspectRatioFrameLayout.java
+++ b/libraries/ui/src/main/java/androidx/media3/ui/AspectRatioFrameLayout.java
@@ -19,7 +19,9 @@ import static java.lang.annotation.ElementType.TYPE_USE;
 
 import android.content.Context;
 import android.content.res.TypedArray;
+import android.graphics.Canvas;
 import android.util.AttributeSet;
+import android.view.View;
 import android.widget.FrameLayout;
 import androidx.annotation.IntDef;
 import androidx.annotation.Nullable;
@@ -176,6 +178,8 @@ public final class AspectRatioFrameLayout extends FrameLayout {
 
     int width = getMeasuredWidth();
     int height = getMeasuredHeight();
+    int originalWidth = width;
+    int originalHeight = height;
     float viewAspectRatio = (float) width / height;
     float aspectDeformation = videoAspectRatio / viewAspectRatio - 1;
     if (Math.abs(aspectDeformation) <= MAX_ASPECT_RATIO_DEFORMATION_FRACTION) {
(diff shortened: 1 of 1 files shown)


hohoaisan:
This PR got no attention at all :(

PR #3063, opened 2026-02-10
Refactor PCM format handling to use Util methods

nift4:
Tidy up PCM format handling by refactoring a few usages that manually handle all PCM formats to use the Util methods that already exist. 

This also adds missing string representations for existing formats (`C.ENCODING_PCM_DOUBLE` and `C.ENCODING_DSD`) to `EventLogger` to complete format coverage.

This is a pure refactoring change and does not alter existing playback behavior.

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java
@@ -2130,17 +2130,12 @@ public final class BoxParser {
       boolean isFloat = (formatSpecificFlags & 1) != 0;
       boolean isBigEndian = (formatSpecificFlags & (1 << 1)) != 0;
       if (!isFloat) {
-        if (bitsPerSample == 8) {
-          pcmEncoding = C.ENCODING_PCM_8BIT;
-        } else if (bitsPerSample == 16) {
-          pcmEncoding = isBigEndian ? C.ENCODING_PCM_16BIT_BIG_ENDIAN : C.ENCODING_PCM_16BIT;
-        } else if (bitsPerSample == 24) {
-          pcmEncoding = isBigEndian ? C.ENCODING_PCM_24BIT_BIG_ENDIAN : C.ENCODING_PCM_24BIT;
-        } else if (bitsPerSample == 32) {
-          pcmEncoding = isBigEndian ? C.ENCODING_PCM_32BIT_BIG_ENDIAN : C.ENCODING_PCM_32BIT;
-        }
-      } else if (bitsPerSample == 32) {
+        pcmEncoding = Util.getPcmEncoding(bitsPerSample, isBigEndian ? BIG_ENDIAN : LITTLE_ENDIAN);
+      } else if (!isBigEndian && bitsPerSample == 32) {
         pcmEncoding = C.ENCODING_PCM_FLOAT;
+      }
+      if (pcmEncoding == C.ENCODING_INVALID) {
+        pcmEncoding = Format.NO_VALUE;
       }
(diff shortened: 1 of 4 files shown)


rohitjoins:
Hi @nift4,

Thank you for raising this PR! On a closer look, it seems that this PR is actually doing quite a few different things at once. The current commit summary "Tidy up pcm format handling in some places" doesn't quite do justice to changes you've included here. To make this clean and easy to review, could I ask you to please split this up into smaller, separate PRs that handle one logical change at a time (not necessarily in this order)?

- A pure refactoring PR to switch to the Util methods.
- A feature PR adding the C.ENCODING_PCM_DOUBLE support, extraction logic, and the ToFloatPcmAudioProcessor format additions.
- A bug fix PR for the 20-bit ALAC issue.

This will make it much easier for us to safely review and merge each independent change. Thank you! Also, please feel free to assign these PRs directly to me.

nift4:
Hi @rohitjoins,

thanks for the quick response. It was done this way because the individual parts are quite small and conflict with each other, but I understand it's harder to review this way. I split it up into three parts as you said: The refactor is included in this PR, the bug fix for 20-bit files is #3089 and the feature one is #3090.

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java
```diff
@@ -2130,18 +2130,13 @@ private static void parseAudioSampleEntry(
...
-        } else if (bitsPerSample == 16) {
-          pcmEncoding = isBigEndian ? C.ENCODING_PCM_16BIT_BIG_ENDIAN : C.ENCODING_PCM_16BIT;
-        } else if (bitsPerSample == 24) {
-          pcmEncoding = isBigEndian ? C.ENCODING_PCM_24BIT_BIG_ENDIAN : C.ENCODING_PCM_24BIT;
-        } else if (bitsPerSample == 32) {
-          pcmEncoding = isBigEndian ? C.ENCODING_PCM_32BIT_BIG_ENDIAN : C.ENCODING_PCM_32BIT;
-        }
-      } else if (bitsPerSample == 32) {
+        pcmEncoding = Util.getPcmEncoding(bitsPerSample, isBigEndian ? BIG_ENDIAN : LITTLE_ENDIAN);
+      } else if (!isBigEndian && bitsPerSample == 32) {
```
rohitjoins: Addition of `!isBigEndian` looks more like a change in logic than pure refactor, was it intentional?

nift4: Oh, my bad I forgot to split that. It's a bug fix because ENCODING_PCM_FLOAT refers to little endian only and big endian would cause static noise playback. Should we merge this here or in another PR?

rohitjoins: Thanks for clarifying! Since it changes behavior, it would ideally be in its own PR. But if creating another one is a pain, no worries at all, I can take care of splitting this change internally on our end. Let me know!

nift4: Because I'll be away from my computer for a few days, it'd probably be best if you split this change internally instead. Thank you!

rohitjoins: The bug fix was merged as [cd19799](https://github.com/androidx/media/commit/cd197994472ff6c1f2443a51e4fc5a25a3df09cc).

nift4: Great, thanks

PR #3064, opened 2026-02-11
Add AdaptiveTrackSelection format priority ordering hooks

AndyWangLYN:
### Summary
This change adds extension points to AdaptiveTrackSelection so apps can control candidate format priority beyond default bitrate ordering, while preserving existing ABR behavior by default.

### What Changed

Added [getFormatPriorityOrder(...)](app://-/index.html#) hook to let developers provide an explicit per-update format priority list.
Added [getFormatIndexForPriorityPosition(...)](app://-/index.html#) hook (default bitrate-descending behavior) for position-based custom ordering.
Updated ideal-index selection to use custom priority when provided.
Kept default behavior unchanged when hooks are not overridden.
Added validation/fallback behavior for invalid custom priority order:
invalid order is ignored with a warning
selection falls back to default ordering
Added unit coverage in AdaptiveTrackSelectionTest proving a custom priority order can be applied.
### Why
Current adaptive selection effectively prioritizes higher bitrate first. This makes it hard to prioritize developer-defined rules (for example, quality score ranking). These hooks allow custom ordering without changing core ABR switching logic.

### Compatibility

No behavior change for existing users.
No API break for default usage.
Buffer-based switch-up/switch-down logic remains unchanged.
### Testing

Ran:
[gradlew :lib-exoplayer:testDebugUnitTest --tests androidx.media3.exoplayer.trackselection.AdaptiveTrackSelectionTest](app://-/index.html#)
Result: BUILD SUCCESSFUL

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
@@ -565,6 +565,40 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
   }
 
   /**
+   * Returns the preferred format evaluation order.
+   *
+   * @return An array with one format index per priority position, or null to use {@link
+   *     #getFormatIndexForPriorityPosition(int, long, long, long)}. If non-null, indices must be
+   *     unique and in range {@code [0, length())}.
+   *
+   * @param nowMs The current time in the timebase of {@link Clock#elapsedRealtime()}, or {@link
+   *     Long#MIN_VALUE} to ignore track exclusion.
+   * @param chunkDurationUs The duration of a media chunk in microseconds, or {@link C#TIME_UNSET}
+   *     if unknown.
+   * @param effectiveBitrate The bitrate available to this selection.
+   */
+  @Nullable
+  protected int[] getFormatPriorityOrder(long nowMs, long chunkDurationUs, long effectiveBitrate) {
+    return null;
+  }
+
+  /**
+   * Returns the format index to evaluate at a given priority position. The default implementation
+   * evaluates formats in order of decreasing bitrate.
+   *
+   * @param priorityPosition The zero-based position within the format evaluation priority.
+   * @param nowMs The current time in the timebase of {@link Clock#elapsedRealtime()}, or {@link
+   *     Long#MIN_VALUE} to ignore track exclusion.
+   * @param chunkDurationUs The duration of a media chunk in microseconds, or {@link C#TIME_UNSET}
+   *     if unknown.
+   * @param effectiveBitrate The bitrate available to this selection.
+   */
+  protected int getFormatIndexForPriorityPosition(
+      int priorityPosition, long nowMs, long chunkDurationUs, long effectiveBitrate) {
+    return priorityPosition;
+  }
+
+  /**
    * Called from {@link #evaluateQueueSize(long, List)} to determine whether an evaluation should be
    * performed.
    *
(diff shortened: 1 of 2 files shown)


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
```diff
@@ -599,19 +633,50 @@ protected long getMinDurationToRetainAfterDiscardUs() {
...
+        formatIndex = i;
+      }
+      if (nowMs == Long.MIN_VALUE || !isTrackExcluded(formatIndex, nowMs)) {
+        Format format = getFormat(formatIndex);
         if (canSelectFormat(format, format.bitrate, effectiveBitrate)) {
-          return i;
+          return formatIndex;
         } else {
-          lowestBitrateAllowedIndex = i;
+          lowestBitrateAllowedIndex = formatIndex;
```
oceanjules: If `lowestBitrateAllowedIndex` is your fallback, then it looks like in the new implementation, this variable will be updated at every iteration of the priority loop and hence will return the last index processed in the priority order (assuming all tracks are filtered out by bitrate constraints). I don't know if it's always the safe fallback track?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
```diff
@@ -599,19 +633,50 @@ protected long getMinDurationToRetainAfterDiscardUs() {
...
       }
     }
     return lowestBitrateAllowedIndex;
   }
 
+  private boolean isValidFormatPriorityOrder(int[] formatPriorityOrder) {
+    if (formatPriorityOrder.length != length) {
+      return false;
+    }
+    boolean[] seen = new boolean[length];
```
oceanjules, line 680: `determineIdealSelectedIndex` calls `isValidFormatPriorityOrder` which allocates the array on every call.  If there are fewer than 64 tracks, maybe you can use a `long` bitmask. If you can see the performance impact through the profiler, maybe maintaining a reusable buffer is better, but maybe it's overkill.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
```diff
@@ -599,19 +633,50 @@ protected long getMinDurationToRetainAfterDiscardUs() {
...
+      formatPriorityOrder = null;
+    }
     for (int i = 0; i < length; i++) {
-      if (nowMs == Long.MIN_VALUE || !isTrackExcluded(i, nowMs)) {
-        Format format = getFormat(i);
+      int formatIndex =
+          formatPriorityOrder == null
+              ? getFormatIndexForPriorityPosition(i, nowMs, chunkDurationUs, effectiveBitrate)
+              : formatPriorityOrder[i];
+      if (formatIndex < 0 || formatIndex >= length) {
```
oceanjules: I don't think this check is needed at this point. Either
* `formatIndex == getFormatIndexForPriorityPosition(i, ...) == i`
* `formatPriorityOrder[i]` was just earlier check to be `isValidFormatPriorityOrder`

tonihei:
> This makes it hard to prioritize developer-defined rules (for example, quality score ranking). 

Do you have an concrete example where this logic applies? If there is a format with a higher bitrate and a lower quality, I don't see a case where it would ever be beneficial to select that format. If that's the case, a better approach to solve your underlying issue may be to ensure that format is never part of the ABR selection ladder to start with.

Example:
 - bitrate: 100, quality: 50
 - bitrate: 200, quality: 70
 - bitrate: 250, quality: 65 -> this Format is strictly worse than the one above in all scenarios and should never be considered
 - bitrate: 300, quality: 80

> These hooks allow custom ordering without changing core ABR switching logic

The concept of bitrate ordering is quite deeply engrained into the `AdaptiveTrackSelection` class and even the interface `TrackSelection` that requires `Format` instances to be sorted by bitrate. So besides the method you updates already, there are more places that would need to be touched.

AndyWangLYN:
Hey @tonihei thank you for reviewing and commenting on my PR. This is an ABR customization issue I've encountered where server-side pruning isn't feasible because the 'best' track is device-dependent. For example, in a mixed-codec manifest, a more efficient codec can provide higher quality at a lower bitrate. Furthermore, on devices without hardware AV1 decoding, a high-bitrate AV1 rendition may cause stuttering compared to a lower-bitrate AVC rendition.
Regarding your point on bitrate ordering being engrained in the interface: I've tried to keep the hooks as non-intrusive as possible, aimming to give developers options for track priority only during ABR. But I'm open to suggestions on how to better align this with the existing architecture. Thank you!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
AndyWangLYN: Thank you for the comment! I updated to compute upfront from bitrate order as the lowest-bitrate non-excluded index, independent of priority traversal.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
AndyWangLYN: Added fast path with long bitmask when length <= 64

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
AndyWangLYN: Removed the extra in-loop formatIndex < 0 || formatIndex >= length check after validated formatPriorityOrder

AndyWangLYN:
Based on the comment above, opened a new PR: https://github.com/androidx/media/pull/3068 to better address the track format sorting from the base level. Intend to close this PR once that one is deemed good.

AndyWangLYN:
Let's move the conversation to this PR: https://github.com/androidx/media/pull/3068 in which I propose a better and more robost way to sort format tracks. Thank you! @tonihei @oceanjules

PR #3068, opened 2026-02-12
Add custom format comparator support to AdaptiveTrackSelection

AndyWangLYN:
### Summary
 Adds optional custom format ordering for `AdaptiveTrackSelection` so apps can influence
  ABR selection priority beyond bitrate-only ordering. **While default behavior remain unchanged**

**This PR is refined version of the PR: https://github.com/androidx/media/pull/3064** in which I received comments about track format sorting. Intend to close above PR if this one is deemed good eventually.

### What changed
- Added `AdaptiveTrackSelection.Factory#setTrackFormatComparator(Comparator<Format>)`
    to allow custom format ordering via the factory.
  - Added `protected final getTrackFormatComparator()` on `Factory` so subclasses that
    override `createAdaptiveTrackSelection()` can access the configured comparator.
  - Added a new `protected` constructor on `AdaptiveTrackSelection` accepting a
    `Comparator<Format>`, enabling subclasses to pass a comparator directly.
  - Updated switch-up/switch-down direction logic in `updateSelectedTrack` to use index
    comparison instead of bitrate comparison, so it works correctly with any comparator
    (lower index = higher priority).
  - Extracted `DEFAULT_FORMAT_COMPARATOR` as a named constant in `BaseTrackSelection`.
  - Updated `TrackSelection` and field Javadoc to say "selection order" instead of
    "decreasing bandwidth order".

### Why
The current implementation always prioritizes higher bitrate. This makes it difficult to
  express developer-defined rules (e.g. a quality score ranking) without subclassing and
  duplicating ABR logic. These hooks allow custom ordering without touching core ABR
  switching behavior.

### Testing Done
- Unit test for comparator and default behavior

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
@@ -15,6 +15,7 @@
  */
 package androidx.media3.exoplayer.trackselection;
 
+import static com.google.common.base.Preconditions.checkNotNull;
 import static java.lang.Math.max;
 import static java.lang.Math.min;
 
@@ -39,11 +40,12 @@ import com.google.common.collect.Multimap;
 import com.google.common.collect.MultimapBuilder;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.List;
 
 /**
  * A bandwidth based adaptive {@link ExoTrackSelection}, whose selected track is updated to be the
- * one of highest quality given the current network conditions and the state of the buffer.
+ * highest priority track given the current network conditions and the state of the buffer.
  */
 @UnstableApi
 public class AdaptiveTrackSelection extends BaseTrackSelection {
@@ -61,6 +63,7 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
     private final float bandwidthFraction;
     private final float bufferedFractionToLiveEdgeForQualityIncrease;
     private final Clock clock;
+    private Comparator<Format> trackFormatComparator;
 
     /** Creates an adaptive track selection factory with default parameters. */
     public Factory() {
@@ -227,6 +230,19 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
       this.bufferedFractionToLiveEdgeForQualityIncrease =
           bufferedFractionToLiveEdgeForQualityIncrease;
       this.clock = clock;
+      this.trackFormatComparator = BaseTrackSelection.DEFAULT_FORMAT_COMPARATOR;
+    }
+
+    /**
+     * Sets the comparator used to order formats in adaptive track selections.
+     * The comparator order controls which formats are considered first during adaptation.
+     *
+     * @param trackFormatComparator Comparator used to order selected formats.
+     * @return This factory, for convenience.
+     */
+    public Factory setTrackFormatComparator(Comparator<Format> trackFormatComparator) {
+      this.trackFormatComparator = checkNotNull(trackFormatComparator);
+      return this;
     }
 
     @Override
@@ -289,7 +305,8 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
           bandwidthFraction,
           bufferedFractionToLiveEdgeForQualityIncrease,
           adaptationCheckpoints,
-          clock);
+          clock,
+          trackFormatComparator);
     }
   }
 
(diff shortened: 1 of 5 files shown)


AndyWangLYN:
hey guys can I get some feedback on this PR? Any feedback can do. Very much appreciated! @tonihei @oceanjules

AndyWangLYN:
Also raised a question regarding this: https://github.com/androidx/media/issues/3077

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
```diff
@@ -458,23 +540,24 @@ public void updateSelectedTrack(
       previousSelectedIndex = formatIndexOfPreviousChunk;
       previousReason = Iterables.getLast(queue).trackSelectionReason;
     }
-    int newSelectedIndex = determineIdealSelectedIndex(nowMs, chunkDurationUs);
+    int newSelectedIndex = determineIdealSelectedIndexForEffectiveBitrate(nowMs, effectiveBitrate);
```
tonihei: This method is still determining the ideal selected index by going through the ordered list of options and returning the first one where `canSelectFormat` returns true. Doesn't that mean we can keep the existing name of the method?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
```diff
@@ -458,23 +540,24 @@ public void updateSelectedTrack(
...
-      if (selectedFormat.bitrate > currentFormat.bitrate
+      if (newSelectedIndex < previousSelectedIndex
           && bufferedDurationUs < minDurationForQualityIncreaseUs) {
-        // The selected track is a higher quality, but we have insufficient buffer to safely switch
+        // The selected track is higher priority, but we have insufficient buffer to safely switch
         // up. Defer switching up for now.
         newSelectedIndex = previousSelectedIndex;
-      } else if (selectedFormat.bitrate < currentFormat.bitrate
+      } else if (newSelectedIndex > previousSelectedIndex
+          && (isUsingDefaultFormatComparator()
```
tonihei: Can you add more details why you'd like to add this additional constraint? It seems the generic logic would still apply `maxDurationForQualityDecreaseUs` without further conditions.

Related to that - I think we should't have a method `isUsingDefaultFormatComparator`. If there is a different behavior needed, it would need to be in a method that describes the actual behavior and can be customized by everyone (not just the default logic).

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
```diff
@@ -597,19 +680,33 @@ protected long getMinDurationToRetainAfterDiscardUs() {
...
+    int lowestBitrateAllowedIndex = C.INDEX_UNSET;
+    int lowestBitrate = Integer.MAX_VALUE;
     for (int i = 0; i < length; i++) {
       if (nowMs == Long.MIN_VALUE || !isTrackExcluded(i, nowMs)) {
         Format format = getFormat(i);
         if (canSelectFormat(format, format.bitrate, effectiveBitrate)) {
           return i;
-        } else {
+        }
+        int formatBitrate = format.bitrate == Format.NO_VALUE ? 0 : format.bitrate;
```
tonihei: Similar to my other comment above, I'd assume this logic can be kept as it was before? If no format returns true for `canSelectFormat`, we just fall back to the lowest quality format that isn't excluded. The only thing that changes is the definition of 'quality'.

libraries/exoplayer/src/test/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelectionTest.java
```diff
@@ -215,6 +216,104 @@ public void updateSelectedTrack_liveStream_switchesUpWhenBufferedFractionToLiveE
...
+    AdaptiveTrackSelection adaptiveTrackSelection =
+        prepareAdaptiveTrackSelectionWithFormatComparator(
+            trackGroup, formatComparatorInOrder(ImmutableList.of(format2, format1, format3)));
+
+    assertThat(adaptiveTrackSelection.getSelectedFormat()).isEqualTo(format1);
+    assertThat(adaptiveTrackSelection.getSelectionReason()).isEqualTo(C.SELECTION_REASON_INITIAL);
+  }
+
+  @Test
+  public void initial_updateSelectedTrack_withInvalidComparator_fallsBackToDefaultOrdering() {
```
tonihei: There is no reason to account for invalid Comparators. If custom app code provides an implementation of a well-defined interface, we should be able to assume they implemented the interface correctly.

libraries/exoplayer/src/test/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelectionTest.java
```diff
@@ -215,6 +216,104 @@ public void updateSelectedTrack_liveStream_switchesUpWhenBufferedFractionToLiveE
     assertThat(adaptiveTrackSelection.getSelectedFormat()).isEqualTo(format3);
   }
 
+  @Test
+  public void initial_updateSelectedTrack_usesFactoryProvidedTrackFormatComparatorOrder() {
```
tonihei, line 223: Thanks for providing tests for your change!

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
AndyWangLYN: Make sense, determineIdealSelectedIndex has been kept as-is now

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
AndyWangLYN: You're right on both points.

The deferral logic applies uniformly regardless of the comparator, so no special-casing is needed. I’ve removed isUsingDefaultFormatComparator entirely.

I also updated the switch-direction logic: With a custom comparator, the higher-priority track may not have a higher bitrate, so bitrate comparison isn’t reliable. Using index comparison makes the logic correct for any comparator while preserving the same semantics in the default case.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelection.java
AndyWangLYN: Agreed, reverted to the original pattern

libraries/exoplayer/src/test/java/androidx/media3/exoplayer/trackselection/AdaptiveTrackSelectionTest.java
AndyWangLYN: For sure! Thank you for recognizing!

AndyWangLYN:
Hey ready for another review whenever you have the time. Thank you @tonihei

AndyWangLYN:
Mind sharing some more insights on this PR? I really believe this could be helpful, especially in mixed codec (bitrate != video quality) cases and could be beneficial (ran into this issue during my work). Thank you! @tonihei

tonihei:
The updated version looks good, I'll start the import process. You may see some updates appearing here, please refrain from uploading a new version to avoid merge conflicts. Thanks!

tonihei:
Oh, actually. I can't merge as it is because it targets the `release` branch`. Can you rebase onto the `main` branch?

AndyWangLYN:
@tonihei Ah I see. Just rebased onto main, thank you!

Issue #1646, opened 2024-08-22
Ktor support


JJSarrasin:
The standard for HTTP clients is increasingly shifting towards Ktor. It avoids to manage multiple http engine in the same project

My main concern is about DrmSessionManagerProvider.setDrmHttpDataSourceFactory + MediaSourceFactory.setDataSourceFactory, maybe there are other entry points.

icbaker:
> My main concern is about DrmSessionManagerProvider.setDrmHttpDataSourceFactory + MediaSourceFactory.setDataSourceFactory, maybe there are other entry points.

Can you please add a bit more concrete info on how Ktor would resolve this? At the moment these are configured with different `DataSource.Factory` instances, and each default to `DefaultHttpDataSource.Factory` for handling HTTP URIs.

What are you suggesting would change here?

Note that we probably wouldn't accept a dependency on Ktor from the main `exoplayer` module, so any integration would need to be in an extension (similar to the `datasource-okhttp` extension).

JJSarrasin:
From my understanding, there is a dependency between media3 and okhttp.

Here is the minimal use we do from exo:
```
val renderersFactory = DefaultRenderersFactory(context)
val dataSourceFactory = OkHttpDataSource.Factory(httpClient)

val drmSessionManagerProvider = DefaultDrmSessionManagerProvider()
drmSessionManagerProvider.setDrmHttpDataSourceFactory(dataSourceFactory)

val mediaSourceFactory = DefaultMediaSourceFactory(context)
    .setLoadErrorHandlingPolicy(DefaultLoadErrorHandlingPolicy(2))
    .setDrmSessionManagerProvider(drmSessionManagerProvider)

mediaSourceFactory.setDataSourceFactory(dataSourceFactory)

val exoPlayer = ExoPlayer.Builder(context, renderersFactory)
    .setMediaSourceFactory(mediaSourceFactory)
    .build()
```

The passed httpClient is a OkHttpclient variable.

Of course, we won't use media3 on the iOS/web part of our project, so we'll only use the okhttp client implementation of ktor wrapper.
We're using Ktor for all the common part of the code and it's kind of the main lib for doing network request in modern KMP projects.

Maybe we should ask Ktor lib to expose a way to get the OkHttpEngine/OkHttpClient instead ?

icbaker:
> Maybe we should ask Ktor lib to expose a way to get the OkHttpEngine/OkHttpClient instead ?

This seems like a a reasonable request to me.

----

[...]

PR #3071, opened 2026-02-14
Add Ktor HTTP data source module

k88936:
# Summary
Introduce a new lib-datasource-ktor module similar to `libraries/datasource_okhttp`, providing KtorDataSource as an alternative HTTP data source implementation using Ktor client to solve issue #1646 .

# What changed
* impl:  androidx.media3.datasource.DataSource using ktor
* test: a unit test for DataSourceContractTest. and a unit test for it own.
* api.txt in the repo root
* some gradle config for include this module

# Why 
I recent met the same problem for maintaining both a okhttp client and a ktor client, similar to https://github.com/androidx/media/issues/1646#issue-2480891366


libraries/datasource_ktor/proguard-rules.txt
```diff
@@ -0,0 +1,11 @@
...
+
+# Options for Ktor and Okio
+-dontwarn io.ktor.**
+-dontwarn okio.**
+-dontwarn javax.annotation.**
+-dontwarn org.conscrypt.**
+
+# Keep Ktor client classes
+-keep class io.ktor.** { *; }
+-keep class kotlinx.coroutines.** { *; }
```
icbaker: Please can you give some details on why these rules are needed? They look extremely broad and seem like they will result in keeping a lot of classes that may not be used or needed, bloating APK size.

If the ktor library doesn't work if these classes are not r8-kept then this config should probably be part of the ktor library config, rather than encoded here (and realistically, probably a bug filed against ktor). This config (in media3) is not the right place to encode this specific knowledge about ktor.

libraries/datasource_ktor/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+    api project(modulePrefix + 'lib-datasource')
+    implementation 'androidx.annotation:annotation:' + androidxAnnotationVersion
+    compileOnly 'com.google.errorprone:error_prone_annotations:' + errorProneVersion
+    compileOnly 'org.checkerframework:checker-qual:' + checkerframeworkVersion
+    compileOnly 'org.jetbrains.kotlin:kotlin-annotations-jvm:' + kotlinAnnotationsVersion
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:' + kotlinxCoroutinesVersion
+    testImplementation project(modulePrefix + 'test-utils')
+    testImplementation 'com.squareup.okhttp3:mockwebserver:' + okhttpVersion
+    testImplementation 'org.robolectric:robolectric:' + robolectricVersion
+    testImplementation 'io.ktor:ktor-client-okhttp:' + ktorVersion
```
icbaker: Is there a reason to use `okhttp` here instead of `ktor-client-android`?

libraries/datasource_ktor/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Ktor is a multiplatform HTTP client developed by JetBrains. It supports HTTP/2,
+WebSocket, and Kotlin coroutines for asynchronous operations.
+
+[HttpDataSource]: ../datasource/src/main/java/androidx/media3/datasource/HttpDataSource.java
+[Ktor]: https://ktor.io/
+
+## Getting the module
+
+The easiest way to get the module is to add it as a gradle dependency:
```
icbaker, line 13: Should we clarify here that you **also** need to depend on a ktor impl e.g. `io.ktor:ktor-client-android`?

Without that, my experiment in the compose demo app crashed the whole app.

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
```diff
@@ -0,0 +1,519 @@
...
+
+    @UnstableApi
+    override fun getResponseCode(): Int {
+        return response?.status?.value ?: -1
+    }
+
+    @UnstableApi
+    override fun getResponseHeaders(): Map<String, List<String>> {
+        val httpResponse = response ?: return emptyMap()
+        val headers = TreeMap<String, List<String>>(String.CASE_INSENSITIVE_ORDER)
```
icbaker: Why is the ordered map here important?

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
```diff
@@ -0,0 +1,519 @@
...
+        httpResponse.headers.names().forEach { name ->
+            headers[name] = httpResponse.headers.getAll(name) ?: emptyList()
+        }
+        return headers
+    }
+
+    @UnstableApi
+    override fun setRequestProperty(name: String, value: String) {
+        requireNotNull(name) { "name cannot be null" }
+        requireNotNull(value) { "value cannot be null" }
```
icbaker: I think we don't need this because the (non-)nullness is enforced at the language-level by kotlin?

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
```diff
@@ -0,0 +1,519 @@
...
+                    setBody(dataSpec.httpBody!!)
+                } else if (dataSpec.httpMethod == DataSpec.HTTP_METHOD_POST) {
+                    setBody(ByteArray(0))
+                }
+            }.execute()
+        }
+    }
+
+    @Throws(IOException::class)
+    private fun <T> executeSuspend(block: suspend () -> T): T {
```
icbaker: It surprises me that you need to write this much boilerplate

I would expect this whole class to be more coroutine-first, with any blocking needed done at the 'top level' (e.g. inside `open()`) - and everything else done with coroutines and async-await.

icbaker (changes requested):
I've pushed a few commits, and added a few comments too

libraries/datasource_ktor/proguard-rules.txt
k88936: oh yes. that's right. it is a [ issue for ktor on android ](https://github.com/ktorio/ktor/issues/379), the buggy matter should be handled by user of ktor ( since he uses it everywhere in his app )

libraries/datasource_ktor/README.md
k88936: it was my thought that since datasource accepts a httpclient as param , user must have created it somewhere else in his app, that is to say he will have his ktor engine impl ( maybe not io.ktor:ktor-client-android ) in his dependency

libraries/datasource_ktor/build.gradle
k88936: no special reason , should we also test on all different impls ?

libraries/datasource_ktor/proguard-rules.txt
icbaker: I removed all these proguard rules and re-ran my integration with the compose demo app (release build), and didn't see any problems.

Given that the issue you linked was marked obsolete almost 5 years ago, I'm inclined to remove all these rules until we have a concrete understanding of why they are needed.

libraries/datasource_ktor/README.md
icbaker: I naively wrote this code to use it from the compose demo app, and **only** added the dependency on `project(modulePrefix + 'lib-datasource-ktor')`:

```kt
ExoPlayer.Builder(context)
  .setMediaSourceFactory(
    DefaultMediaSourceFactory(context)
      .setDataSourceFactory(KtorDataSource.Factory(HttpClient())))
  .build().apply {
    setMediaItems(mediaItems)
    prepare()
  }
```

It compiled, but crashed at runtime with:

> Caused by: java.lang.IllegalStateException: Failed to find HTTP client engine implementation: consider adding client engine dependency. See https://ktor.io/docs/http-client-engines.html

libraries/datasource_ktor/build.gradle
icbaker: Interesting, did you see that failure before or after i moved the tests to be instrumentation instead of robolectric: https://github.com/androidx/media/pull/3071/changes/ef560497ed30ee3286331c43e62ab0a5ec946fe9

I tried it, and all the tests passed for me (on an emulator)

icbaker: > should we also test on all different impls ?

I don't think that's necessary, picking one seems fine, i just (naively!) assumed `-android` would be the 'default' choice.

k88936: oh I see it:  -android passed all.

> I tried it, and all the tests passed for me (on an emulator)

libraries/datasource_ktor/README.md
k88936: ... for ktor,  both dep: -core and a -impl are required ( explicit in their docs )  , as for HttpClient(WITHOUT A PARAM HERE)  ktor will try to find a suitable impl at runtime, that makes the crush.

k88936: but ,,, looked the bad test case on impls other than -android, you have a point , to provide default impl

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
k88936: I noticed that ktor's header are case-insensitive already, but not support case-insensitive look up. OR test for header part will fail

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
```diff
@@ -0,0 +1,496 @@
...
+    /**
+     * Sets the Cache-Control header that will be used.
+     *
+     * The default is `null`.
+     *
+     * @param cacheControl The cache control header value that will be used, or `null` to clear a
+     *   previously set value.
+     * @return This factory.
+     */
+    fun setCacheControl(cacheControl: String?): Factory {
```
icbaker: Do you have an explicit need for this, or did you just add it because it's on `OkHttpDataSource.Factory`? It doesn't seem to be present on the other 3 current `HttpDataSource` impls (Default, Cronet & HttpEngine) so I'm tempted to drop it from here for now unless there's a strong argument to keep it (we can always add it later if people request it).

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
```diff
@@ -186,27 +187,18 @@ private constructor(
...
     return response?.status?.value ?: -1
   }
 
   override fun getResponseHeaders(): Map<String, List<String>> {
-    val httpResponse = response ?: return emptyMap()
-    // ordered map use case-insensitive comparator to support case-insensitive lookup
-    val result = TreeMap<String, List<String>>(String.CASE_INSENSITIVE_ORDER)
-    for (name in httpResponse.headers.names()) {
-      result[name] = httpResponse.headers.getAll(name) ?: emptyList()
+    val headers = response?.headers ?: return emptyMap()
```
k88936, line 84: Oh yes , instead of creating a new map,  wrap the existing one with a view ensures case-insensitive !!

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
```diff
@@ -0,0 +1,484 @@
...
+                  DataSpec.HTTP_METHOD_GET -> HttpMethod.Get
+                  DataSpec.HTTP_METHOD_POST -> HttpMethod.Post
+                  DataSpec.HTTP_METHOD_HEAD -> HttpMethod.Head
+                  else -> HttpMethod.Get
+                }
+
+              if (dataSpec.httpBody != null) {
+                setBody(dataSpec.httpBody!!)
+              } else if (dataSpec.httpMethod == DataSpec.HTTP_METHOD_POST) {
+                setBody(ByteArray(0))
```
icbaker: Is this needed? 

I realised we didn't have any contract tests around `POST` requests so I added some (https://github.com/androidx/media/commit/5eb12590b50f06e8b665a7a64a4bdf98236cf421) and rebased this PR on top - but they all pass if I remove this line.

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
icbaker: yeah, seems to pass the contract tests

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
k88936: I just added that because it is on OkHttpDataSource.Factory, no special reson

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
k88936: some http clients/servers may treat no body or zero len body differently, , , , , I leave this for compatibility

icbaker: Thanks for the context. I used charles proxy to analyse the behaviour with and without the `setBody(ByteArray(0))` logic.

In both cases charles shows that the body is 'present' with length zero (indicated by `Content-Length: 0` header).

The only difference is whether `Content-Type` is set. With this line, it is set to `application/octet-stream` and without it the header is not present.

Therefore I think this isn't needed (since `Content-Type` is not a required header) - so I plan to remove it.

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
icbaker: OK thanks, I'll remove it for now.

libraries/datasource_ktor/src/androidTest/java/androidx/media3/datasource/ktor/KtorDataSourceTest.kt
```diff
@@ -0,0 +1,161 @@
...
+    assertThat(headers["1"]).isEqualTo(propertyFromSetter)
+    assertThat(headers["2"]).isEqualTo(propertyFromDataSpec)
+    assertThat(headers["3"]).isEqualTo(propertyFromDataSpec)
+    assertThat(headers["4"]).isEqualTo(propertyFromDataSpec)
+    assertThat(headers["5"]).isEqualTo(propertyFromSetter)
+    assertThat(headers["6"]).isEqualTo(propertyFromDataSpec)
+  }
+
+  @Test
+  fun open_invalidResponseCode() {
```
icbaker: This test is flaky. I ran it 100 times, and it failed 12 times. When it fails, it's because the `exception.responseBody` below is empty.

I added a bit of logging, and I see that it's because `read=0` in this part of the impl when generating `errorResponseBody`:

```
val read = ch.readAvailable(buffer)
```

Please can you take a look and fix the flakiness?

I've rebased this on top of `main` and force-pushed, so you should probably start by pulling the latest version of the code from GitHub to avoid accidentally diverging.

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
```diff
@@ -0,0 +1,464 @@
...
+     * Sets the user agent that will be used.
+     *
+     * The default is `null`, which causes the default user agent of the underlying [HttpClient] to
+     * be used.
+     *
+     * @param userAgent The user agent that will be used, or `null` to use the default user agent of
+     *   the underlying [HttpClient].
+     * @return This factory.
+     */
+    @CanIgnoreReturnValue
```
icbaker: Please can you add some tests that cover this setter and the ones below.

libraries/datasource_ktor/src/main/java/androidx/media3/datasource/ktor/KtorDataSource.kt
```diff
@@ -43,13 +43,14 @@ import io.ktor.http.HttpMethod
 import io.ktor.http.contentLength
 import io.ktor.http.contentType
 import io.ktor.utils.io.ByteReadChannel
-import io.ktor.utils.io.availableForRead
 import io.ktor.utils.io.readAvailable
+import io.ktor.utils.io.readRemaining
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.runBlocking
+import kotlinx.io.readByteArray
```
icbaker: I'm afraid this `kotlinx.io` dependency is not available when we build the library internally (partly due to still being in alpha). Please can you re-work the flakiness fix to not use it directly?

icbaker:

PR #3085, opened 2026-02-22
Skip HDR10+ metadata processing when Dolby Vision Profile 8 is detected

### LINKED ISSUE #1895: Firestick 4k max (files with hdr10+ and dolby vision profile 8.1)
### Version

ExoPlayer 2.18.4

### More version details

_No response_

**Bug Description**
There is an old bug that won't let you watch mkv video files that have dolby vision/hdr10.

Kodi developers was able to create a workaround for their app on Android TV (by disallowing either HDR10+ or DV). Additionally these devices seem to struggle with Profile 7, forcing 8.1 as a hack seems to work as well.

**Devices:**
- TCL C825
- TCL MQLED80
- Firestick 4k max
Other Android TVs may experience the same issue, but haven't had the chance to test as I only have access to these two TV models.

**More information:**
This happend on all mkv files that have dolby vision 8.1/hdr10+.

Unable to play this media with firestick:
General
Unique ID : 287260790794359083915345060219622240757 (0xD81C6A775E611494CE0759D557FFB1F5)
Complete name : G:\Downloads\Filmes\Cocaine Bear\Cocaine Bear - Get in line..mkv
Format : Matroska
Format version : Version 4
File size : 16.7 GiB
Duration : 1 h 35 min
Overall bit rate : 25.0 Mb/s
Frame rate : 23.976 FPS
Encoded date : 2023-03-14 13:49:23 UTC
Writing application : mkvmerge v73.0.0 ('25 or 6 to 4') 64-bit
Writing library : libebml v1.4.4 + libmatroska v1.7.1

Video
ID : 1
Format : HEVC
Format/Info : High Efficiency Video Coding
Format profile : Main 10@L5@High
HDR format : Dolby Vision, Version 1.0, dvhe.08.06, BL+RPU, HDR10 compatible / SMPTE ST 2094 App 4, Version 1, HDR10+ Profile B compatible
Codec ID : V_MPEGH/ISO/HEVC
Duration : 1 h 35 min
Bit rate : 24.3 Mb/s
Width : 3 836 pixels
Height : 1 602 pixels
Display aspect ratio : 2.40:1
Frame rate mode : Constant
Frame rate : 23.976 FPS
Color space : YUV
Chroma subsampling : 4:2:0 (Type 2)
Bit depth : 10 bits
Bits/(Pixel*Frame) : 0.165
Stream size : 16.2 GiB (97%)
Language : English (US)
Default : Yes
Forced : Yes
Color range : Limited
Color primaries : BT.2020
Transfer characteristics : PQ
Matrix coefficients : BT.2020 non-constant
Mastering display color primaries : Display P3
Mastering display luminance : min: 0.0050 cd/m2, max: 4000 cd/m2
Maximum Content Light Level : 461 cd/m2
Maximum Frame-Average Light Level : 305 cd/m2

Audio
ID : 2
Format : E-AC-3 JOC
Format/Info : Enhanced AC-3 with Joint Object Coding
Commercial name : Dolby Digital Plus with Dolby Atmos
Codec ID : A_EAC3
Duration : 1 h 35 min
Bit rate mode : Constant
Bit rate : 768 kb/s
Channel(s) : 6 channels
Channel layout : L R C LFE Ls Rs
Sampling rate : 48.0 kHz
Frame rate : 31.250 FPS (1536 SPF)
Compression mode : Lossy
Stream size : 525 MiB (3%)
Language : English (US)
Service kind : Complete Main
Default : Yes
Forced : Yes
Complexity index : 16
Number of dynamic objects : 15
Bed channel count : 1 channel
Bed channel configuration : LFE

N/A

No

### Reproduction steps

1. Install Kodi/Stremio/Plex/Jellyfin on Android TV.
2. Watch any video (mkv format) that have dolby vision/hdr10 using the ExoPlayer as player.

### Expected result

Should be able to watch mkv files that have dolby vision/hdr10  

### Actual result

The app freezes completely, and when it does, my TV becomes completely unresponsive for up to 5 minutes until the app crashes. During the freeze, none of the remote buttons work, and there's no power button on the TV itself, so I’m literally forced to play the waiting game every time this happens. One might suggest unplugging the power cord, but doing so several times a day isn’t exactly a great solution.

### Media

https://github.com/google/ExoPlayer/files/11150792/bugreport-kara-PS7624.3338N-2023-04-04-13-26-43.zip

### Bug Report

cgarst:
# Strip in-band HDR10+ SEI from HEVC bitstream during Dolby Vision Profile 8 playback

## Summary

Fixes freezes, black screens, and purple screens when playing DV Profile 8 MKV files that contain both Dolby Vision RPU and HDR10+ dynamic metadata on devices with buggy chipsets.

## Related Issues

Fixes #1895

## Problem

MKV files containing Dolby Vision Profile 8.1 (`dvhe.08`) combined with HDR10+ metadata cause playback failures on many Android TV devices:

- Fire TV Stick 4K Max (Gen 1 and Gen 2)
- Google TV Streamer (MediaTek MT8696)
- TCL C825 and MQLED80
- Other Android TV devices with similar chipsets

The HEVC base layer bitstream contains HDR10+ dynamic metadata as **in-band SEI NAL units** (ITU-T T.35 / ST 2094-40). When a native Dolby Vision codec processes this stream, it also handles DV RPU for dynamic HDR. The conflicting dynamic metadata crashes the chipset.

## Solution

Strip HDR10+ SEI NAL units from `buffer.data` in `onQueueInputBuffer()` before the buffer is submitted to the native DV codec. This is the same approach used by [Kodi](https://github.com/xbmc/xbmc/pull/24584).

The stripping **only** activates when both conditions are met:
1. The format is Dolby Vision Profile 8
2. A **native DV codec** was selected (`codecMimeType == VIDEO_DOLBY_VISION`)

When the TV doesn't support Dolby Vision, `doesDisplaySupportDolbyVision()` returns `false`, `getDecoderInfos()` falls back to HEVC, `codecMimeType` is `VIDEO_H265`, and stripping is **not** applied — HDR10+ metadata is preserved for the HEVC decoder.

> What if the device supports DV and HDR10+, but the TV only supports HDR10+?

In this case `doesDisplaySupportDolbyVision()` returns `false` → HEVC fallback is selected → `stripHdr10PlusSeiNalUnits` stays `false` → HDR10+ SEI is preserved → **no regression**. This scenario is covered by the `dvProfile8_hevcFallback_preservesHdr10PlusSeiInBuffer` integration test.

### `MediaCodecVideoRenderer.java`
- Added `stripHdr10PlusSeiNalUnits` boolean field, set in `onCodecInitialized()`
- Added stripping call at the top of `onQueueInputBuffer()`
- Added `stripHdr10PlusSeiFromBuffer()`: scans AnnexB buffer for start codes, identifies HEVC PREFIX\_SEI/SUFFIX\_SEI NAL units (types 39/40), parses SEI payloads to detect HDR10+ (ITU-T T.35 with country code `0xB5`, provider `0x003C`, oriented `0x0001`, app\_id 4, version 0 or 1), removes matching NAL units by compacting the buffer in-place

### `MediaCodecInfo.java`
- Reverted `isHdr10PlusOutOfBandMetadataSupported()` to VP9-only — the HEVC extension was unreachable for this bug

### `MediaCodecVideoRendererTest.java`
- Removed prior tests that targeted the unreachable out-of-band path
- Added unit tests for `stripHdr10PlusSeiFromBuffer()` and integration tests for the fix

### Bug fix validation

These tests confirm that HDR10+ SEI NAL units are correctly stripped from the HEVC bitstream before it reaches a native Dolby Vision codec. They fail on unpatched code and pass with the fix.

| Test | Purpose | Before fix | After fix |
|------|---------|------------|-----------|
| `stripHdr10PlusSei_removesHdr10PlusSeiNalUnit` | Unit test: given AnnexB buffer `[VCL][HDR10+ SEI][VCL]`, verifies the HDR10+ SEI NAL is removed, VCL NALs are preserved, and the buffer limit is reduced | **FAIL** — buffer unchanged | **PASS** |
| `dvProfile8_nativeDvCodec_stripsHdr10PlusSeiFromBuffer` | Integration test: full render pipeline with a native DV codec selected for DV Profile 8 content. Captures actual buffer data submitted to the codec and asserts no HDR10+ SEI NAL units remain | **FAIL** — HDR10+ SEI still present | **PASS** |

### Regression protection

These tests confirm that stripping does not activate when it shouldn't — non-HDR10+ SEI is never touched, buffers without SEI are left alone, and HEVC fallback playback preserves HDR10+ metadata for devices/TVs that need it.

| Test | Purpose | Before fix | After fix |
|------|---------|------------|-----------|
| `stripHdr10PlusSei_preservesNonHdr10PlusSei` | Unit test: given AnnexB buffer `[VCL][non-HDR10+ SEI][VCL]`, verifies all NAL units are preserved unchanged. Guards against accidentally stripping unrelated SEI messages (e.g. CEA-608 captions) | PASS | **PASS** |
| `stripHdr10PlusSei_noOpWhenNoSeiPresent` | Unit test: given AnnexB buffer with only VCL NAL units, verifies the buffer is completely unchanged. Guards against corruption of normal video frames | PASS | **PASS** |
| `dvProfile8_hevcFallback_preservesHdr10PlusSeiInBuffer` | Integration test: device has DV + HEVC codecs but TV only supports HDR10+ (no DV), so HEVC fallback is selected. Captures codec input buffers and asserts HDR10+ SEI NAL units are still present. | PASS | **PASS** |

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/video/MediaCodecVideoRenderer.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/video/MediaCodecVideoRenderer.java
@@ -67,6 +67,7 @@ import androidx.media3.common.util.Size;
 import androidx.media3.common.util.TraceUtil;
 import androidx.media3.common.util.UnstableApi;
 import androidx.media3.common.util.Util;
+import androidx.media3.container.NalUnitUtil;
 import androidx.media3.container.ObuParser;
 import androidx.media3.decoder.DecoderInputBuffer;
 import androidx.media3.exoplayer.CodecParameters;
@@ -211,7 +212,7 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer
   private @MonotonicNonNull CodecMaxValues codecMaxValues;
   private boolean codecNeedsSetOutputSurfaceWorkaround;
   private boolean codecHandlesHdr10PlusOutOfBandMetadata;
-  private boolean isDolbyVisionProfile8;
+  private boolean stripHdr10PlusSeiNalUnits;
   private @MonotonicNonNull VideoSink videoSink;
   private boolean hasSetVideoSink;
   private @VideoSink.FirstFrameReleaseInstruction int nextVideoSinkFirstFrameReleaseInstruction;
@@ -1552,7 +1553,7 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer
     codecNeedsSetOutputSurfaceWorkaround = codecNeedsSetOutputSurfaceWorkaround(name);
     codecHandlesHdr10PlusOutOfBandMetadata =
         checkNotNull(getCodecInfo()).isHdr10PlusOutOfBandMetadataSupported();
-    isDolbyVisionProfile8 =
+    stripHdr10PlusSeiNalUnits =
         isDolbyVisionProfile8(configuration.format)
             && MimeTypes.VIDEO_DOLBY_VISION.equals(configuration.codecInfo.codecMimeType);
     maybeSetupTunnelingForFirstFrame();
@@ -1629,6 +1630,9 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer
   @CallSuper
   @Override
   protected void onQueueInputBuffer(DecoderInputBuffer buffer) throws ExoPlaybackException {
+    if (stripHdr10PlusSeiNalUnits && buffer.data != null && buffer.data.remaining() > 0) {
+      stripHdr10PlusSeiFromBuffer(buffer.data);
+    }
     if (av1SampleDependencyParser != null
         && checkNotNull(getCodecInfo()).mimeType.equals(MimeTypes.VIDEO_AV1)
         && buffer.isKeyFrame()
@@ -1827,15 +1831,6 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer
   protected void handleInputBufferSupplementalData(DecoderInputBuffer buffer)
       throws ExoPlaybackException {
     if (!codecHandlesHdr10PlusOutOfBandMetadata) {
-      return;
-    }
-    // Workaround for https://github.com/androidx/media/issues/1895
-    // Skip HDR10+ metadata when a native Dolby Vision Profile 8 codec is active. DV Profile 8
-    // carries its own dynamic HDR metadata (RPU), so HDR10+ is redundant, and the conflicting
-    // metadata can cause severe issues (freezes, black screens). This is intentionally restricted
-    // to the native DV codec case: when falling back to HEVC base-layer decoding (e.g. on a device
-    // that doesn't support DV), the DV metadata is not processed and HDR10+ should still be used.
-    if (isDolbyVisionProfile8) {
       return;
     }
     ByteBuffer data = checkNotNull(buffer.supplementalData);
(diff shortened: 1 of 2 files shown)


mark-wiese:
OMG, @cgarst you are an absolute legend!

nift4:
Wouldn't this cause a regression if playback device doesn't support Dolby Vision, but does support HDR10+, and the file has both?
Currently it would play the DV file in backward compatible HDR10 mode and could use HDR10+ dynamic metadata, but with this patch it seems impossible.

cgarst:
@nift4 good call - this should be fixed now.

nift4:
Just to start with an disclaimer, I'm not sure about this as I don't have a test setup and my understanding is limited.

That said, I wonder if this could still cause a regression in the following case:

1. Playback device is not affected by the linked bug
2. Playback device supports DV and HDR10+
3. TV supports HDR10+ only
4. File has both DV and HDR10+

Edit: I was a bit mistaken in an initial analysis in this comment, sorry. (I edited it out now as it is not relevant, to prevent confusion.) Your prior fix seems to have already considered that scenario, and the change seems fine now.

cgarst:
@nift4 Thanks again for the feedback. The fix has been redone to handle this at a different level after digging deeper into how the HDR10+ metadata flows.

The HDR10+ in these MKV files is in-band as SEI NAL units in `buffer.data`, not out-of-band supplemental data, so the new approach strips HDR10+ SEI directly from the AnnexB buffer in `onQueueInputBuffer()`. It only kicks in when a native DV codec is actually active. It gates on `codecMimeType == VIDEO_DOLBY_VISION`, not just the format.

I've added the scenario you mentioned as an integration test (`dvProfile8_hevcFallback_preservesHdr10PlusSeiInBuffer`) to validate that HDR10+ SEI is preserved when HEVC fallback is selected.

I work on some additional testing with this change.

nift4:
You're right, that should just work with your new commit, because ExoPlayer already correctly falls back to HEVC codec, even if DV codec is present in STB but TV does not have DV. (I was mistaken about that not happening in my prior comment.) So everything should be good now!

tonihei:
Thanks for trying to solve this issue. 

The general idea probably works and we would be fine with merging it, but could you please
 - Rebase the pull request on the main branch (and not the release branch)? You may need to open a new PR to do this.
 - Limit the scope of the workaround to affected devices? From what I can tell this mainly affects Fire TV sticks, so there should be some conditional check for the workaround to only apply on certain devices/chipsets/codecs. 
 - Check if it really covers all the intended use cases, [this comment](https://github.com/androidx/media/issues/1895#issuecomment-4121292894) suggests the issue is not fully fixed by it, but instead produces black screens.

PR #3089, opened 2026-02-24
Fixed issue with ALAC 20-bit files not having PCM format detected

nift4:
After https://github.com/androidx/media/pull/2738 common alac files had bit depth populated correctly, but I recently learnt that there is also 20-bit alac files (you can create them with afconvert). it doesn't make any sense and it's very complex to handle 20-bit PCM in the media3 pipelines because one byte may belong to multiple samples, but we also don't need to. Instead of failing with `C.ENCODER_INVALID`, this change intercepts the 20-bit depth in the extractor and falls back to `Format.NO_VALUE`, expecting decoders to output byte-aligned 24-bit or 16-bit PCM instead. (test file created from other test files in this repo included.)


rohitjoins:
Thanks for putting this together!

While this approach definitely fixes the playback issues, introducing `C.ENCODING_PCM_20BIT` into the public API seems risky as it isn't byte-aligned.

What do you think about containing this workaround entirely within the extractor? We can simply intercept the 20-bit depth in `BoxParser.java` and map it to a safe value like `Format.NO_VALUE` (as we already do for `TYPE_pcmC`) or`C.ENCODING_PCM_24BIT`?

It would also be beneficial to add playback test for the 20 bit media file, to see the output produced by the decoder.

nift4:
> While this approach definitely fixes the playback issues, introducing C.ENCODING_PCM_20BIT into the public API seems risky as it isn't byte-aligned.

I agree it's not perfect, but it was the best I could think of given that...

> What do you think about containing this workaround entirely within the extractor? We can simply intercept the 20-bit depth in BoxParser.java and map it to a safe value like C.ENCODING_PCM_24BIT?

...I have various places in the app where I want to get the bit depth (for display purpose), currently using Util.getByteDepth(format.pcmEncoding)*8 to get the bit depth. Remapping 20-bit to 16 or 24 bit in extractor would end up with getting simply wrong values when querying.

To correctly parse it I'd now need to special case ALAC (to parse from initialization data) everywhere. And given that ALAC is not the only format supporting 20-bit files (WAV also does, WavPack too by extension) I wanted to propose a more general solution first.

Do you think introducing it, despite the warnings I added in the comments, is too risky? If so, I can still do the special-casing for the handful of formats, it's just less nice.

> It would also be beneficial to add playback test for the 20 bit media file, to see the output produced by the decoder.

The ALAC decoder is currently not part of AOSP, so I think that test would probably fail? Qualcomm has one in MediaCodec, and in my app I use a Java BSD-licensed implementation to support ALAC on every device: https://github.com/FoedusProgramme/Gramophone/blob/beta/misc/alacdecoder/src/main/java/com/beatofthedrum/alacdecoder/AlacDecodeUtils.java

rohitjoins:
Thank you providing the context! It makes sense why you would want to preserve the 20 bit value of the UI in your app.

> Do you think introducing it, despite the warnings I added in the comments, is too risky?

Yes, introducing this constant into the public API is still too risky for the library.

`C.PcmEncoding` constants are used deeply throughout the Media3 audio pipeline to allocate buffers and calculate sizes (for example, `Util.getPcmFrameSize()` relies directly on `Util.getByteDepth()`). It will result in player to crash if used for this non-standard bit depth.

> If so, I can still do the special-casing for the handful of formats, it's just less nice.

`Format` object's primary job is to safely configure the playback pipeline. For non standard codec parameters that the pipeline cannot handle, parsing initialization data at app level is unfortunately the safer option. 

> The ALAC decoder is currently not part of AOSP, so I think that test would probably fail?

Agreed. It is fine to not add the playback test in this case. Even fine to remove the extractor tests with dump files, if we just fallback to `Format.NO_VALUE`. 

Falling back to `Format.NO_VALUE` for handling unsupported bit depths is something we already do in `MatroskaExtractor` and `BoxParser`, so we can do the same here.

nift4:
I have changed it, it should be good now.

nift4:
Thanks!

PR #3090, opened 2026-02-24
Support double pcm encoding in extractor

nift4:
- now that we have ENCODING_PCM_DOUBLE, introduce a helper for it. we get wav/mka/mp4 64-bit float support for free. (test files created from other test files in this repo included.)
- Adds missing pcm formats to ToFloatPcmAudioProcessor, incl tests.


nift4:
(I fixed the conflict with the web UI, I hope that worked properly.)

nift4:
Thanks!

Issue #2559, opened 2025-06-24
Missing parsing of 'displayAlign' from <style> tags from Ttml subtitles


klilian:
### Version

Media3 main branch

### More version details

Issue present in main 1.8.0-beta01 branch

Example of ttml subtitles where 'displayAlign' is not correctly parsed from existing TtmlParser.java class:
```
<?xml version="1.0" encoding="UTF-8"?><tt xml:lang="it" xmlns="http://www.w3.org/ns/ttml" xmlns:ttm="http://www.w3.org/ns/ttml#metadata" xmlns:tts="http://www.w3.org/ns/ttml#styling" xmlns:smpte="http://www.smpte-ra.org/schemas/2052-1/2010/smpte-tt" ><head>
<smpte:information smpte:mode="Enhanced" />
<styling>
<style xml:id="backgroundStyle" tts:fontFamily="proportionalSansSerif" tts:textAlign="center" tts:origin="0% 66%" tts:extent="100% 33%" tts:backgroundColor="rgba(0,0,0,0)" tts:displayAlign="center" />
<style xml:id="speakerStyle" style="backgroundStyle" tts:color="white" tts:textOutline="black 1px" tts:backgroundColor="transparent" />
<style xml:id="textStyle" style="speakerStyle" tts:color="white" tts:textOutline="none" tts:backgroundColor="black" />
</styling><layout>
<region xml:id="full" tts:origin="0% 0%" tts:extent="100% 100%" tts:zIndex="1" />
<region xml:id="speaker" style="speakerStyle" tts:zIndex="1" />
<region xml:id="background" style="backgroundStyle" tts:zIndex="0" />
</layout>
</head><body>
<div>
<p region="speaker" begin="00:19:32.640" end="00:19:36.160"><span style="textStyle">Lorem ipsum dolor sit amet consectetur adipiscing elit<br/>Quisque faucibus ex sapien vitae pellentesque sem placerat</span></p>
<p region="speaker" begin="00:19:36.320" end="00:19:38.560"><span style="textStyle">Ad litora torquent per conubia nostra inceptos himenaeos.</span></p>
<p region="speaker" begin="00:19:45.880" end="00:19:48.800"><span style="textStyle">Quisque faucibus ex sapien vitae pellentesque sem placerat</span></p>
<p region="speaker" begin="00:19:52.040" end="00:19:55.040"><span style="textStyle">Ad litora torquent per conubia nostra inceptos himenaeos.</span></p>
</div>
</body></tt>
 
[...]

PR #3094, opened 2026-02-28
Fallback to `displayAlign` from `style` for TTML regions

FineFindus:
Adds support for using the `displayAlign` from the `style` of a region, if it has no explicit `displayAlign` set.

Closes: https://github.com/androidx/media/issues/2559

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/text/ttml/TtmlParser.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/text/ttml/TtmlParser.java
@@ -462,6 +462,15 @@ public final class TtmlParser implements SubtitleParser {
     @Nullable
     String displayAlign =
         XmlPullParserUtil.getAttributeValue(xmlParser, TtmlNode.ATTR_TTS_DISPLAY_ALIGN);
+    if (displayAlign == null) {
+      String styleId = XmlPullParserUtil.getAttributeValue(xmlParser, TtmlNode.ATTR_STYLE);
+      if (styleId != null) {
+        TtmlStyle style = globalStyles.get(styleId);
+        if (style != null) {
+          displayAlign = style.getDisplayAlign();
+        }
+      }
+    }
     if (displayAlign != null) {
       switch (Ascii.toLowerCase(displayAlign)) {
         case "center":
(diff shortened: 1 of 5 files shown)


libraries/extractor/src/main/java/androidx/media3/extractor/text/ttml/TtmlParser.java
```diff
@@ -462,6 +462,15 @@ private static TtmlRegion parseRegionAttributes(
     @Nullable
     String displayAlign =
         XmlPullParserUtil.getAttributeValue(xmlParser, TtmlNode.ATTR_TTS_DISPLAY_ALIGN);
+    if (displayAlign == null) {
+      String styleId = XmlPullParserUtil.getAttributeValue(xmlParser, TtmlNode.ATTR_STYLE);
+      if (styleId != null) {
+        TtmlStyle style = globalStyles.get(styleId);
+        if (style != null) {
+          displayAlign = style.getDisplayAlign();
```
icbaker, line 470: I think for consistency with the way we handle other attributes, this 'inheritance' logic should be implemented as part of `TtmlStyle.chain` which is invoked here: https://github.com/androidx/media/blob/8bafa18c4d1da208ab1286cd05931bf3e8bf802c/libraries/extractor/src/main/java/androidx/media3/extractor/text/ttml/TtmlParser.java#L309

See my comment here: https://github.com/androidx/media/issues/2559#issuecomment-3070066315

icbaker (changes requested):

libraries/extractor/src/main/java/androidx/media3/extractor/text/ttml/TtmlParser.java
FineFindus: Done, should I also add a test for inheriting `displayAlign` from another style?

icbaker: You can probably tweak your existing test file to exercise this?

FineFindus: Done.

PR #3098, opened 2026-03-03
Prototype for Track selection state comoposable

amanc3765:
@tonihei @oceanjules

This is just a prototype for `TrackSelection` composable state. 
I have tried to implement **Option 1** and **Option 2** mentioned in the doc. It is not meant to be merged.


oceanjules:
Developement moved to https://github.com/androidx/media/blob/main/libraries/ui_compose/src/main/java/androidx/media3/ui/compose/state/TrackSelectionState.kt

PR #3102, opened 2026-03-04
Handle tunnelingAudioSessionId set after renderer enabled.

zgzong:
As now the audio session is not immediately available when create player, which may cause that the tunnelingAudioSessionId set after renderer onEnabled, that will cause checkState failure as tunnelingAudioSessionId has not been set yet.


microkatz:
@zgzong 

Thank you for offering a fix for the issue. I believe that we will be going another way than this and instead sending the MSG_SET_AUDIO_SESSION_ID to the Renderers in the same process as the audio session id generation. You are free to close this ticket or I will message this thread and close the ticket after our solution is pushed.

zgzong:
@microkatz Thanks for looking into this. Looking forward to hearing updates from you soon.

PR #3104, opened 2026-03-05
Allow to use float output in any case if output provider doesn't support 16-bit int

nift4:
This commit is a continuation of 4f4ad5f82d156767f89b9a6dec44c37bb8d153f4 and fe4f235c103167c5ee8be4f713ebf264b72f6bec by Toni, decoupling DefaultAudioSink logic from assumptions that AudioTrack is used.

* If audio output provider doesn't support 16-bit int PCM, we should use float PCM in any case, even if input isn't high-resolution PCM
* Unlike when this check was written, ToFloatPcmAudioProcessor now supports every linear PCM format, similar to ToIntPcmAudioProcessor, it is no longer limited to high-resolution only
* While AudioTrack (and by extension AudioTrackAudioOutputProvider) always supports 16-bit int (hence, this change is a no-op for them), it may not apply to other ways of outputting PCM
* Additionally, the same applies to the provider possibly not supporting float PCM output. Currently, we have the enableFloatOutput option, but this is scheduled to be replaced by a more generic one. While we're here, we can address it as well by adding a check.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/DefaultAudioSink.java
@@ -686,7 +686,7 @@ public final class DefaultAudioSink implements AudioSink {
     // For PCM formats, convert the format to what our audio processors will produce.
     boolean transcodingViaAudioProcessors = false;
     if (Util.isEncodingLinearPcm(format.pcmEncoding)) {
-      boolean usesFloatPcm = shouldUseFloatOutput(format.pcmEncoding);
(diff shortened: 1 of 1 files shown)


tonihei:
An audio output **not** supporting 16-bit PCM sounds very unusual. I appreciate that the abstraction of the AudioTrack-based output makes this theoretically possible, but I wonder if you have a concrete use case in mind? 

(PR looks good otherwise, just making sure we really need this bit of logic)

nift4:
I was thinking of a custom PCM mixer for my WIP libusb output and that should use the highest determinator format. A PCM mixer can only mix what has the same format, and I intend to do with float32 for being said highest determinator. So I need to convert all audio to float32.

Issue #2215, opened 2025-03-06
The streaming continues beyond the duration if VMAP has scheduled advertises after duration itself


chemickypes:
### Version

Media3 1.5.1

### More version details

Media3 1.5.1

### Devices that reproduce the issue

All

### Devices that do not reproduce the issue

N/A

### Reproducible in the demo app?

Not tested

### Reproduction steps

Using a VMAP file with advertise scheduled beyond the content duration you will see that streaming continues to the first ads possible after the duration and than it finishes.

### Expected result

The streaming finishes when it reaches the duration.

### Actual result

the Streaming continues beyond the duration till the first ads after this time.

### Media

[Here](https://drive.google.com/drive/folders/1RHdjqHsGt_zbh55EDzWABZUnekNqPQ_9?usp=sharing) two images to show that buffered position is greater than window duration and that the buffered position is the same of the adBreaks.

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

chemickypes:
yes you understood the problem but between the end of the content and the commercial, the streaming is frozen and the user have to wait for the time gap between the end and the ads break.

I am going to write you an email with link so you can test by yourself.

marcbaechinger:
Sorry for my late response and thankd for providing the assets for testing.

I was able to use the media an VMAP file you sent  by email. Besides the pre roll the ads have a load errors, but generally I think it confirms what we discussed above.

The VMAP file delivers ad cue points that are after the end of the window. When then the end of the window is reached, the player attemps to play these ads which at this point don't have a URI. The player waits for the ads SDK to deliver the URIs, but because the position at which the SDK would start fetching the ad data is never reached, the SDK never delivers an URI and the player stalls playback waiting for that. That's how I understand what's going on.

I think the best solution for this would be to not ad these ad groups after the end of the window to the `AdPlaybackState`, but this isn't easily possible with the current approach. The reason is that the cue points are inserted to the `AdPlaybackState` at the very first moment they are delivered from the SDK and before the content timeline is available, that would deliver the duration of the window. This is required to start with the pre roll as soon as possible.

Since Media3 1.6, `AdsMediaSource` allows the ad groups to grow. So as an alternative, there would probably a way to defer insertion of all ad groups except the pre roll, then wait for the timeline and start inserting the other ad groups by taking into account the now available duration of the window. That would be a rather larger change.

[...]

PR #3126, opened 2026-03-19
Multi period content support for AdsMediaSource

kotucz:
Multi period DASH content ads support inspired by https://github.com/androidx/media/issues/1642#issuecomment-2772077256
Reopens: https://github.com/androidx/media/pull/2501

At Timeline level, the original `AdPlaybackState` is divided and adjusted for each of the periods.

Removing single period assertions from `AdsMediaSource`.
Providing `MultiPeriodAdTimeline` as an alternative to current `SinglePeriodAdTimeline`.
From input `AdPlaybackState`, new modified `AdPlaybackState` is created - one for each period:

- ad group time is offset relative to period start time
- ad groups after period end time are marked as skipped (not needed now thanks to https://github.com/androidx/media/commit/ed56ed22fb30ff66d75cc79aec33636c46aaf921
https://github.com/androidx/media/issues/2215)
- post-roll ad group is kept only for last period
- ad group count and indices are kept constant

More detail is also here:
https://blog.hotstar.com/monetisation-multi-period-dash-x-exoplayer-db8e8c00e521

**We have encountered an issue, which is also handled in this PR:**

Player is using `MaskingMediaSource` to reduce start lag. This means preroll ads are loaded in parallel to resolving content manifest. Meanwhile placeholder manifest is used for content temporarily having single period.

Ads are initialized as follows:
- preroll - skipped (no fill, there is no preroll)
- the rest of the midrolls are playable, but not in the period 0

After the content manifest (MPD) is resolved content will have multiple periods.
For each period we will create an adjusted copy the main `AdPlaybackState` to have specific config of ads.

Preroll will stay unchange (no fill), while midrolls for period 0 (bumper/disclaimer) will be marked as SKIPPED as these should not be played in this period.

Note: if preroll is available it refreshes the timeline when it is done playing at this moment real timeline with periods is available and `AdPlaybackState`s are reflected as expected

ExoPlayer has evaluates this update can be ignored. (Placeholder and real timeline have same Id and are not distinguished)
Player plays until the ad group is reached. But since the ad group is only some time after the period end, player is stuck in the end of period 0, waiting for the midroll to be played.

Note: Seekbar or fast forward button can be used to unstuck the player (forcing Timeline update).

FIX: Player timeline will be recreated when period count is changed

Example:

```
-------------- placeholder period -------------------

|      period 0          |           period 1                                          ...          | end of content
0s                       5s                                         15s
preroll(empty)                                                     midroll
```
What happens:
- playback period 0 till its end 5s
- stuck for 10 sec in the end of period 0
- play midroll at 15s
- play period 1

Midroll is not in period 0, but this change is not reflected in the Timeline on the placeholder timeline -> multiperiod timeline transition
```
periodIdWithAds.nextAdGroupIndex -1
oldPeriodId.nextAdGroupIndex 1
```


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/ExoPlayerImplInternal.java
```diff
@@ -3670,12 +3670,15 @@ private static PositionUpdateForPlaylistChange resolvePositionForPlaylistChange(
...
                 && periodIdWithAds.nextAdGroupIndex >= oldPeriodId.nextAdGroupIndex);
+    final boolean samePeriodCount =
+        playbackInfo.timeline.getPeriodCount() == timeline.getPeriodCount();
     // Drop update if we keep playing the same content (MediaPeriod.periodUid are identical) and
     // the only change is that MediaPeriodId.nextAdGroupIndex increased. This postpones a potential
     // discontinuity until we reach the former next ad group position.
     boolean sameOldAndNewPeriodUid = oldPeriodId.periodUid.equals(newPeriodUid);
     boolean onlyNextAdGroupIndexIncreased =
         sameOldAndNewPeriodUid
+            && samePeriodCount
```
tonihei: This specific condition about the having the same number of periods is probably too restrictive overall (imagine using the AdsMediaSource in a playlist with other items).

Is it possible to check for `Timeline.Period.isPlaceholder` in this condition to not apply this check when we go from placeholder to non-placeholder? This would of course only work if the `isPlaceholder` flag is changing at the right moment too. Alternatively, I wonder if we need additional checks for this logic to verify that the time of the ad group hasn't changed or is within the period duration. As you can see from the comment, the purpose of this check has really nothing to do with the multi-period ads, but with failed future ads that are not supposed to interrupt playback immediately.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/AdsMediaSource.java
```diff
@@ -510,7 +510,11 @@ private void maybeUpdateSourceInfo() {
         refreshSourceInfo(contentTimeline);
       } else {
         adPlaybackState = adPlaybackState.withAdDurationsUs(getAdDurationsUs());
-        refreshSourceInfo(new SinglePeriodAdTimeline(contentTimeline, adPlaybackState));
+        if (contentTimeline.getPeriodCount() == 1) {
+          refreshSourceInfo(new SinglePeriodAdTimeline(contentTimeline, adPlaybackState));
```
tonihei: The `SinglePeriodAdTimeline` now sounds like a subset of the `MultiPeriodAdTimelime` - could this just be changed to be simply an `AdTimeline`? We likely need to keep `SinglePeriodAdTimeline` as deprecated to avoid breakages, but otherwise it looks like the new class can be just be used instead of the old one.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/AdTimeline.java
```diff
@@ -0,0 +1,101 @@
...
+
+
+/**
+ * A custom {@link Timeline} for sources that have {@link AdPlaybackState} split among multiple periods.
+ * <br/>
+ * For each period a modified {@link AdPlaybackState} is created for each period:
+ * <ul>
+ * <li> ad group time is offset relative to period start time </li>
+ * <li> post-roll ad group is kept only for last period </li>
+ * <li> ad group count and indices are kept unchanged </li>
```
tonihei: This part is surprising to me because it sounds like we would play all ads repeatedly for all periods. I assume some of the internal player logic prevents that from happening, but it would be much cleaner if we could actually split the AdPlaybackState fully into the individual ads for each period.

tonihei: Second note: When looking for other adjustments that might be needed in `AdsMediaSource`, I realized you avoided making these adjustments by keeping the `adGroupIndex` stable across all periods. So if you follow my suggestion here to split the `AdPlaybackState` properly, the logic in `AdsMediaSource.createPeriod/releasePeriod/onChildSourceInfoRefreshed` and potentially other places would need to change to reverse the mapping from `adGroupIndex` in a period to `adGroupIndex` in the 'global' `AdPlaybackState`. 

Given all that, I think your approach of keeping the indices stable without further mapping is actually quite useful. I'm curious to hear whether you tried the other idea of splitting the state further still.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/AdTimeline.java
```diff
@@ -0,0 +1,101 @@
...
+    for (int adGroupIndex = 0; adGroupIndex < adPlaybackState.adGroupCount; adGroupIndex++) {
+      final long adGroupTimeUs = adPlaybackState.getAdGroup(adGroupIndex).timeUs;
+      if (adGroupTimeUs == C.TIME_END_OF_SOURCE) {
+        if (!isLastPeriod) {
+          adPlaybackState = adPlaybackState.withSkippedAdGroup(adGroupIndex);
+        }
+      } else {
+        // start time relative to period start
+        adPlaybackState = adPlaybackState.withAdGroupTimeUs(adGroupIndex,
+            adGroupTimeUs - periodStartOffsetUs);
```
tonihei: If this is negative or beyond the period duration, should it also be marked as "skipped" to more clearly signal that this ad group should not be played for this period?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/MultiPeriodAdTimeline.java
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright (C) 2017 The Android Open Source Project
```
tonihei: 2026 :)

tonihei:
Thanks for the pull request (and rebasing your earlier attempt from https://github.com/androidx/media/pull/2501)! I'll leave some comment for discussion, but overall looks like the right approach.

tonihei:
Re-reading the discussion of https://github.com/androidx/media/issues/1642#issuecomment-3036919802, I also wonder if you thought about adjustments needed in `AdsLoader` related classes like IMA's `AdTagLoader`. I see a few places that retrieve positions from the player or access `AdPlaybackState` information via the `Timeline` obtained from the `Player` object, which may now have the per-period adjustments. Did you see any issues with that or were you using your own custom `AdsLoader`?

kotucz:
> Re-reading the discussion of [#1642 (comment)](https://github.com/androidx/media/issues/1642#issuecomment-3036919802), I also wonder if you thought about adjustments needed in `AdsLoader` related classes like IMA's `AdTagLoader`. I see a few places that retrieve positions from the player or access `AdPlaybackState` information via the `Timeline` obtained from the `Player` object, which may now have the per-period adjustments. Did you see any issues with that or were you using your own custom `AdsLoader`?

One thing I updated in our `AdsLoader` (which is similar to `AdTagLoader` ) is calculating `contentDurationUs` in `onTimelineChanged`
```
        val contentDurationUs =
            // sum of period durations
            (0 until timeline.periodCount).sumOf { i ->
                timeline.getPeriod(i, period).durationUs
            }
```
Previously there was also `AdTagLoader: checkArgument(timeline.getPeriodCount() == 1);` which was removed in this commit: `andrewlewis* 23-Nov-20 19:29 Add support for ad playlists with ImaAdsLoader`
I see difference in `long contentDurationUs = timeline.getPeriod(player.getCurrentPeriodIndex(), period).durationUs;` where there was `periodIndex = 0` so not sure if this is applicable.


In most cases in our `AdsLoader` we are using `player.currentAdIndexInAdGroup` and `player.currentAdGroupIndex`. Thanks to keeping indices stable, the rest was working out of the box.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/AdTimeline.java
kotucz: We kept it intentionally. This ensures ads play across period boundaries. As if the content was single period.
E.g. if user seeks to credits period in the end. The last midroll should still play, even if it is in the content period.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/AdsMediaSource.java
kotucz: Renamed to `AdTimeline`. I think it behaves same as `SinglePeriodAdTimeline` for the case when period count == 1. But still rather kept it for compatibility.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/ExoPlayerImplInternal.java
kotucz: I had to think about this. I implemented more specific check:
If the cuepoint being removed is not part of the period anymore, need to trigger update.
The update may not be postponed as the cuepoint will not be reached before the period ends. 

Note: I am bit hesitant about using `period` being mutable in/out argument. Are there any best practices for using this?

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/AdTimeline.java
kotucz: Ad states are updated in the main `AdPlaybackState` in `AdsLoader`. On every update it is being split into the multiple `AdPlaybackState`s with the actual states (e.g. played/skipped). The child `AdPlaybackState`s are not updated individually.

> I'm curious to hear whether you tried the other idea of splitting the state further still.

In development we tried this. We have tried where period 0 `AdPlaybackState` had only preroll (or empty ad group).
Before content manifest is resolved, `MaskingMediaSource` provides placeholder `Timeline`.
At this point, the timeline has single period and so `SinglePeriodAdTimeline` is used (to minimize code changes impact on single period content). As a result, fist period has full `AdPlaybackState` and **nextAdGroup is 1.** 

After content manifest is resolved, `MultiPeriodAdTimeline` is used and now 0th period has single empty pre-roll ad group. No ad group with index 1 results in AIOOB exception.

```
java.lang.ArrayIndexOutOfBoundsException: length=1; index=1
	at com.google.android.exoplayer2.source.ads.AdPlaybackState.getAdGroup(AdPlaybackState.java:507)
	at com.google.android.exoplayer2.Timeline$Period.getAdGroupTimeUs(Timeline.java:728)
	at com.google.android.exoplayer2.MediaPeriodQueue.getUpdatedMediaPeriodInfo(MediaPeriodQueue.java:413)
	at com.google.android.exoplayer2.MediaPeriodQueue.updateQueuedPeriods(MediaPeriodQueue.java:350)
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMediaSourceListInfoRefreshed(ExoPlayerImplInternal.java:1824)
	at com.google.android.exoplayer2.ExoPlayerImplInternal.mediaSourceListUpdateRequestedInternal(ExoPlayerImplInternal.java:742)
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:538)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.os.HandlerThread.run(HandlerThread.java:67)
```
	
I think there would be more issues, like you described a lot of changes. So I think this was easier approach.

jeanalexandre-iragne-deltatre:
@kotucz  do wou want to also apply this change in AdTagLoader ?

kotucz:
@jeanalexandre-iragne-deltatre

> @kotucz do wou want to also apply this change in AdTagLoader ?

I have tried to implement this. Found it is more complex as we usually had single content.
With the same approach as above, these tests were failing:
`playbackWithTwoAdsMediaSources_preloadsSecondAdTag`
`playbackWithTwoAdsMediaSources_preloadsSecondAdTagWithBackgroundResume`
I tried to fix it for multiple media items/windows. See https://github.com/androidx/media/pull/3126/changes/9b0233c77c2ef88b894bb0ef192d24643699b121
However, I am struggling writing tests for this change
There seems to be mismatch in `TimelineWindowDefinition.periodCount` in some places.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/source/ads/AdTimeline.java
tonihei: Interesting, that makes a lot of sense actually.

libraries/exoplayer_ima/src/main/java/androidx/media3/exoplayer/ima/AdTagLoader.java
```diff
@@ -490,7 +490,23 @@ public void onTimelineChanged(Timeline timeline, @Player.TimelineChangeReason in
     }
     Player player = this.player;
     this.timeline = timeline;
-    long contentDurationUs = timeline.getPeriod(player.getCurrentPeriodIndex(), period).durationUs;
+    long contentDurationUs;
```
tonihei: I'll continue without touching AdTagLoder for now. Given you have a custom logic, the easiest way to move forward is to merge the general functionality, add assertions in the IMA AdTagLoader about only supporting a single period still and then change that separately.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/ExoPlayerImplInternal.java
tonihei: Thanks, that should work I think. The APIs around Timeline.Period are quite old and a bit clunky. Best practices within our codebase are to never assume anything about such an object when passed to method (only use it as a 'scratch' instance to write to and then use the value immediately). I'll the logic accordingly.

tonihei:
I'll continue the internal merge process now, you may see some updates here based on the internal review. Please refrain from uploading new changes as it may interfere with that.

One problem is though that your CLA doesn't seem to work anymore. I'm pretty sure it passed the test before. Could you check again so that I can import the change? Thanks!

kotucz:
> One problem is though that your CLA doesn't seem to work anymore. I'm pretty sure it passed the test before. Could you check again so that I can import the change? Thanks!

@tonihei I recently left the company. I will contact them, how we can handle the CLA now

Issue #2487, opened 2025-05-30
Proper handling of specific DTS-HD profiles


nift4:
### [REQUIRED] Use case description
I am trying to determine in my app whether the playback format is lossless or not, and display format to the user so that they can verify if setup works correctly.
However, even though https://www.iana.org/assignments/media-types/audio/vnd.dts.hd states:
> profile is a required parameter.

ExoPlayer cannot detect whether a file is DTS-HD, DTS-HD High Resolution Audio or DTS-HD Master Audio profile, and uses a mime type without profile parameter for everything except LBR.

https://github.com/androidx/media/blob/ae9a2339679826f870690092509a135dffc0d7e4/libraries/common/src/main/java/androidx/media3/common/MimeTypes.java#L90

For example, in MP4 extractor:
https://github.com/androidx/media/blob/ae9a2339679826f870690092509a135dffc0d7e4/libraries/common/src/main/java/androidx/media3/common/MimeTypes.java#L476-L477
even "dtsl" box which is "Extension Substream containing only XLL" -> lossless -> mime should be `audio/vnd.dts.hd;profile=dtsma`

Another example, ts extractor using DtsUtil always sets DTS Express mime type:
https://github.com/androidx/media/blob/ae9a2339679826f870690092509a135dffc0d7e4/libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java#L516
but in ETSI TS 102 114 V1.6.1 Annex F a non-LBR format is explicitly mentioned as supported: "For DTS-HD Lossless formats, the value of [...]" which means DTS-HD MA profile is supported for sure, and I see no indication that other DTS-HD supports are not supported at all, hence this hardcoding to LBR is wrong.

### Proposed solution
Edit ExoPlayer and extractors to extract and pass around accurate DTS-HD profile information in the mime type, including:
- MP4 extractor
- TS extractor
- MKV extractor

and do not omit the required profile parameter in the mime type.

### Alternatives considered
N/A

nift4:
Some amendment, 

https://github.com/androidx/media/blob/ae9a2339679826f870690092509a135dffc0d7e4/libraries/common/src/main/java/androidx/media3/common/MimeTypes.java#L92

AUDIO_DTS_X also is a somewhat misleading name because there is a variant of DTS:X that builds on DTS-HD MA (and that is significantly more widespread than DTS-UHD content). 
[...]

PR #3138, opened 2026-03-24
Rename AUDIO_DTS_X to AUDIO_DTS_UHD_P2

nift4:
* The AudioFormat constant is named ENCODING_DTS_UHD_P2, and the internal platform mime type constant too is named MEDIA_MIMETYPE_AUDIO_DTS_UHD_P2. It's a correct and clear way to refer to this format.
* DTS:X is a marketing name that can be misleading because DTS:X is a proprietary extension to other DTS formats, however DTS:X supports both DTS-HD based and DTS-UHD based formats. Equating DTS-UHD with DTS:X is bound to cause confusion when DTS-HD based DTS:X is encountered.

Issue: #2487


icbaker:

nift4:
Thanks!

PR #3139, opened 2026-03-24
Support float big endian in mp4 extractor

nift4:
Completes ISO 23003-5 support


icbaker:

nift4:
Thanks for the quick review :)

PR #3141, opened 2026-03-25
Add support for parsing chapters from MKV files

tymmesyde:
Hi, 
this include changes to support parsing MKV chapters using the recent changes to introduce [metadata Chapter interface](https://github.com/androidx/media/commit/1cf665ba4774cbdbcd817b1ddb45bfb4e7dab7af)
This is already implemented for MP4: [Nero](https://github.com/androidx/media/commit/8e274454b431068f102fae2d17bcbb91099d44ab#diff-48909297df550e21ea3cac7a0a1ba76b61fb1a4a25f2a62059ae04507da7cbde) & [QuickTime](https://github.com/androidx/media/commit/0fde8633a190f31b237f00d087ca922ca42b1912)
Used this doc as reference: https://www.matroska.org/technical/elements.html

Added `sample_with_chapters.mkv` for `mkvSample_withChapters` test
Generated with:

```
;FFMETADATA1
[CHAPTER]
TIMEBASE=1/1000
START=0
END=1999
title=Chapter 1

[CHAPTER]
TIMEBASE=1/1000
START=2000
END=4999
title=Chapter 2
```

```
ffmpeg \                                                                                       
  -f lavfi -i testsrc=duration=5:size=320x240:rate=30 \
  -f lavfi -i sine=frequency=440:duration=5 \
  -i chapters.txt \
  -map 0:v -map 1:a -map_chapters 2 \
  -c:v libx264 -c:a aac \
  sample_with_chapters.mkv
```


nift4:
How many of the chapter features does this support?
https://github.com/hubblec4/Matroska-Playback/blob/master/src/PlayerOverview.md

icbaker:
The tests seem to fail at the moment with

```
deduplicateConsecutiveFormats=false so TrackOutput must receive at least one sampleMetadata() call between format() calls.
```

Please can you get them passing?

tymmesyde:
> The tests seem to fail at the moment with
> 
> ```
> deduplicateConsecutiveFormats=false so TrackOutput must receive at least one sampleMetadata() call between format() calls.
> ```
> 
> Please can you get them passing?

They should pass now sorry, i guess i didn't actually run them after setting WRITE_TO_LOCAL
This is what was causing them to fail: https://github.com/androidx/media/pull/3141/changes/81efac3e59e24820bdb66728df1aa322519afafc#diff-e2187c4432661736c918e5ca266afbeec7c6709b184729dec6a4aa613abf0c0aL968, seems like this should only be called by the logic that handle the samples?

tymmesyde:
> How many of the chapter features does this support? https://github.com/hubblec4/Matroska-Playback/blob/master/src/PlayerOverview.md

Only Basic Chapters, and it only expose the title (ChapString) and timeStart/timeEnd for the [Chapter](https://github.com/androidx/media/blob/main/libraries/extractor/src/main/java/androidx/media3/extractor/metadata/Chapter.java#L41) interface

icbaker:
@nift4 Since the media3 `Chapter` interface is new, we have an opportunity to change it before it gets included in a release.

In particular I'm wondering whether to add a concept of 'disabled/enable' and/or 'hidden/shown' to the `Chapter` interface to reflect these fields from the Matroska spec. I think all other features we might want to support can be safely added later through additional interface methods, possibly with default implementations, e.g. multiple titles with associated languages. But the `isDisabled`/`isHidden` methods probably need to be on the interface from the start (so apps know to check them).

It sems like a 'hidden' chapter should still be played, but just not shown to a user in a Table-of-Contents (i.e. this can be implemented "app side").

Whereas a 'disabled' chapter should be skipped during playback. I think we are unlikely to add automatic support for this part in the Player (at least not for a while). It could probably still be implemented app-side using [player messages](https://developer.android.com/media/media3/exoplayer/listening-to-player-events#firing-events) or similar.

Do you think it makes sense to add these methods to the 'generic' `Chapter` interface? They would return 'enabled' and 'not hidden' for other sources of chapters that don't encode the same concept (e.g. MP4 Nero, or ID3).

nift4:
@icbaker Yes, I think that's a good idea.

icbaker:
@tymmesyde Would it be possible to open a new PR from an individual-owned fork? We can't push changes to organization-owned forks like this one unless we have collaborator access. If that's not possible then we can still merge this PR but it will result in an ['evil' merge](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefevilmergeaevilmerge). See more info here: https://github.com/androidx/media/blob/release/CONTRIBUTING.md#push-access-to-pr-branches

tymmesyde:
sure, here you go: https://github.com/androidx/media/pull/3144

tymmesyde:
Closed in favor of https://github.com/androidx/media/pull/3144

icbaker:
> @icbaker Yes, I think that's a good idea.

Thanks for the feedback - I plan to merge this PR (or rather https://github.com/androidx/media/pull/3144) roughly as-is, then tweak the interface (and the matroska integration with it) in a follow-up change. That will be less annoying than needing to rebase this PR past the changes.

PR #3144, opened 2026-03-27
Add support for parsing chapters from MKV files

tymmesyde:
Hi, 
this include changes to support parsing MKV chapters using the recent changes to introduce [metadata Chapter interface](https://github.com/androidx/media/commit/1cf665ba4774cbdbcd817b1ddb45bfb4e7dab7af)
This is already implemented for MP4: [Nero](https://github.com/androidx/media/commit/8e274454b431068f102fae2d17bcbb91099d44ab#diff-48909297df550e21ea3cac7a0a1ba76b61fb1a4a25f2a62059ae04507da7cbde) & [QuickTime](https://github.com/androidx/media/commit/0fde8633a190f31b237f00d087ca922ca42b1912)
Used this doc as reference: https://www.matroska.org/technical/elements.html

Added `sample_with_chapters.mkv` for `mkvSample_withChapters` test
Generated with:

```
;FFMETADATA1
[CHAPTER]
TIMEBASE=1/1000
START=0
END=1999
title=Chapter 1

[CHAPTER]
TIMEBASE=1/1000
START=2000
END=4999
title=Chapter 2
```

```
ffmpeg \                                                                                       
  -f lavfi -i testsrc=duration=5:size=320x240:rate=30 \
  -f lavfi -i sine=frequency=440:duration=5 \
  -i chapters.txt \
  -map 0:v -map 1:a -map_chapters 2 \
  -c:v libx264 -c:a aac \
  sample_with_chapters.mkv
```


libraries/extractor/src/main/java/androidx/media3/extractor/mkv/MatroskaExtractor.java
```diff
@@ -190,6 +190,17 @@ public static ExtractorsFactory newFactory(SubtitleParser.Factory subtitleParser
...
   private static final int ID_SEGMENT_INFO = 0x1549A966;
+  private static final int ID_CHAPTERS = 0x1043A770;
+  private static final int ID_EDITION_ENTRY = 0x45B9;
+  private static final int ID_CHAPTER_FLAG_HIDDEN = 0x98;
+  private static final int ID_CHAPTER_ATOM = 0xB6;
+  private static final int ID_CHAPTER_UID = 0x73C4;
+  private static final int ID_CHAPTER_TIME_START = 0x91;
+  private static final int ID_CHAPTER_TIME_END = 0x92;
+  private static final int ID_CHAPTER_TRACK = 0x8F;
+  private static final int ID_CHAPTER_TRACK_UID = 0x89;
```
icbaker, line 203: I didn't find this documented in https://datatracker.ietf.org/doc/rfc9559/ - where is this defined?

libraries/extractor/src/main/java/androidx/media3/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1103,6 +1166,9 @@ protected void integerElement(int id, long value) throws ParserException {
       case ID_TRACK_NUMBER:
         getCurrentTrack(id).number = (int) value;
         break;
+      case ID_TRACK_UID:
+        getCurrentTrack(id).uid = (int) value;
```
icbaker: This is documented in https://datatracker.ietf.org/doc/rfc9559/ with `range:  not 0 (1-18446744073709551615)` which I think means:

1. We should use `long` here (since `18446744073709551615 = 2^64 - 1` - so actually even Java signed `long` kinda isn't large enough if you consider signedness - but better than `int`)
2. We should use `0` as the 'unset' value, not `Format.NO_VALUE` which is `-1` in Java signed int math, which is`18446744073709551615` as a 64-bit unsigned int (i.e. a technically valid value).

icbaker (commented):
I can't work out how to delete the stale comment made against an old copy of the code, sorry for the duplication.

libraries/extractor/src/main/java/androidx/media3/extractor/mkv/MatroskaExtractor.java
tymmesyde: I don't see it there, i followed this spec: https://www.matroska.org/technical/elements.html 
So i guess it's not a RFC standard? Should i remove related logic?

tymmesyde: It is specified in the sources tho: https://github.com/ietf-wg-cellar/matroska-specification/blob/master/ebml_matroska.xml#L1521, probably was not deployed to the webiste

srgsf (approved):
I'm not sure how to resolve threads, everything is correct now.

libraries/extractor/src/main/java/androidx/media3/extractor/mkv/MatroskaExtractor.java
icbaker: ack, thanks for the link.

PR #3145, opened 2026-03-27
Add DTS fallback to MediaCodec

nift4:
Currently, ExoPlayer only has a fallback for passthrough case:

https://github.com/androidx/media/blob/7ce3aa2619b19009e4799319b4dd694a4a7577df/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/AudioCapabilities.java#L397-L400

Mirror the logic for MediaCodec.

Issue: #2487


nift4:
Thanks!

PR #3147, opened 2026-03-28
Support DTS-HD with packets across PES boundary

nift4:
In DTS-HD streams, it can happen that core packet is in one PES and the
extension substream is in another PES. Ensure we can read ahead into next PES
in this case. We need to use the old timestamp as extss and core always belong
together.


FongMi:
Not enough. Please follow the ffmpeg implementation.

nift4:
@FongMi please can you elaborate a bit?

FongMi:
https://github.com/FongMi/media/blob/release-1.10.0-fongmi/libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
https://github.com/FongMi/media/blob/release-1.10.0-fongmi/libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java

This is what I wrote using copilot + sonnet 4.6

nift4:
@FongMi Well, I see that you did a lot of changes and most of them seem to address the same issue that we shouldn't emit core format but extension format instead (I solved it with the PatientTrackOutput wrapper and you added a new state and coreFormatPendingEmit instead). My fix is a bit subtle because it intercepts the format() call and waits until extension header is parsed before sending, allowing to overwrite the format which is cached. But it works and it's readable.

I also see another issue which I didn't solve but you solved it, that we should wait for core frame after seeking instead of sending extension substream frame first - it makes sense and I'll add it to my patch.

Is there any other problem I need to address?

FongMi:
That should be all. I'll test the ISO file again after you finish.

nift4:
@FongMi It should be good now, let me know if it works. Thanks :)

FongMi:
Users have reported that the playback quality of DTS-HD is not as good as my version, and that stuttering occurs. This may be because your Core+EXTSS is sent in segments, while mine is sent all at once.

FongMi:
https://github.com/FongMi/media/commit/6755c3ee4ea55f45ecb58e70c90950948e411a83
Please also implement DTS-MA and DTS-X recognition.
I've switched DtsUtil to your version to reduce merge conflicts.

nift4:
@FongMi 

> Users have reported that the playback quality of DTS-HD is not as good as my version, and that stuttering occurs. This may be because your Core+EXTSS is sent in segments, while mine is sent all at once.

Should be fixed, please check again. Thanks for testing!

> Please also implement DTS-MA and DTS-X recognition.

It's on my TODO list, but I will do it in a seperate PR when this is merged. This one is already quite big.

FongMi:
@nift4 Users say it's still very laggy.

## Core Architectural Difference

### nick — `PatientTrackOutput` double-buffer

```
TS input → PatientTrackOutput internal buffer → flush() → TrackOutput
```

Every frame goes through **two copies**:

```java
// Copy 1: into internal ParsableByteArray
public void sampleData(ParsableByteArray data, int length) {
    this.data.ensureCapacity(this.data.limit() + length);
    ByteBuffer tmp = ByteBuffer.wrap(this.data.getData()); // new object every call
    tmp.put(data.getData(), data.getPosition(), length);
}

// Copy 2: out to the real TrackOutput
public void flush() {
    output.sampleData(data, data.limit());
}
```

### fongmi — zero-copy passthrough

```
TS input → TrackOutput (direct)
```

Instead of buffering, it adds a `STATE_CHECKING_FOR_EXTSS_AFTER_CORE` state: after reading a core frame, it peeks 4 bytes to check whether an EXSS follows, then decides to combine or emit immediately — **no intermediate buffer needed**.

---

## Performance Comparison

| Criterion | fongmi | nick |
|-----------|--------|------|
| Data copies per frame | **1** (direct write) | **2** (buffered) |
| Per-frame heap allocation | None | `ByteBuffer.wrap()` each call |
| `ensureCapacity` reallocation risk | None | Yes (large frames) |
| DTS:X (XLL-X) auto-detection | **Yes** | No |
| Post-seek resync | `skipExtssUntilCore` (precise) | `waitingForResyncAfterSeek` (basic) |
| GC pressure | Low | Higher |

---

## Verdict

**fongmi is faster**, for three reasons:

1. **Zero-copy**: data is written directly to `TrackOutput`, eliminating the `PatientTrackOutput` intermediate buffer and second memcpy. For large DTS-HD MA frames (potentially hundreds of KB each), this is significant.

2. **No per-frame heap allocation**: nick allocates a new `ByteBuffer` wrapper on every `sampleData()` call, increasing GC pressure.

3. **Richer functionality**: fongmi adds XLL-X scanning to auto-detect DTS:X object audio and properly upgrades the track format to `AUDIO_DTS_X`, which nick lacks entirely.

nick's `PatientTrackOutput` is a cleaner abstraction conceptually, but the double-copy cost makes it a poor trade-off for a hot path that processes audio frames continuously.

nift4:
Hi @FongMi, I ran my sample file through both versions of extractors and only difference I saw in result was average bitrate was set on mine. That's set from core header, I imagine it might cause buffer allocation to be too small, so I removed it. Can you re-test?

> No per-frame heap allocation: nick allocates a new ByteBuffer wrapper on every sampleData() call, increasing GC pressure.

Also, I don't think that's the problem causing audible lags, but I changed it to only do that on the first frame (after seek).

rohitjoins:
Hi @nift4,

Thank you for all the work and iterations on this!

The PR in its current form, touches / fixes lots of different issues in extractor. Could we break this down further into smaller issues, focusing one thing at a time?

Suggestion:

- Fix for mime type in TsExtractor
- Format emission issue in TsExtractor
- Fix for MatroskaExtractor
- Fix for Mp4Extractor.

Once done, I'm happy to look at them individually and merge them.

nift4:
Hi @rohitjoins,

> The PR in its current form, touches / fixes lots of different issues in extractor. Could we break this down further into smaller issues, focusing one thing at a time?

Because the other parts depend on mime type detection change in TsExtractor I had it as one branch, otherwise there would be conflicts. But we can instead merge this change first (in this PR, I removed everything except "Fix for mime type in TsExtractor") and I will add the other changes to new PRs once this one is merged.

Also please note that the `format = null;` change is required (and not left over by mistake) in order to have the DTS-HD MA sample ever work with DumpFileAsserts, because without this line, a seek back to 0 would yield different result and test would not work.

Looking forward to review, thanks!

rohitjoins:
> Also please note that the format = null; change is required (and not left over by mistake) in order to have the DTS-HD MA sample ever work with DumpFileAsserts, because without this line, a seek back to 0 would yield different result and test would not work.

Thanks for clarifying! While setting `format = null;` makes the test pass, we unfortunately can't use this in production. 

Clearing the format on `seek()` forces the extractor to re-emit the format. This causes the player to reinitialize the audio decoder on every seek, which leads directly to audio dropouts and the stuttering reported IMO. 

The correct way to make the `DumpFileAsserts` test pass is to avoid the two-step (Core -> HD) format emission entirely. If we wait to check for the extension substream *before* emitting the format (e.g., using a peek state), the extractor will only ever emit the final DTS-HD format. This ensures the initial playback and the `seek()` behavior match naturally, fixing the test without causing decoder resets.

nift4:
Hi @rohitjoins,

> The correct way to make the DumpFileAsserts test pass is to avoid the two-step (Core -> HD) format emission entirely. If we wait to check for the extension substream before emitting the format (e.g., using a peek state), the extractor will only ever emit the final DTS-HD format. This ensures the initial playback and the seek() behavior match naturally, fixing the test without causing decoder resets.

Yes, I did do this, as you can see on my final branch here it's no longer needed to reset the format:
https://github.com/nift4/media/blob/dtshdints-backup/libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java#L124

It's only in this PR because the DTS-HD MA sample (which is added to demonstrate the mime type change) is separated from the fix to emit format only once.

So, I guess it was the wrong order to merge changes. Instead, I will remove mime type change from the branch, and add format-only-once fix here. Then we can merge that first, then the mime type fix, and then the MKV/MP4 related changes; and `format = null;` won't be needed at any point in time.

I pushed this, so now the format-only-fix is in this PR. I opted for this wrapper in order to be minimally invasive, and most importantly, don't duplicate DTS related business logic. FongMi has an approach where there are a bunch of additional states added, which is a different set of tradeoffs for code flow with the same end result. Let me know if mine is OK or if you'd rather have the different state approach.

rohitjoins:
Hi @nift4,

Thanks for explaining your rationale. While I have not compared the performance differences for either of the approaches, I think it would make sense to avoid temporary copies in the memory before being flushed out to the actual output?

> don't duplicate DTS related business logic

Can you please expand more on this? Can we avoid this by refactoring into reusable helper methods?

nift4:
Hi @rohitjoins,

>  I think it would make sense to avoid temporary copies in the memory before being flushed out to the actual output?

Please note there is only _one_ copy, for the very first core frame ,before the format is emitted. After format is emitted to ExtractorOutput, there is no copy anymore and buffer is released (the flush() calls are only used to trigger emitting sampleMetadata, and not sampleData, at this point), as I removed it since FongMi pointed it out. (edit: force pushed to fix a small bug causing useless memory copies)

Given the constraints that:
* consume() in DtsReader cannot peek into future bytes, or go back
* there is no guarantee we get all bytes of a frame or even just header on first invocation of consume()
* we need to read extension substream's header before calling output.format() - and hence before calling output.sampleData()

...I think there is no way to get rid of this one temporary memory copy in any way.

> Can you please expand more on this? Can we avoid this by refactoring into reusable helper methods?

We need to read extension substream's header before calling format(). That involves:
1. first finding core syncword if any (there may be none for DTS Express or DTS-HD MA without backward compatible)
  a. parsing core header
  b. storing core sample data in temporary buffer
3. then finding extension syncword if any (if there is another core one, emit the core sample data and format BEFORE parsing new one)
  a. then parsing its header
  b. call output.format() based on extension header data
  c. only now we can emit core sampleData()

This is the business logic I was talking about. Almost entirety of consume() method code is required for this job. But consume() method is optimized to directly pass data to ExtractorOutput, even though we cannot pass any data to it until we called format().

The difficulty in extracting helper method here is that consume() method may be called multiple times until we finish with any one specific aspect of this parsing. So there is a need to keep state in the class, as is currently done already.

But due to different needs of:
- first frame (cannot call ExtractorOutput methods in any case and must cache data in the class with temporary buffer)
- second frame (based on first frame type, we have to maybe parse header before emitting format, or maybe after, then emit old sample data if any, then proceed as normal)
- subsequent frames (should instantly output data to ExtractorOutput without special logic)

I am not sure how code can be shared in elegant way. I'd be interested if you see any way to do it nicely.

Let me know if that's clear or if you have further questions, Thanks for the feedback!

rohitjoins:
Hi @nift4,

Thank you for the detailed breakdown! Your logic makes complete sense based on the constraint that we need to read the extension substream's header before calling `output.format()`, and hence before calling `output.sampleData()`.

However, in Media3's `TrackOutput` architecture, that constraint doesn't actually exist for `sampleData()`. The `sampleData()` method simply pipes raw bytes into a format-agnostic queue. The format is only strictly required and checked at the very end when `sampleMetadata()` is called to commit the frame.

Because of this, we can stream raw bytes via `sampleData()` before the format is fully resolved. For example, if you look at [H264Reader.java](https://github.com/androidx/media/blob/release/libraries/extractor/src/main/java/androidx/media3/extractor/ts/H264Reader.java#L139) or [H265Reader.java](https://github.com/androidx/media/blob/release/libraries/extractor/src/main/java/androidx/media3/extractor/ts/H265Reader.java#L127), they stream the NAL units via `output.sampleData(data, data.bytesLeft());` as they arrive, parse the SPS headers to build the format later, call `output.format()`, and finally call `sampleMetadata()`.

Using this pattern, we can achieve the zero-copy state machine for DTS-HD without needing any temporary buffer copies, even for the first frame. Inspired by @FongMi's approach, I went ahead and implemented this here: https://github.com/androidx/media/commit/a0c4698bee38328a6353f7090825bc831e74839b.

Could you take a look and let me know if this approach seems fine to you? I can either send my commit directly for internal review, or if you prefer, you can update this PR to include these changes and I can import it from there. Either way, getting this submitted will enable you to add your dependent PRs.

nift4:
Hi @rohitjoins,

> that constraint doesn't actually exist for sampleData()

Duh... I don't know why I thought it did...

> Could you take a look and let me know if this approach seems fine to you? I can either send my commit directly for internal review

Sure, sending it to internal review is fine, looks like it should work just fine and solves the problem.

Thanks for looking into it!

FongMi:
It’s time to use AI to address the shortcomings that Media3 previously had. 🤣

rohitjoins:
Hi @nift4,

This change is merged as [b168816](https://github.com/androidx/media/commit/b1688162525dd590a328bb7daaa39ce264c5a5dc). Please feel free to send other pending PRs dependent on this change.

FongMi:
@rohitjoins My users have reported that your version still lags, but adding some checks will fix it.
You can compare it with my version and see where the differences are, except...xllX Scanning section.
https://github.com/FongMi/media/blob/16022359591a91ebcdb30e73917c94a23b188f97/libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java#L311

FongMi:
**Item 1 (`packetStarted` timeUs guard) — necessary**

DTS-HD EXTSS continuation packets typically carry no PTS, so `pesTimeUs` will be `C.TIME_UNSET`. Without this guard, the upstream code would overwrite a valid timestamp with `C.TIME_UNSET`, causing the `checkState(timeUs != C.TIME_UNSET)` assertion downstream to crash. This fix prevents a DTS-HD playback crash and must be kept.

---

**Item 2 (`packetFinished` isEndOfInput guard) — necessary**

Without `isEndOfInput`, the core format is emitted every time a packet finishes while in `STATE_CHECKING_FOR_EXTSS_AFTER_CORE`. In DTS-HD streams, the EXTSS follows the core in the same PES packet, so `packetFinished` is called only once at the true end. Gating on `isEndOfInput` prevents a premature format emit before the EXTSS has been parsed and the format potentially upgraded to DTS-HD MA or DTS:X.

---

**Item 5 (`setAverageBitrate`) — not necessary for DTS-HD detection**

`DtsHeader.bitrate` is 0 for EXTSS/UHD frames, so `resolvedBitrate` evaluates to `Format.NO_VALUE` — effectively a no-op. Only DTS Core frames carry a meaningful bitrate value. This change has no impact on DTS-HD detection logic; it is merely a cosmetic improvement that fills in bitrate metadata opportunistically and can be dropped without affecting functionality.

Note: within the same diff block there is also a `coreFormatPendingEmit = false` after `output.format()` — that part *is* necessary. Without clearing the flag, the next call to `updateFormatWithDtsHeaderInfo` would re-enter the format-rebuild branch unconditionally and emit a duplicate format event.

nift4:
Hi all,

> This change is merged as [b168816](https://github.com/androidx/media/commit/b1688162525dd590a328bb7daaa39ce264c5a5dc). Please feel free to send other pending PRs dependent on this change.

Thanks @rohitjoins , I've opened https://github.com/androidx/media/pull/3184 which seems to be the next logical step.

> My users have reported that your version still lags, but adding some checks will fix it.
You can compare it with my version and see where the differences are

Strange. I have filtered it and it is a quite small number of changes, I guess it seems safe enough?

```patch
diff --git a/libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java b/libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
index 3959a45735..ebe0de6919 100644
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
@@ -144,7 +144,9 @@ public final class DtsReader implements ElementaryStreamReader {
 
   @Override
   public void packetStarted(long pesTimeUs, @TsPayloadReader.Flags int flags) {
-    timeUs = pesTimeUs;
+    if (pesTimeUs != C.TIME_UNSET) {
+      timeUs = pesTimeUs;
+    }
   }
 
   @Override
@@ -294,8 +296,7 @@ public final class DtsReader implements ElementaryStreamReader {
 
   @Override
   public void packetFinished(boolean isEndOfInput) {
-    if (state == STATE_CHECKING_FOR_EXTSS_AFTER_CORE) {
-      checkNotNull(output);
+    if (isEndOfInput && state == STATE_CHECKING_FOR_EXTSS_AFTER_CORE) {
       if (coreFormatPendingEmit) {
         output.format(checkNotNull(format));
         coreFormatPendingEmit = false;
```

FongMi:
@nift4 Thank you. That's enough.

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
```diff
@@ -294,8 +296,7 @@ public void consume(ParsableByteArray data) throws ParserException {
 
   @Override
   public void packetFinished(boolean isEndOfInput) {
-    if (state == STATE_CHECKING_FOR_EXTSS_AFTER_CORE) {
-      checkNotNull(output);
+    if (isEndOfInput && state == STATE_CHECKING_FOR_EXTSS_AFTER_CORE) {
```
rohitjoins: The tests break with `isEndOfInput` guard because the last sample is never committed in this case.

nift4:
@FongMi Does the stuttering also occur when the timeUs check is added alone? That is, are you sure adding isEndOfInput check is required?

FongMi:
@nift4 I'm having users test it out; I'll get back to you later.

FongMi:
**Fix Description**

`packetFinished` needs to decide whether to commit the Core sample when in `STATE_CHECKING_FOR_EXTSS_AFTER_CORE`. The original upstream version (with `isEndOfInput` guard) only committed on end-of-stream, causing all samples to be lost in Core-only streams. Removing the guard fixed Core-only streams, but for DTS-HD streams `coreSampleSize` was cleared prematurely at a PES packet boundary — the subsequent EXTSS then computed `combinedSize = sampleSize + 0`, producing incorrect frame boundaries and causing stuttering.

A `hasSeenExtss` flag is added and set to `true` on the first EXTSS header parse, allowing `packetFinished` to branch correctly:

---

**Behavior Comparison**

| Scenario | upstream (isEndOfInput guard) | guard removed | This fix (hasSeenExtss) |
|---|---|---|---|
| DTS Core-only, mid-stream frames | ❌ not committed, samples lost | ✅ committed normally | ✅ committed normally |
| DTS Core-only, last frame | ✅ committed | ✅ committed | ✅ committed |
| DTS-HD, Core + EXTSS in same PES | ✅ state already left, branch skipped | ✅ same | ✅ same |
| DTS-HD, Core ends at PES boundary | ✅ not committed, coreSampleSize preserved | ❌ committed, coreSampleSize zeroed → EXTSS combinedSize wrong → stutter | ✅ hasSeenExtss=true, not committed |
| DTS-HD, last frame | ✅ isEndOfInput flush | ✅ flushed normally | ✅ isEndOfInput flush |

image

nift4:
@FongMi Thanks. Can you send a file that stutters (DTS-HD, Core ends at PES boundary) so I can reproduce the issue locally? If you want to send it privately my email is nift4@protonmail.com

FongMi:
Sorry, I don't have the physical ISO file either, because the file is stored on Quark Cloud Drive in China, and you need a Chinese mobile phone number to register for the cloud drive.

FongMi:
I can ask my users for the cloud drive URL, but you may need to find a way to download it yourself.😂

FongMi:
@nift4 https://115cdn.com/s/swf1cnv3hqk?password=1234#

nift4:
Hi, I see, indeed I have no way to download from that site. Is it possible to use another site such as https://transfer.it/ ?

FongMi:
I can't download it either, because I'm Taiwanese.😂

libraries/extractor/src/main/java/androidx/media3/extractor/ts/DtsReader.java
nift4: Hi @rohitjoins,

Thanks for the heads up and sorry for the delay.

I thought a bit about it and what FongMi posted about checking whether extension substream was previously found seemed like a valid workaround, but it bugged me that `isEndOfInput` wasn't working as I was confident that would be correct, so I discarded the workaround and investigated more.

It turned out to be a bug in PesReader that `isEndOfInput` wasn't set if the packet duration was known. Fixing this bug for some reason adds new samples to various files unrelated to DTS-HD, I would need to defer to your expertise if this fixed a valid bug preventing emission of samples in a large amount of readers or if I just broke something.

Thank you!

rohitjoins: Thank you for digging into this! 

> It turned out to be a bug in PesReader that isEndOfInput wasn't set if the packet duration was known.

Spot on! However, the workaround of changing the state transition logic to `if (payloadSize == 0 && padding > 0)` introduces a latency regression for other codecs. By requiring `padding > 0`, `PesReader` is forced to stay in `STATE_READING_BODY` until the next PES packet arrives. For PES-aligned metadata streams like ID3 and SCTE-35 (which rely exclusively on `packetFinished` to emit their samples), this means a tag won't be emitted until the next tag arrives, which could be minutes later.

Instead of delaying the transition out of `STATE_READING_BODY`, the correct fix should allow `PesReader` to process the extractor's synthesized EOF PUSI even if it has cleanly returned to `STATE_READING_HEADER`. Wdyt?

Update `consume()` to:

```
if ((flags & FLAG_PAYLOAD_UNIT_START_INDICATOR) != 0) {
      boolean isEndOfInput = (data.limit() == 0);
      switch (state) {
        case STATE_FINDING_HEADER:
        case STATE_READING_HEADER:
          if (isEndOfInput) {
            reader.packetFinished(true);
          }
          break;
// ...
```

And allowing the EOF flush when idle in `canConsumeSynthesizedEmptyPusi()`:

```
 return (state == STATE_READING_HEADER || (state == STATE_READING_BODY && payloadSize == C.LENGTH_UNSET))
        && !(isModeHls && reader instanceof H262Reader)
        && headerParsed;
```

Also can you please re-run `TsPlaybackTest` to update the dump files in it?

Can you split the PR into two? One focusing on making the `PesReader` fix and the other on the DTS optimization.

nift4: Hi @rohitjoins,

> Instead of delaying the transition out of STATE_READING_BODY, the correct fix should allow PesReader to process the extractor's synthesized EOF PUSI even if it has cleanly returned to STATE_READING_HEADER. Wdyt?

That was my first approach, but I wasn't sure if there's any Reader that relies on packetFinished() being called exactly once per packet (if it's in STATE_READING_HEADER, we would in practice call it twice, first with isEndOfInput=false and then with isEndOfInput=true once we know there is EOF).

While writing this it occured to me that it would perhaps be best to split packetFinished(boolean) into a packetFinished() and a endOfInputReached(), as that would allow each codec to make the decision at the place it needs to without confusingly calling packetFinished() twice nor increasing latency when not needed. Do you think that would be a good idea? If so, I'll open a new PR with that done and playback dump files updated.

rohitjoins: Sounds good. Please proceed with your suggestion. Thank you!

nift4: Uploaded PR https://github.com/androidx/media/pull/3206 - please review, thanks!

rohitjoins: Thank you! Can you also please update this PR to just have the DTS optimizations? I'll ensure I get it merged in correct order to avoid the failing tests.

nift4: Sure, I force pushed it. If applied in the wrong order it would cause compilation failure. Thanks for taking care of it!

rohitjoins:
Hi @nift4,

I have added a test with single sample across PES boundary. When run on `main` branch, it would output two samples in the dump file as expected and the reason for stuttering. 

It runs correctly to produce a single sample at this PR with the fixes.

nift4:
Hi @rohitjoins,

I still don't have an affected sample myself (and I have to admit I'm a bit curious how you made this one), but it's good to know we can verify PR fixes it! Thanks again. It looks good to merge from my side.

rohitjoins:
I first tried using `ffmpeg` with small PES payload sizes, but it prefers to keep complete audio frames together. So I used a custom Python script to take a raw DTS frame and manually mux it into TS packets, forcing a new PES boundary right in the middle of the frame!

nift4:
Thanks!

PR #3151, opened 2026-04-01
Fix overlapping subtitles with negative line numbers

### LINKED ISSUE #2237: Overlapping subtitles with VTT file.
It is pretty common for anime subtitles to have times that overlap. This leads to ExoPlayer showing the text of the subtitle on top of the previous subtitle, making them unreadable.  Is there a way to fix that?

This is an small sample of the vtt file that shows the issue (whole file attached):

```
01:56.260 --> 01:59.400
<i>They are known as "trash games."</i>

01:59.720 --> 02:02.260
Not Trash Games

01:59.720 --> 02:06.730
Semi-Redeemable Trash Games

02:00.100 --> 02:04.230
<i>Although most people 
rightfully despise trash games,</i>

02:01.550 --> 02:06.730
Never Play Again

02:04.230 --> 02:06.730
<i>there are those who gleefully seek them out.</i>
```

[eng-2 vtt.txt](https://github.com/user-attachments/files/19234197/eng-2.vtt.txt)

johnpc:
## Summary

Fixes overlapping subtitle rendering when multiple WebVTT cues have overlapping timestamps and are assigned negative line numbers.

## Problem

When multiple WebVTT cues overlap in time, `WebvttSubtitle.getCues()` correctly assigns them negative line numbers (-1, -2, -3, etc.) to stack them from the bottom. However, `SubtitlePainter` calculates each cue's vertical position using only `firstLineHeight` (the height of a single text line), not accounting for the actual multi-line height of preceding cues.

This causes cues to visually overlap when:
1. Multiple cues are displayed simultaneously
2. Any of the cues wrap to multiple lines

This is particularly common in anime subtitles where overlapping timestamps are used for effects like showing both dialogue and on-screen text.

## Solution

1. Added `getLastDrawnTextHeight()` to `SubtitlePainter` to expose the actual rendered height of a text cue
2. Modified `CanvasSubtitleOutput.dispatchDraw()` to track cumulative height of bottom-stacked cues (those with negative `LINE_TYPE_NUMBER`) and adjust the drawing boundary for subsequent cues

## Testing

Added unit tests for:
- `SubtitlePainterTest`: Tests for `getLastDrawnTextHeight()` with various cue types
- `CanvasSubtitleOutputTest`: Tests for rendering overlapping cues without visual overlap

Fixes #2237

### DIFF
--- a/libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
+++ b/libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
@@ -52,8 +52,8 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
   /** Ratio of inner padding to font size. */
   private static final float INNER_PADDING_RATIO = 0.125f;
 
-  /** The bottom Y position of the last drawn text cue, or 0 if no text cue was drawn. */
-  private int lastDrawnTextBottom;
+  /** The height of the last drawn cue, or 0 if no cue was drawn. */
+  private int lastDrawnCueHeight;
 
   // Styled dimensions.
   private final float outlineWidth;
@@ -221,26 +221,25 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
     if (isTextCue) {
       checkNotNull(cueText);
       setupTextLayout();
-      lastDrawnTextBottom = textLayout != null ? textLayout.getHeight() : 0;
+      lastDrawnCueHeight = textLayout != null ? textLayout.getHeight() : 0;
     } else {
       checkNotNull(cueBitmap);
       setupBitmapLayout();
-      lastDrawnTextBottom = 0;
+      lastDrawnCueHeight = bitmapRect != null ? bitmapRect.height() : 0;
     }
     drawLayout(canvas, isTextCue);
   }
 
   /**
-   * Returns the bottom Y position of the last drawn text cue relative to the parent bounds.
+   * Returns the height of the last drawn cue.
    *
-   * <p>This can be used to stack multiple cues without overlap. For bottom-stacked cues (negative
-   * line numbers), this represents how far up from the bottom the cue extends. For top-stacked cues
-   * (positive line numbers), this represents how far down from the top the cue extends.
+   * <p>This can be used to stack multiple cues without overlap by adjusting the drawing bounds for
+   * subsequent cues by this amount.
    *
-   * @return The bottom Y position of the last drawn text cue, or 0 if no text cue was drawn.
+   * @return The height of the last drawn cue in pixels, or 0 if no cue was drawn.
    */
-  public int getLastDrawnTextBottom() {
-    return lastDrawnTextBottom;
+  public int getLastDrawnCueHeight() {
+    return lastDrawnCueHeight;
   }
 
   @RequiresNonNull("cueText")
--- a/libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
+++ b/libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
@@ -52,6 +52,9 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
   /** Ratio of inner padding to font size. */
   private static final float INNER_PADDING_RATIO = 0.125f;
 
+  /** The bottom Y position of the last drawn text cue, or 0 if no text cue was drawn. */
+  private int lastDrawnTextBottom;
+
   // Styled dimensions.
   private final float outlineWidth;
   private final float shadowRadius;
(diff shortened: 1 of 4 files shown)


libraries/ui/src/test/java/androidx/media3/ui/CanvasSubtitleOutputTest.java
```diff
@@ -0,0 +1,244 @@
+/*
+ * Copyright 2025 The Android Open Source Project
```
icbaker, line 2: Please update this to the current year (same in the next file too)

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -218,13 +221,26 @@ public void draw(
...
   }
 
+  /**
+   * Returns the height of the last drawn text cue in pixels.
+   *
+   * <p>This can be used to stack multiple cues without overlap.
+   *
+   * @return The height of the last drawn text cue, or 0 if no text cue was drawn.
+   */
+  public int getLastDrawnTextHeight() {
```
icbaker: please clarify in both the method name & docs whether this returns the top or bottom of the text (this may be more fiddly if you update the PR to handle the possibly-overlapping positive line numbers case too)

icbaker:
Thanks for the PR - is there a reason you are specifically handling negative line numbers? I think the same problem can occur with positive line numbers too.

Would you be able to handle the positive line numbers case in this PR as well? I think a list of cues could have cues with both positive & negative line numbers (some at the top, some at the bottom of the screen, at the same time) - so handling both cases separately would probably be best. I think it makes sense to handle both cases at the same time so that we don't accidentally over-fit the solution to only one of the two cases.

Please also rebase the PR to target the `main` branch as that's where we merge contributions.

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -218,13 +221,28 @@ public void draw(
     if (isTextCue) {
       checkNotNull(cueText);
       setupTextLayout();
+      lastDrawnTextBottom = textLayout != null ? textLayout.getHeight() : 0;
     } else {
       checkNotNull(cueBitmap);
       setupBitmapLayout();
+      lastDrawnTextBottom = 0;
```
icbaker: Is there a reason this logic is text-specific?

I believe bitmap cues can also have `line`-based position

johnpc:
@icbaker I've addressed all your feedback - updated copyright years, renamed the method to `getLastDrawnTextBottom()` with clearer docs, added support for positive line numbers, and rebased onto main. Ready for another look when you have time!

johnpc:
Gentle ping on this, not sure what review timeline expectations there are for external contributions. Just want to make sure it's on your radar!

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
```diff
@@ -218,13 +221,28 @@ public void draw(
...
+  /**
+   * Returns the bottom Y position of the last drawn text cue relative to the parent bounds.
+   *
+   * <p>This can be used to stack multiple cues without overlap. For bottom-stacked cues (negative
+   * line numbers), this represents how far up from the bottom the cue extends. For top-stacked cues
+   * (positive line numbers), this represents how far down from the top the cue extends.
+   *
+   * @return The bottom Y position of the last drawn text cue, or 0 if no text cue was drawn.
+   */
+  public int getLastDrawnTextBottom() {
```
icbaker: this method is documented to return a position relative to the parent bounds, but I think it actually returns the height of a single text layout (assigned on L224 above) which is not the same thing.

It's quite possible the logic is correct, just the naming/documentation is confusing - or maybe I'm misunderstanding something.

johnpc:
I've tested this by building into the jellyfin-androidtv app and confirming that it fixes my original subtitle issue https://github.com/jellyfin/jellyfin-androidtv/pull/4522

It does work. 

Screenshot 2026-04-18 at 12 43 22 PM

Worth noting that in my case, the subtitles are still out-of-order, but I think that's fine since it's sort of undefined behavior in an srt anyway, and it matches what happens in jellyfin-web

johnpc:
pinging on this again @icbaker , what do we need to get this over the finish line

johnpc:
Hey @icbaker just checking in again. I believe I've addressed your feedback, please let me know if there's anything else I can do for this

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
icbaker, line 364: I understand how this change works for several cues with the same `line` value

But what about cues with **different** `line` values?

Previously, if we had two simultaneous cues with `line = -1` and `line = -2` they would be positioned:
1. With the bottom of the text box at the bottom of the viewport
2. With the bottom of the text box one line above the bottom of the viewport

With your change, the second cue with `line = -2` will also be positioned against a 'smaller' viewport, so it will be moved up more than needed: By both the height of the first cue **and** two extra lines. I don't think this is correct.

And what if the cues actually arrive in the opposite order (and assuming both cues fit on a single line)? If we first process the `line = -2` cue then we will place this cue 'correctly' but if we then process the `line = -1` cue with the 'shortened' viewport, then the cue that should slot **underneath** will actually be displayed above.

(I think all the reasoning above also applies to positive line values)

johnpc:
@icbaker I've updated the approach to track cumulative offsets per line value using a map, so only cues sharing the same line number get stacked. Cues with different line numbers (e.g., -1 and -2) are now positioned independently by the existing line-number math without any viewport adjustment.

libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java
johnpc: resolved in https://github.com/androidx/media/pull/3151/commits/f00f1883bf9329186d7f4df117e330136fc062b2

johnpc:
gentle ping, would be great to get this overlapping subtitle issue resolved!

johnpc:
gentle ping, would be great to get this overlapping subtitle issue resolved! cc @icbaker

johnpc:
Hello - I am curious about what my expectation should be as an external contributor. Is there any mechanism for communication about this PR beyond the PR correspondence? I have been blocked without feedback since May, and I'd really like to see this improvement land.

PR #3156, opened 2026-04-02
Update DefaultAudioSink.java

AndresBC-123:
Hello:
In my application, I implemented AudioProcessors that allow processing audio encoded in all encoding types supported by Android, including Float. Why can't they be implemented if the audio encoding type is Float? With this change, AudioProcessors can be used even if the audio is of type Float.


rohitjoins:
Thanks for the PR! I'm passing this over to @ivanbuper to evaluate, as this may also need changes to our default audio processors.

AndresBC-123:
Hello @rohitjoins @ivanbuper 
This is a friendly reminder to merge before the next version release.

Thank you.

PR #3184, opened 2026-04-20
Enable accurate mime type assignment for DTS-HD in TS

nift4:
Read more of the DTS-HD header in order to find out extension substream type, to get correct mime type which is relevant for buffer size decision logic (as DTS Express has way lower maximum bit rate than DTS-HD).

Issue: #2487
Issue: #3147

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
@@ -15,6 +15,7 @@
  */
 package androidx.media3.extractor;
 
+import static com.google.common.base.Preconditions.checkNotNull;
 import static java.lang.annotation.ElementType.TYPE_USE;
 import static java.lang.annotation.RetentionPolicy.SOURCE;
 
@@ -45,7 +46,7 @@ public final class DtsUtil {
   /** Information parsed from a DTS frame header. */
   public static final class DtsHeader {
     /** The mime type of the DTS bitstream. */
-    public final @DtsAudioMimeType String mimeType;
+    @Nullable public final @DtsAudioMimeType String mimeType;
 
     /** The audio sampling rate in Hertz, or {@link C#RATE_UNSET_INT} if unknown. */
     public final int sampleRate;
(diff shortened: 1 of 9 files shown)


libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
nift4, line 1: Hi @rohitjoins,

I don't understand why this is needed: according to my reading of Table 7-5/7-6/7-7 you don't need to check this and can just assume embedded stereo/6ch flags as false if static fields aren't present.

rohitjoins: Thank you for looking into these commits and providing feedback. 

I took a closer look at the ETSI TS 102 114 V1.6.1 specification to verify this, and unfortunately, assuming those flags are `false` when `bStaticFieldsPresent == 0` violates the spec and will cause a parser desync.

According to Section 7.5.2 (Page 98), the definition of `bStaticFieldsPresent` states: `If the bStaticFieldsPresent is false, the metadata fields that are static over the duration of an encoded stream are omitted from the extension substream header.`

Because they are `static over the duration of an encoded stream`, they are simply omitted to save bandwidth, they do not become false? They retain whatever value was established in the initial keyframe.

Since `DtsUtil.parseDtsHdHeader` is a stateless utility, it has no memory of the stream's true `bEmbeddedStereoFlag`. If we blindly assume the flag is false on an intermediate frame, but the stream actually has embedded stereo, our parser will fail to skip those 8 bits and instantly desync. By the time it tries to read the coding mode for the MIME type, it will be reading misaligned garbage data.

Therefore, because we cannot reliably parse the Dynamic Metadata statelessly, safely confining the parsing to if (`staticFieldsPresent`) and using the `sampleMimeType` fallback in `DtsReader` is required to prevent crashes on intermediate frames?

Please let me know if my interpretation of the spec is correct, and if I should send this PR at its current state for internal review. Thanks!

nift4: Thanks for explaining, you're right :)

FongMi: @nift4 Is your next PR going to modify the DTS format detection of MKV files?

nift4: Yes

nift4:
Thanks @rohitjoins, I've uploaded #3226 as next step

Issue #2263, opened 2025-03-24
MediaSource reuse and BaseMediaSource releaseSource.


Tolriq:
To prepare for some requested features for my application I now have to manage multiple instance of ExoPlayer 1.6.0.

There's cases where I will prepare a player for a `mediaSource` but then have to release that player. 
In that case when I try to reuse the `mediaSource` in another player there's an error (See below).
This is because the `mediaSource` was already prepared and have a `looper` associated that will be different from the new player.

The release of the player is done via:

```
clearMediaItems() 
stop() 
playbackParameters = PlaybackParameters.DEFAULT 
removeListener(customEventListener)
release()
```

I tried to retrace how / if the `releaseSource` is called in that case but code is hard to follow.

Is it normal that the looper is not reset in that case? Is there something I can do to ensure that it is ?

PS: 
1) Each player have it's own looper for performance reasons, ANRs and not using MainThread.
2) The media source preparation is costly so it would be best if I could reuse it.

Error:
```
ExoAnalyticsLogger.e@62: ExoPlayerImplInternal - Playback error (Ask Gemini)
androidx.media3.exoplayer.ExoPlaybackException: Unexpected runtime error
	at androidx.media3.exoplayer.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:849)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loopOnce(Looper.java:248)
	at android.os.Looper.loop(Looper.java:338)
	at android.os.HandlerThread.run(HandlerThread.java:85)
Caused by: java.lang.IllegalArgumentException
	at androidx.media3.common.util.Assertions.checkArgument(Assertions.java:40)
	at androidx.media3.exoplayer.source.BaseMediaSource.prepareSource(BaseMediaSource.java:248)
	at androidx.media3.exoplayer.source.CompositeMediaSource.prepareChildSource(CompositeMediaSource.java:123)
	at androidx.media3.exoplayer.source.WrappingMediaSource.prepareChildSource(WrappingMediaSource.java:240)
	at androidx.media3.exoplayer.source.MaskingMediaSource.createPeriod(MaskingMediaSource.java:127)
	at androidx.media3.exoplayer.MediaSourceList.createPeriod(MediaSourceList.java:338)
	at androidx.media3.exoplayer.MediaPeriodHolder.createMediaPeriod(MediaPeriodHolder.java:498)
[...]

PR #3188, opened 2026-04-26
Add descriptive error message for looper mismatch in BaseMediaSource

NoNews:
Reusing a cached `MediaSource` across players with different playback threads throws a message-less `IllegalArgumentException`, silently swallowed into error code 1004. 

**Impact at Reddit Android App**
We experienced ~2M errors/day. It only showed up in one specific edge case, so we had nothing to go on but telemetry with 1004 error code — and the empty exception made it much harder to pin down. (We spent a week debugging this).

We're currently experimenting with a single shared looper and seeing the problem addressed.

See [#2263](https://github.com/androidx/media/issues/2263)

Thank you!

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -53,6 +53,9 @@
         `DefaultBandwidthMeter.Builder.setInitialBitrateSupplier()`.
     *   Enable dynamic scheduling by default. `ExoPlayer` playback work loop
         will run dynamically as opposed to on a static interval.
+    *   Add a descriptive error message when `MediaSource` is prepared on a
+        different looper than it was originally bound to, making this previously
+        silent 1004 failure easier to diagnose.
 *   CompositionPlayer:
 *   Transformer:
     *   Fix an issue where `ExportResult.fileSizeBytes` may be over-reported.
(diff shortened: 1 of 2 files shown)


icbaker:
Did the stack trace you had not point to the line of the `checkArgument` call (and so the cause would be obvious by looking at the source)?

I think in general we write exception checks & messages on the assumption that the reader will cross-reference with the relevant source code for additional context. This is partly to keep APK bloat in check by avoiding long strings on every edge-case exception check.

NoNews:
Hi @icbaker and thank you for the feedback!

> Did the stack trace you had not point to the line of the `checkArgument` call (and so the cause would be obvious by looking at the source)?

Yes, once we were able to reproduce it _(which was the hardest part as only tiny fraction of users were affected)_, we were able to see the trace.
 
> I think in general we write exception checks & messages on the assumption that the reader will cross-reference with the relevant source code for additional context. This is partly to keep APK bloat in check by avoiding long strings on every edge-case exception check.

Got it! Do you think we can still have some distinction and use a shorter message, to reduce the impact? Like:

```java
heckArgument(
      this.looper == null || this.looper == looper,
      "Wrong looper used");
```

 The reason we ask: without a message, our error reporting only showed `Unexpected error` with code `1004` — nothing pointing to a looper or a MediaSource. We report the error code together with the exception message, so even a short string like`Wrong looper used` would have appeared in our telemetry right away and told us where to look, saving a week of investigation time

icbaker:
Thanks for the extra context.

I chatted about this with others on the team, and we are a bit surprised that your error telemetry completely drops stack traces and only reports exception messages. We think it's unreasonable to expect every exception message in the library to be unique (in order to identify the root cause from just the message), and it would make more sense for error-reporting telemetry to include the stack trace. I assume this would be obfuscated, but could of course be unobfuscated with the appropriate mapping file.

Given this, we don't plan to add additional exception messages routinely in the codebase, and that includes this PR - so I'm going to close this.

PR #3191, opened 2026-04-30
Fix: Support empty ILST entry

Tolriq:
Some files can have empty ILST that will prevent playback due to unbounded attempt to read them.

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/MetadataUtil.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/MetadataUtil.java
@@ -152,6 +152,10 @@ import com.google.common.collect.ImmutableList;
     int type = ilst.readInt();
     int typeTopByte = (type >> 24) & 0xFF;
     try {
+      if (ilst.bytesLeft() == 0) {
+        Log.d(TAG, "Skipped empty metadata entry: " + Mp4Box.getBoxTypeString(type));
+        return null;
+      }
       if (typeTopByte == TYPE_TOP_BYTE_COPYRIGHT || typeTopByte == TYPE_TOP_BYTE_REPLACEMENT) {
         int shortType = type & 0x00FFFFFF;
         if (shortType == SHORT_TYPE_COMMENT) {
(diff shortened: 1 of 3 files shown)


libraries/extractor/src/main/java/androidx/media3/extractor/mp4/MetadataUtil.java
```diff
@@ -152,6 +152,10 @@ public static Metadata.Entry parseIlstElement(ParsableByteArray ilst) {
     int type = ilst.readInt();
     int typeTopByte = (type >> 24) & 0xFF;
     try {
+      if (ilst.bytesLeft() == 0) {
```
rohitjoins: Thank you for sending this PR!

Isn't the check using ilst.bytesLeft() == 0 only checking if this is the very last element in the entire buffer? If an empty entry occurs in the middle of the list, this check would be bypassed.

Should we instead check if the current element is empty?

Tolriq: I might have been unlucky and all the breaking files from the users where on the last one so did the wrong check yes. If there's a better way to test that case it should be used. Just need to prevent a `readInt()` later that will trigger out of range errors or read the wrong data.

rohitjoins:

libraries/extractor/src/main/java/androidx/media3/extractor/mp4/MetadataUtil.java
```diff
@@ -151,6 +151,10 @@ public static Metadata.Entry parseIlstElement(ParsableByteArray ilst) {
     int endPosition = position + ilst.readInt();
     int type = ilst.readInt();
     int typeTopByte = (type >> 24) & 0xFF;
+    if (endPosition - ilst.getPosition() < Mp4Box.HEADER_SIZE) {
+      Log.d(TAG, "Skipped empty metadata entry: " + Mp4Box.getBoxTypeString(type));
+      return null;
```
Tolriq, line 154: Not sure of the impact but it was in the try block to ensure that the finally correctly update the endposition.

rohitjoins: Thanks for pointing that out, moved inside the `try` block.

PR #3193, opened 2026-05-02
Add a disc subtitle field to MediaMetadata 

tautropfli:
For ID3 we read the [`TSST` frame][1] which is part of ID3v2.4.

Vorbis comments don't have an ['official'][2] equivalent.
However, there is the de-facto standard `DISCSUBTITLE` field which is
used by taggers such as [Picard][3] and libraries such as [taglib][4].

I also noticed that the disc number is only read from vorbis comments but not from ID3 tags.
Let me know if I should submit that commit as its own PR.

[1]: https://id3.org/id3v2.4.0-frames
[2]: https://xiph.org/vorbis/doc/v-comment.html
[3]: https://picard-docs.musicbrainz.org/en/latest/appendices/tag_mapping.html#disc-subtitle
[4]: https://taglib.org/api/p_propertymapping.html#:~:text=DISCSUBTITLE

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -132,6 +132,8 @@
         framework issue existing prior to API 31.
 *   Text:
 *   Metadata:
+    *   Add `MediaMetadata.discSubtitle` field and parse it from ID3v2.4 `TSST`
+        and Vorbis `DISCSUBTITLE` data.
 *   Image:
     *   Fix issue in scrubbing mode where image updates would only take effect
         when the user "stops scrubbing"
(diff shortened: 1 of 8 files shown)


icbaker:
> I also noticed that the disc number is only read from vorbis comments but not from ID3 tags.
> Let me know if I should submit that commit as its own PR.

yes please put this in a separate PR

tautropfli:
> yes please put this in a separate PR

gotcha, extracted into https://github.com/androidx/media/pull/3202 😃

icbaker:

tautropfli:
Awesome, thanks 🚀

PR #3198, opened 2026-05-05
Handle gapless MP3 Xing/Info durations

### LINKED ISSUE #3183: MP3 Gapless issue
### Version

Media3 main branch

### More version details

Can reproduce on main branch

All

N/A

Not tested

### Reproduction steps

Play the 2 songs in a playlist.

### Expected result

Should be fully gapless (and was previously)

### Actual result

There's a small glitch between the files.

Logs shows 
```
2026-04-03 22:03:52.921 Verbose ExoPlayer  Mp3Extractor - Data size mismatch between stream (33218028) and Xing frame (33217900), using Xing value.
2026-04-03 22:04:16.129 Verbose ExoPlayer  Mp3Extractor - Data size mismatch between stream (9268883) and Xing frame (9268755), using Xing value.
```

Some of the recent changes around that probably trigger this now.

### Media

[joni 0.zip](https://github.com/user-attachments/files/26896551/joni.0.zip)

### Bug Report

Tolriq:
LAME Xing/Info headers can include encoder delay and padding. Mp3Extractor already propagates those values into Format so decoded audio is rendered gaplessly, but the SeekMap duration was still computed from the untrimmed MPEG frame count. For CBR Info files this can expose a source duration that is longer than the samples that will actually be played, which can break gapless transitions.

This change also updates average bitrate to be based on the gapless duration.

Fixes #3183.

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/mp3/ConstantBitrateSeeker.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/mp3/ConstantBitrateSeeker.java
@@ -37,7 +37,8 @@ import androidx.media3.extractor.SeekPoint;
    * Constructs an instance.
    *
    * <p>The duration exposed from {@link #getDurationUs()} is computed from {@code inputLength} and
-   * the frame bitrate, or is {@link C#TIME_UNSET} if {@code inputLength} is unknown.
+   * the bitrate of {@code mpegAudioHeader}, or is {@link C#TIME_UNSET} if {@code inputLength} is
+   * unknown.
    *
    * @param inputLength The length of the stream in bytes, or {@link C#LENGTH_UNSET} if unknown.
    * @param firstFramePosition The position of the first frame in the stream.
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/mp3/ConstantBitrateSeeker.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/mp3/ConstantBitrateSeeker.java
@@ -18,6 +18,8 @@ package androidx.media3.extractor.mp3;
 import androidx.media3.common.C;
 import androidx.media3.extractor.ConstantBitrateSeekMap;
 import androidx.media3.extractor.MpegAudioUtil;
+import androidx.media3.extractor.SeekMap.SeekPoints;
+import androidx.media3.extractor.SeekPoint;
 
 /**
  * MP3 seeker that doesn't rely on metadata and seeks assuming the source has a constant bitrate.
@@ -28,10 +30,14 @@ import androidx.media3.extractor.MpegAudioUtil;
   private final int bitrate;
   private final int frameSize;
   private final boolean allowSeeksIfLengthUnknown;
+  private final long durationUs;
   private final long dataEndPosition;
 
   /**
    * Constructs an instance.
+   *
+   * <p>The duration exposed from {@link #getDurationUs()} is computed from {@code inputLength} and
+   * the frame bitrate, or is {@link C#TIME_UNSET} if {@code inputLength} is unknown.
    *
    * @param inputLength The length of the stream in bytes, or {@link C#LENGTH_UNSET} if unknown.
    * @param firstFramePosition The position of the first frame in the stream.
@@ -53,23 +59,30 @@ import androidx.media3.extractor.MpegAudioUtil;
         mpegAudioHeader.bitrate,
         mpegAudioHeader.frameSize,
         allowSeeksIfLengthUnknown,
-        /* isEstimated= */ true);
+        /* durationUs= */ C.TIME_UNSET);
   }
 
-  /** See {@link ConstantBitrateSeekMap#ConstantBitrateSeekMap(long, long, int, int, boolean)}. */
+  /**
+   * See {@link ConstantBitrateSeekMap#ConstantBitrateSeekMap(long, long, int, int, boolean)}. Uses
+   * {@code durationUs} as the duration exposed from {@link #getDurationUs()}, or computes the
+   * duration from {@code inputLength} and {@code bitrate} if {@code durationUs} is {@link
+   * C#TIME_UNSET}.
+   */
   public ConstantBitrateSeeker(
       long inputLength,
       long firstFramePosition,
       int bitrate,
       int frameSize,
-      boolean allowSeeksIfLengthUnknown) {
+      boolean allowSeeksIfLengthUnknown,
+      long durationUs) {
     this(
         inputLength,
         firstFramePosition,
         bitrate,
         frameSize,
         allowSeeksIfLengthUnknown,
-        /* isEstimated= */ true);
+        /* isEstimated= */ true,
+        durationUs);
   }
 
   private ConstantBitrateSeeker(
@@ -78,7 +91,8 @@ import androidx.media3.extractor.MpegAudioUtil;
       int bitrate,
       int frameSize,
       boolean allowSeeksIfLengthUnknown,
-      boolean isEstimated) {
+      boolean isEstimated,
+      long durationUs) {
     super(
         inputLength,
         firstFramePosition,
@@ -88,14 +102,26 @@ import androidx.media3.extractor.MpegAudioUtil;
         isEstimated);
     this.firstFramePosition = firstFramePosition;
     this.bitrate = bitrate;
-    this.frameSize = frameSize;
+    this.frameSize = frameSize == C.LENGTH_UNSET ? 1 : frameSize;
     this.allowSeeksIfLengthUnknown = allowSeeksIfLengthUnknown;
+    this.durationUs = durationUs;
     dataEndPosition = inputLength != C.LENGTH_UNSET ? inputLength : C.INDEX_UNSET;
   }
 
   @Override
   public long getTimeUs(long position) {
     return getTimeUsAtPosition(position);
+  }
+
+  @Override
+  public SeekPoints getSeekPoints(long timeUs) {
+    if (durationUs != C.TIME_UNSET && timeUs >= durationUs && dataEndPosition != C.INDEX_UNSET) {
+      long finalFramePosition = Math.max(firstFramePosition, dataEndPosition - frameSize);
+      long frameDurationUs = getTimeUsAtPosition(firstFramePosition + frameSize);
+      return new SeekPoints(
+          new SeekPoint(Math.max(0, durationUs - frameDurationUs), finalFramePosition));
+    }
+    return super.getSeekPoints(timeUs);
   }
 
   @Override
@@ -106,6 +132,11 @@ import androidx.media3.extractor.MpegAudioUtil;
   @Override
   public long getDataEndPosition() {
     return dataEndPosition;
+  }
+
+  @Override
+  public long getDurationUs() {
+    return durationUs != C.TIME_UNSET ? durationUs : super.getDurationUs();
   }
 
   @Override
@@ -120,6 +151,7 @@ import androidx.media3.extractor.MpegAudioUtil;
         bitrate,
         frameSize,
         allowSeeksIfLengthUnknown,
-        /* isEstimated= */ false);
+        /* isEstimated= */ false,
+        durationUs);
   }
 }
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/mp3/ConstantBitrateSeeker.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/mp3/ConstantBitrateSeeker.java
@@ -26,10 +26,6 @@ import androidx.media3.extractor.SeekPoint;
  */
 /* package */ final class ConstantBitrateSeeker extends ConstantBitrateSeekMap implements Seeker {
 
-  private final long firstFramePosition;
-  private final int bitrate;
-  private final int frameSize;
-  private final boolean allowSeeksIfLengthUnknown;
   private final long durationUs;
   private final long dataEndPosition;
 
@@ -101,10 +97,6 @@ import androidx.media3.extractor.SeekPoint;
         frameSize,
         allowSeeksIfLengthUnknown,
         isEstimated);
-    this.firstFramePosition = firstFramePosition;
-    this.bitrate = bitrate;
-    this.frameSize = frameSize == C.LENGTH_UNSET ? 1 : frameSize;
-    this.allowSeeksIfLengthUnknown = allowSeeksIfLengthUnknown;
     this.durationUs = durationUs;
     dataEndPosition = inputLength != C.LENGTH_UNSET ? inputLength : C.INDEX_UNSET;
   }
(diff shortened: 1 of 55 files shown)


icbaker:
GitHub isn't letting me add review comments for some reason, so here's a quick bit of initial feedback

----

The part of this PR related to the `LAME_TO_DECODED_PCM_TRIM_OFFSET_SAMPLES` constant & its usages seem unrelated to the duration vs raw duration split. Please can you send this as a separate PR?

---

In the `bear-vbr-xing-header-no-toc.mp3.cbr-seeking-always.0.dump` dump file, I'd expect `getPosition(DURATION)` to return `timeUs=DURATION` (concretely: `getPosition(2783979) = [[timeUs=2783979, position=38396]]`). But it seems `timeUs` is still returning the 'raw' duration (same for many other dump files).

Tolriq:
Yeah sorry I did not fully verify that part, I completely underestimated the complete mess that is mp3 seeking and gapless before going into this.

I've moved the lame part in another PR, but used AI to rebase it on main to not depend on this one. It's probably best to not really look at the second one deeply until this is one is done or not.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Seeker.java
```diff
@@ -51,6 +51,14 @@
    */
   int getAverageBitrate();
 
+  /**
+   * Returns the raw duration before subtracting encoder delay/padding, or {@link C#TIME_UNSET} if
+   * not known.
+   */
+  default long getRawDurationUs() {
```
icbaker: i'm wondering if we can avoid introducing this method (and therefore avoid having to distinguish between these concepts of 'duration' and 'raw duration').

It looks like it's used in two ways:

1. To calculate a bitrate in `Mp3Extractor` as part of the 'enhanced CBR' logic
2. In the implementation of `ConstantBitrateSeeker.getSeekPoints` to ensure the final byte position is reported correctly.

For (1), I see that `Seeker` already has a `getAverageBitrate()` method, which **should** fit for the use-case (based on its documentation of `Returns the average bitrate (usually derived from the duration and length of the file)`) - except it seems that `XingSeeker` and `VbriSeeker` implement this incorrectly by directly returning the bitrate of the MPEG frame that contains the seek metadata.

For (2) I think you can use `getDataEndPosition()` instead?

If this sounds right, would you be up for:

1. Sending a PR to fix the bug in the `getAverageBitrate()` impls in `XingSeeker` and `VbriSeeker`
2. Rebasing this PR on top once that's merged, and then removing `getRawDurationUs` and using `getAverageBitrate()` and `getDataEndPosition()` instead.

Tolriq: I can use Codex to split things and move things yes if that's not a problem for you. Way less time consuming.

Also Codex: `Caveat: getDataEndPosition() is documented as immediately after audio data, so using it
  directly may be one byte/frame past the desired seek point. The implementation should likely use the last valid seek position derived from data end, for example dataEndPosition - normalizedFrameSize, while returning timeUs = durationUs.`  we start to be a little outside of my confort zone so I don't really have value on the decisions, just did tests and debug to try to reach solutions. I now mostly rely on your expertise.

icbaker: Ah yeah, that's a good point about `getDataEndPosition()` - but `ConstantBitrateSeeker` already has a `frameSize` field so this calculation should be easy.

I think the time of this seek point should **also** be adjusted to be `durationUs - frameDurationUs` - since if you seek there, it will play the last frame, which has non-zero duration.

Tolriq:
Rebased after the merge of the other one.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/IndexSeeker.java
```diff
@@ -33,14 +33,26 @@
...
+      long durationUs, long dataStartPosition, long dataEndPosition, int averageBitrate) {
     this.indexSeekMap =
         new IndexSeekMap(
             /* positions= */ new long[] {dataStartPosition},
             /* timesUs= */ new long[] {0L},
             durationUs);
     this.dataStartPosition = dataStartPosition;
     this.dataEndPosition = dataEndPosition;
-    this.averageBitrate = computeAverageBitrate(dataEndPosition - dataStartPosition, durationUs);
+    this.averageBitrate =
```
icbaker: is this diff related to the gapless fix or should it be in a separate PR?

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3Extractor.java
```diff
@@ -533,23 +535,27 @@ private Seeker computeSeeker(ExtractorInput input) throws IOException {
...
+                inputLength,
+                dataStart,
+                averageBitrate,
+                C.LENGTH_UNSET,
+                /* allowSeeksIfLengthUnknown= */ false,
+                resultSeeker.getDurationUs());
+      } else {
+        resultSeeker =
+            getConstantBitrateSeeker(
+                input, (flags & FLAG_ENABLE_CONSTANT_BITRATE_SEEKING_ALWAYS) != 0);
```
icbaker: it is unfortunate to add another branch to this logic (which is already very branch-y), especially since this is a duplicate of L564.

Please can you refactor this to avoid this duplication?

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3Extractor.java
```diff
@@ -670,15 +676,11 @@ private Seeker getConstantBitrateSeeker(
...
-    int averageBitrate =
-        Ints.checkedCast(
-            Util.scaleLargeValue(
-                audioLength,
-                C.BITS_PER_BYTE * C.MICROS_PER_SECOND,
-                durationUs,
-                RoundingMode.HALF_UP));
+    // See also https://github.com/androidx/media/issues/1376.
+    int averageBitrate = computeAverageBitrate(audioLength, infoFrame.computeEncodedDurationUs());
+    if (averageBitrate == C.RATE_UNSET_INT) {
```
icbaker, line 684: how can this happen? `durationUs == C.TIME_UNSET` already results in `return null` above and `audioLength` is assigned to a non-unset value on either L669 or L672 above.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/IndexSeeker.java
Tolriq: This come up during the rebase because of https://github.com/androidx/media/commit/48836cb9febaf0327c6863f12ef3ecb836112dc8 that was changed in the middle.

This is Codex answer when I asked confirmation about that change (Due to previous tests no more passing after that commit) :

```
• You’re right to question it. The failures were not from the new tests; they were existing Mp3ExtractorTest dump comparisons for FLAG_ENABLE_INDEX_SEEKING.

  What happened:

  - Our gapless change makes XingSeeker.getDurationUs() return the gapless duration.
  - We separately made Xing/Info average bitrate use encoded duration, so normal Xing paths keep the old bitrate.
  - But with FLAG_ENABLE_INDEX_SEEKING, Mp3Extractor wraps the unseekable XingSeeker in an IndexSeeker.
  - IndexSeeker was recomputing averageBitrate from its constructor durationUs, which is now the gapless duration.
  - That changed dump output from 108719 to 109656.

  So my IndexSeeker change was: when wrapping an existing seeker, pass through the existing average bitrate instead of recomputing it from the gapless duration. The added IndexSeekerTest only documents that behavior.

  I would not fix this by updating dumps. Updating dumps would accept a real behavior change in Format.averageBitrate for index-seeking MP3s, which contradicts the goal of “gapless durations while preserving average bitrate calculations.”

  That said, if we want the upstream diff to avoid touching IndexSeeker, a narrower alternative is possible: keep IndexSeeker unchanged and have Mp3Extractor remember the pre-wrap average bitrate for Format.Builder. But that leaves IndexSeeker.getAverageBitrate() internally
  inconsistent, so I think the current small IndexSeeker constructor overload is the cleaner fix.

```

It's answer convinced me it was better than updating the average, but it's maybe wrong.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3Extractor.java
Tolriq: Just an habit of defensive coding as the function can return it, did not check all the possible cases.
Can remove if wanted and safe.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/ConstantBitrateSeeker.java
```diff
@@ -69,7 +72,28 @@ public ConstantBitrateSeeker(
         bitrate,
         frameSize,
         allowSeeksIfLengthUnknown,
-        /* isEstimated= */ true);
+        /* durationUs= */ C.TIME_UNSET);
+  }
+
+  /**
+   * See {@link ConstantBitrateSeekMap#ConstantBitrateSeekMap(long, long, int, int, boolean)}. Uses
+   * {@code durationUs} as the duration exposed from {@link #getDurationUs()}.
```
icbaker: this comment is imprecise, because it doesn't make it clear that if `durationUs = TIME_UNSET` is passed then `getDurationUs()` falls back to returning a bitrate-derived duration.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/ConstantBitrateSeeker.java
icbaker: this constructor is now unused so imo can be removed.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3Extractor.java
icbaker: thanks, i think you can remove this then

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/IndexSeeker.java
icbaker: >   - We separately made Xing/Info average bitrate use encoded duration, so normal Xing paths keep the old bitrate.

Ah right, gotcha - i guess this comes down to a question of what 'average bitrate' means when considering delay & padding.

The `Format.averageBitrate` field is only 'directly' used by ExoPlayer during adaptive playback (e.g. HLS or DASH) for track selection decisions - so for MP3 files it is pretty much just for information (so it doesn't matter if we change it - within reason).

I think there's a pretty strong argument that if we are reducing the reported duration we should increase the reported bitrate - since the audio data still occupies the same number of bytes (the delay & padding are necessary parts of encoding the audio), but the audible duration is less.

Which I think means you might be able to simplify this PR a bit by always calculating bitrate from the 'gapless' duration.

Tolriq:
So I rebased and run a Codex review and the added back the defensive check due to :

```
• Yes, this can happen. The earlier checks only prove:

  - durationUs is valid, from frameCount.
  - There is some stream length source, from Info.dataSize or fallbackStreamLength.

  They do not prove audioLength > 0.

  Concrete reachable case:

  audioLength = infoFrame.dataSize - infoFrame.header.frameSize;

  XingFrame.parse() accepts dataSize directly from the Info header without validating it against the MPEG frame size. So a malformed Info header can have dataSize == header.frameSize or smaller. Then durationUs is still valid because it comes from frameCount, but
  computeAverageBitrate(audioLength, durationUs) returns C.RATE_UNSET_INT.

  That is exactly what our regression test creates:

  writeBigEndianInt(fileBytes, infoTagOffset + 12, infoFrameHeader.frameSize);

  So the guard is not hypothetical.

  One possible improvement is to make the code more explicit:

  if (audioLength <= 0) {
    return null;
  }
  int averageBitrate = computeAverageBitrate(audioLength, durationUs);
  if (averageBitrate == C.RATE_UNSET_INT) {
    return null;
  }

  But the current guard is still necessary because computeAverageBitrate can also reject overflow/out-of-range values.
```

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3Extractor.java
icbaker: ack, ok to leave this guard then.

Tolriq:
@icbaker there's maybe something wrong I got at least one mail with comments but can't find them on Github.

I think everything was addressed in the last update. But Github is a mess.

icbaker:
I think this comment thread is still open/unresolved: https://github.com/androidx/media/pull/3198/changes#r3339773432

(I agree it's easy to lose comments on github)

Tolriq:
Okay should be addressed and all the other comment I could find are fixed too hopefully.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3Extractor.java
```diff
@@ -389,6 +390,28 @@ private long computeTimeUs(long samplesRead) {
...
+   * is present, this trims the encoder delay and padding so EOF finalization does not replace an
+   * initially gapless Xing/Info duration with the longer encoded duration.
+   */
+  private long computeFinalIndexSeekerDurationUs(long samplesRead) {
+    long durationUs = computeTimeUs(samplesRead);
+    if (!gaplessInfoHolder.hasGaplessInfo()) {
+      return durationUs;
+    }
+    long finalGaplessSampleIndex =
+        Util.durationUsToSampleCount(durationUs, synchronizedHeader.sampleRate)
```
icbaker: is it necessary go go from a sample count (`samplesRead`) to a duration then back to a sample count here?

I wonder about instead:

```java
private long computeFinalIndexSeekerDurationUs(long samplesRead) {
  long gaplessSampleCount =
      gaplessInfoHolder.hasGaplessInfo()
          ? samplesRead - gaplessInfoHolder.encoderDelay - gaplessInfoHolder.encoderPadding - 1
          : samplesRead;
  return gaplessSampleCount >= 0 ? computeTimeUs(gaplessSampleCount) : C.TIME_UNSET;
}
```

All the tests seem to pass with this, so I'll upload it after the rebase i mention below - but please flag if you think it's incorrect.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3Extractor.java
```diff
@@ -533,24 +553,26 @@ private Seeker computeSeeker(ExtractorInput input) throws IOException {
...
+                  /* allowSeeksIfLengthUnknown= */ false,
+                  durationUs);
+        }
+      }
+    }
+
+    if (shouldFallbackToConstantBitrateSeeking(resultSeeker)) {
       // Either we found no seek or VBR info, so we must assume the file is CBR (even without the
       // flag(s) being set), or an 'enable CBR seeking flag' is set and we found some seek info, but
       // not enough to do 'enhanced' CBR seeking with. In either case, we fall back to CBR seeking
```
icbaker: While reviewing this PR I realised this comment is imprecise (and this whole `computeSeeker` method is a bit of a mess). I'm going to send a clean-up change internally first, then once it's submitted I'll rebase this on top.

libraries/extractor/src/main/java/androidx/media3/extractor/mp3/Mp3Extractor.java
Tolriq: Looks good, but maybe keep the var name ? With the -1 it's more an index than a count.

icbaker: Done, thanks

icbaker:

PR #3202, opened 2026-05-06
Read the ID3v2.3 TPOS frame

tautropfli:
Split off from https://github.com/androidx/media/pull/3193.

The [`TPOS` frame][1] can be either a number, or two numbers separated with a '/' similar to how track numbers in the `TRCK` frame:

> The 'Part of a set' frame is a numeric string that describes which
> part of a set the audio came from. This frame is used if the source
> described in the "TALB" frame is divided into several mediums, e.g. a
> double CD. The value may be extended with a "/" character and a
> numeric string containing the total number of parts in the set. E.g.
> "1/2".

[1]: https://id3.org/d3v2.3.0

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -111,6 +111,7 @@
 *   Metadata:
     *   Add `MediaMetadata.discSubtitle` field and parse it from ID3v2.4 `TSST`
         and Vorbis `DISCSUBTITLE` data.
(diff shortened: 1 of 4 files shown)


icbaker:
Please rebase this on the `main` branch

tautropfli:
Oh dang I forgot to target `main` instead of `release` on this one. Fixed ✅

Issue #3113, opened 2026-03-10
CEA-608 Closed Captions end up in the same media group


adadukin-dolby:
### Version

Media3 1.9.2

### Devices that reproduce the issue

Android Emulator (sdk_gphone64_arm64) with Android 16
Emulator version (36.4.9-14788078)

### Reproducible in the demo app?

Yes

### Reproduction steps

1. Update [`media.exolist.json`](https://github.com/androidx/media/blob/53602878bcc6d150a969ec8bdf7755f7d2e5c730/demos/main/src/main/assets/media.exolist.json) asset under `demos/main/src/main/asset` with the entry below

```json
      {
        "name": "DASH with CEA-608 tracks",
        "uri": "https://livesim2.dashif.org/vod/testpic_2s/cea608.mpd"
      },
```
2. Put a breakpoint (or log `track groups`) in [`PlayerControlView#gatherSupportedTrackInfosOfType`](https://github.com/androidx/media/blob/53602878bcc6d150a969ec8bdf7755f7d2e5c730/libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java#L1470)

3. Build and run sample app. Find newly added entry and play it

4. While the caption list seem correct from the UI side, the debugger should produce something similar to the screenshot below

<img width="944" height="288" alt="Image" src="https://github.com/user-attachments/assets/9f368df5-fdbc-4cea-bcda-76465c667762" />

### Expected result

According to the [`TrackGroup`](https://github.com/androidx/media/blob/release/libraries/common/src/main/java/androidx/media3/common/TrackGroup.java#L46) documentation,

>  Note that audio tracks whose languages differ are not grouped, because content in different
> languages is not considered to be the same. Conversely, audio tracks in the same language that
> only differ in properties such as bitrate, sampling rate, channel count and so on can be grouped.
 > **This also applies to text tracks.**

It seems that the text tracks in different languages should end up in different track groups. The issue has also been discussed in [a separate bug report](https://github.com/androidx/media/issues/887#issuecomment-1856118787) which confirms our suspicion that this is not an intended behaviour.

> I agree that merging the english and spanish tracks into a single track group seems wrong

Sample app seem to handle it correctly though, because it iterates over all available `Format`s across all track groups [in here](https://github.com/androidx/media/blob/53602878bcc6d150a969ec8bdf7755f7d2e5c730/libraries/ui/src/main/java/androidx/media3/ui/PlayerControlView.java#L1470-L1486).

### Actual result

[...]

PR #3205, opened 2026-05-07
Separate CEA captions into different groups based on language

adadukin-dolby:
Problem: `CEA-608` and `CEA-708` captions end up in the same group for `DASH`.
Proposal: split all text captions into different `TrackGroup`s, similar to `HlsMediaPeriod` implementation 

More context available at: https://github.com/androidx/media/issues/3113

### DIFF
--- a/libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/DashMediaPeriod.java
+++ b/libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/DashMediaPeriod.java
@@ -748,7 +748,7 @@ import java.util.regex.Pattern;
       primaryGroupClosedCaptionTrackFormats[i] =
           getClosedCaptionTrackFormats(adaptationSets, groupedAdaptationSetIndices[i]);
       if (primaryGroupClosedCaptionTrackFormats[i].length != 0) {
-        numEmbeddedTrackGroups++;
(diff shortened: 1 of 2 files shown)


tonihei:
Thanks for the fix - unfortunately the PR is based on the release branch, so I can't merge it as it is. Could you open a new one based on the `main` branch? Also happy to simply apply the same changes internally if you prefer.

(Second note: the branch is not accessible and will results in an 'evil' merge. Previous PRs from Dolby had the same problem and I was told that's no issue, so I assume the same is true here).

adadukin-dolby:
> Thanks for the fix - unfortunately the PR is based on the release branch, so I can't merge it as it is. Could you open a new one based on the `main` branch? Also happy to simply apply the same changes internally if you prefer.
> 
> (Second note: the branch is not accessible and will results in an 'evil' merge. Previous PRs from Dolby had the same problem and I was told that's no issue, so I assume the same is true here).

Hey @tonihei,
Pointed the branch to `main`

adadukin-dolby:
@tonihei ptal

adadukin-dolby:
Hey @tonihei, did you have an opportunity to take a look at the changes?

tonihei:
Yes, it just needs to be submitted internally

adadukin-dolby:
> Yes, it just needs to be submitted internally

Ty for reply, let me know if anything is pending from our side

PR #3206, opened 2026-05-07
Fix issue where end of stream wasn't signaled if last pes had length field set

nift4:
In PesReader, endOfInput wasn't set to true if last pes had length field set. Originally, ExoPlayer was always dropping last frame in pes, it was fixed in https://github.com/androidx/media/pull/419 for pes without length and this fixes it for pes with length set. Thus, the dumps changed because last sample is now correctly parsed by ExoPlayer for both TsExtractor and PsExtractor.

Split packetFinished(boolean) into two methods because:
- If pes length is unset, we only know packet is finished once next one starts or stream ends, so we know packet end and stream end at the same time. (This is what was handled so far)
- If pes length is set, we may know packet end far before stream ends. Waiting for next pes to know whether to set isEndOfInput would introduce latency issues for ID3 or SCTE-35, but when we don't wait we can't deliver end of stream event later.

Thus split off a endOfInputReached() method to deliver end of stream event later, and every codec can override the method that best fits without latency issues.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -80,6 +80,9 @@
     *   MP4: Prevent infinite loops and out-of-bounds reads when parsing empty
         `ilst` metadata tag items
         ([#3191](https://github.com/androidx/media/pull/3191)).
+    *   MPEG-TS: Ensure the last frame is rendered for streams where the last
+        PES packet has a known length
+        ([#3206](https://github.com/androidx/media/pull/3206)).
 *   Inspector:
 *   Audio:
     *   Update `MediaCodecAudioRenderer` to extract the spatial channelMask from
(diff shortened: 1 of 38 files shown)


FongMi:
Is there anything else that needs adjusting? If not, I'd like to merge and test it.

rohitjoins:
@FongMi,

Please feel free to test this PR. If you also want to test the stuttering you reported in https://github.com/androidx/media/pull/3147, you can also rebase that PR on top of this to verify it works smoothly now?

FongMi:
Okay, thank you.

nift4:
I'm really grateful for you being so responsive and helpful while working on getting this chain merged, thank you!

Issue #2361, opened 2025-04-22
Add Matroska zlib support


optix2000:
Before filing a feature request:
-----------------------
- Search existing open issues, specifically with the label ‘enhancement’:
  https://github.com/androidx/media/labels/enhancement
- For ExoPlayer-related feature requests, please also check for existing feature
  requests on the ExoPlayer tracker:
  https://github.com/google/ExoPlayer/labels/enhancement
- Search existing pull requests:
    - On this tracker: https://github.com/androidx/media/pulls,
    - On the ExoPlayer tracker: https://github.com/google/ExoPlayer/pulls

When filing a feature request:
-----------------------
Replace the content in the sections below.

### [REQUIRED] Use case description
Today, the Matroska extractor [only supports header stripping](https://github.com/androidx/media/blob/839c4a90f2ab36e48be73e1b5e907f3283dce72e/libraries/extractor/src/main/java/androidx/media3/extractor/mkv/MatroskaExtractor.java#L1012-L1018) (method 3). However, zlib compressed tracks (method 0) exist in the wild, not uncommonly for subtitles (eg PGS). (Google `"Muxing mode : zlib"`)

### Proposed solution
Add support for zlib compressed tracks in the Matroska extractor.
[Spec](https://www.matroska.org/technical/elements.html#:~:text=other%20methods.%0A0%20%2D-,zlib,-%2C%0A1%20%2D%20bzlib)

### Alternatives considered
N/A

nielsvanvelzen:
I've gotten reports of people using zlib compressed subtitles in mkv files with my app so this feature would be nice to have for us. I see the issue is still in "needs triage" so I'd like to ask, is this something the media3 team is interested in working on? Or would pull requests for this functionality be accepted?

rohitjoins:
Hi @nielsvanvelzen,

We're not planning to work on this in the near term, but we would accept a high quality PR implementing this functionality in Matroska extractor. Please let us know if you have any more questions.

inorichi:
I created a PR that adds this functionality, please let me know if it's good enough or needs more changes.

nielsvanvelzen:
Hi @rohitjoins, a pull request implementing this was made a while back by @inorichi. To me it seems like it is of good quality and it is really a small change (ignoring all the test data ofc). I'd appreciate it if this could be looked at as a lot of my users are affected by this. (My app, Jellyfin, is a self-hosted media server. So people bring their own media files). If there's anything I can do to help out, let me know.

[...]

PR #3211, opened 2026-05-08
Add support for Matroska zlib compressed subtitle tracks 

inorichi:
This PR solves #2361

The first commit implements the changes to support zlib compressed subtitles in SUBRIP, ASS, SSA and VTT.
PGS and VOBSUB already support zlib compressed subtitles (I verified this by creating a sample file with zlib compression) so no changes were needed.
The second commit adds the samples and new tests and dumps for those samples.

The implementation looks for the master element `ID_CONTENT_COMPRESSION` and when present, enables a boolean flag in the current track. 

Later, when a subtitle block is processed, a new `inflatedSubtitleSample` is used to hold the uncompressed data, and follows a similar approach to `PgsParser` and `VobsubParser` by using the method `Util.maybeInflate` and finally copies the result to the real `subtitleSample`. 

The method `reset` could not be used because the input already contains the codec prefix at the beginning, so I used `System.arraycopy` after the prefix. Let me know if you prefer another approach.

The samples were created from the existing ones with the closest tool that created them in the first place, so I've used both ffmpeg and mkvmerge. When running diffs on the dumps of the original mkv and derived mkv with zlib compression, the only changes are in the seekMap positions, so I don't think we can reuse other dumps as it's done for other tests, but I'm no expert on the seekMap so let me know if it's possible.

I've also mixed usages of the null terminated samples because zlib data may contain a null byte, so the decompression must be done before looking for a null byte.

The commands used were either of the two:
```
mkvmerge -o sample_output.mkv --compression 2:zlib sample_input.mkv
ffmpeg -i sample_input.mkv -map 0 -c copy -metadata:s:s:0 compression=zlib sample_output.mkv
```


ethaldeman:
Any chance of this making it into 1.11.0? I am suffering downstream issues that this pull request would solve.

tzi4:
Could this PR be reviewed for an upcoming release please? In my setup MKV files with explicit ContentCompAlgo=0 on PGS tracks cause playback failure in Jellyfin Android TV even with subtitles disabled which triggers software transcoding that overwhelms my Raspberry Pi server.

Issue #3219, opened 2026-05-13
Subtitles not shown for Segmented Webvtt after seeking


jruesga:
### Version

Media3 main branch

### More version details

When playing a stream with segmented webvtt, no subtitles are displayed after seeking. 

Attached to the ticket are a list of test streams. The former 2 are not segmented webvtt (mp4 container and single text/vtt file) and subtitles are displayed correctly after seek. The latter 2 are segmented webvtt (single and multiperiod text/vtt files). No subtitles displayed after seek.

### Devices that reproduce the issue

Any device

### Reproducible in the demo app?

Yes

### Reproduction steps

1. Play a segmented webvtt (single or multiperiod) asset
2. Enable webvtt subtitles (select English)
3. Seek before a cue (eg 1:38 [`This blade has a dark past.` at 1:47] or 10:16 [`Scales?` at 10:21]).
4. Observe subtitles displayed by the player

### Expected result

Subtitles after seeking should be displayed

### Actual result

Not subtitle or wrong subtitles displayed

### Media

```json
{
    "name": "WebVTT",
    "samples": [
      {
        "name": "WebVTT (MP4)",
        "uri": "https://storage.googleapis.com/shaka-demo-assets/sintel-mp4-wvtt/dash.mpd"
      },
      {
        "name": "WebVTT",
        "uri": "https://raw.githubusercontent.com/jruesga/test-streams/refs/heads/main/webvtt/dash.mpd"
      },
      {
        "name": "WebVTT (segmented)",
        "uri": "https://raw.githubusercontent.com/jruesga/test-streams/refs/heads/main/webvtt/dash-segmented.mpd"
      },
      {
        "name": "WebVTT (segmented multiperiod)",
        "uri": "https://raw.githubusercontent.com/jruesga/test-streams/refs/heads/main/webvtt/dash-segmented-multiperiod.mpd"
      }
    ]
```

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

jruesga:
A potential fix has been provided in https://github.com/androidx/media/pull/3220

icbaker:
I don't believe segmented WebVTT files that aren't encapsulated into MP4 containers are supported by the [DASH-IF IOP](https://dashif.org/docs/DASH-IF-IOP-v4.3.pdf), in particular this part:

> If segmented subtitles are needed, such as for live sources, segments shall be packaged in ISOBMFF containers

You either need to:
[...]

PR #3220, opened 2026-05-13
Fix segmented webvtt parser status after seek

jruesga:
When playing a stream with segmented webvtt, no subtitles are displayed after seeking. This PR tries to reset the status to allow to reinitialized the webvtt parser state, so it can display segmented webvtt after seeking.

See https://github.com/androidx/media/issues/3219


icbaker:
We don't plan to support this since it's beyond the DASH-IF IOP guidelines, as discussed in https://github.com/androidx/media/issues/3219.

PR #3226, opened 2026-05-14
Support DTS Express in Matroska

nift4:
FFmpeg outputs DTS Express files with "DTS" codec ID, so we need to add it to the detection. That however is free because we must do the detection anyway for DTS-HD, and we can even reuse all the code needed for TS logic.

This is also a building block for a future commit detecting DTS-HD MA, and the util method is useful for a future commit for the MP4 extractor.


FongMi:
@nift4 I hope that future versions of DTS-MA can be changed like this.
image
image

nift4:
@FongMi Yes, I'll do it once DTS Express is taken care of for MP4 and MKV. The reason is that changing this to a new MIME type will need some care in order to not break passthrough for Android TV devices, so I think it will be easier to review if it's seperated from the changes that are for DTS Express.

FongMi:
@nift4 thanks,
My user has already tested it, and he said the amplifier can display the formats correctly, including DTS:X.

nift4:
Thanks again @rohitjoins, I've uploaded the last shard of my original PR: https://github.com/androidx/media/pull/3229
After that I will probably look into a separate mime type for DTS-HD MA.

nift4:
@FongMi I've now uploaded the DTS-HD MA PR: #3234

PR #3229, opened 2026-05-15
Apply DTS-HD and DTS Express detection to MP4

nift4:
MP4 sadly doesn't store information required to understand which DTS format is which either (similar to MKV), so use the helper to detect format in MP4 container.

Issue: https://github.com/androidx/media/pull/3147


nift4:
Thank you very much for your patience and efforts to get the shards of the PR merged.

FYI, I will need some time to prepare the DTS-HD MA PR (which wasn't part of the original PR) as some unexpected work came up.

nift4:
I've now uploaded PR #3234 which implements everything except object-audio (DTS:X) detection.

PR #3231, opened 2026-05-22
Fix MP3 metadata average bitrates

Tolriq:
This is the first part of the split from https://github.com/androidx/media/pull/3198

@icbaker

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -102,6 +102,7 @@
     *   MPEG-TS: Ensure the last frame is rendered for streams where the last
         PES packet has a known length
         ([#3206](https://github.com/androidx/media/pull/3206)).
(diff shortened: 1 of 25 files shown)


icbaker:

JannetGen:
This looks like it could be a parsing edge case with how Media3 handles malformed or non-standard container metadata (especially with fields like track numbers, missing album art, or ID3 versions). When the framework hits a tag it doesn't expect, it can trip up the timeline or session state tracking.
As a quick diagnostic step to see if it’s the file structure causing the issue rather than your player implementation, you can upload the media file to a web tool like [https://editmp3tags.com/](https://editmp3tags.com/) to rewrite the ID3 tags cleanly and see if the crash or parsing error persists with a sanitized file. If rewriting the tags fixes the playback, it’s likely an extraction bug in the current release wrapper.

PR #3234, opened 2026-05-23
Support DTS-HD MA profiles

nift4:
Android supports a seperate AudioFormat.ENCODING_DTS_HD_MA and a MediaCodec
mime type with ;profile=dtsma since SDK 34, but ExoPlayer wasn't using this
even though the platform docs explicitly say to use the new encoding for
DTS-HD MA. Implement detection of DTS-HD MA and add plumbing to use this
constant where available, and fall back to previous behavior of using
ENCODING_DTS_HD otherwise for backwards compatibility with devices that
weren't updated to support the new constant/MIME type.

DTS-HD MA exists in two versions, in multi-layer version where fallback to
DTS decoders is valid and in a single-layer version where it is not. Add
detection for this and change fallbacks accordingly.

Issue: https://github.com/androidx/media/issues/2487


nift4:
Hi @rohitjoins, should I fix the conflict or did you already import this PR / will fix it yourself? Thanks.

rohitjoins:
I've imported this PR, will handle these merge conflicts, thanks for checking!

nift4:
Hi @rohitjoins, is there any update on merging this PR? Thanks!

PR #3236, opened 2026-05-24
Add Format.selectionPriority and parse from HLS SCORE

AndyWangLYN:
### Summary

  Adds `Format.selectionPriority` and populates it from the HLS `SCORE` attribute on `EXT-X-STREAM-INF` /
  `EXT-X-I-FRAME-STREAM-INF`.

  **Extraction-only — no selection behavior changes in this PR**

  ### Spec
  
  [draft-pantos-hls-rfc8216bis-22](https://www.ietf.org/archive/id/draft-pantos-hls-rfc8216bis-22.txt) §4.4.6.2
  (SCORE on EXT-X-STREAM-INF), §4.4.6.3 (inherits to EXT-X-I-FRAME-STREAM-INF), §6.3.1 (client behavior).

  ### Testing

  - [x] `./gradlew :lib-common:test`
  - [x] `./gradlew :lib-exoplayer-hls:connectedAndroidTest`
  - [x] `git diff -U0 main... | google-java-format-diff.py -p1 -i`


libraries/common/src/main/java/androidx/media3/common/Format.java
```diff
@@ -988,6 +1004,16 @@ public Format build() {
   /** Track role flags. */
   public final @C.RoleFlags int roleFlags;
 
+  /**
+   * A relative preference among formats in the same {@link TrackGroup}. A higher value indicates a
+   * more preferred format, or {@link #NO_VALUE} if unset.
+   *
+   * <p>Populated from the HLS {@code SCORE} attribute on {@code EXT-X-STREAM-INF} and {@code
+   * EXT-X-I-FRAME-STREAM-INF}, and from the DASH {@code @selectionPriority} attribute on {@code
```
tianyif: Do you plan to have a similar PR for DASH as well?

tianyif (commented):
Thanks for the contribution! The PR looks good to me. I just left one comment on the plan of same support for DASH, and you don't have to address it in this PR.

I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review.

libraries/common/src/main/java/androidx/media3/common/Format.java
AndyWangLYN: Thank you for reviewing! Yes I do have a local draft PR for DASH support (hence the comment about dash in code as well), which I plan to raise later today since this PR looks good to you. Separated them for easy review. Will tag this PR when I raise the dash one. Thanks!

tianyif: Sounds good. To keep the current PR clean for HLS, I will alter this javadoc a bit without mentioning DASH, you can add it back in the next PR for DASH.

Issue #3177, opened 2026-04-15
DASH - Not supporting Variable length description in sgpd


psc-globant:
### Version

Media3 main branch

### More version details

Hello,

We are integrating DRM Key Rotation in our DASH streams, and we see Media3 is not able to start playing the stream, throwing the following related error:

"Variable length description in sgpd found (unsupported)"

Could you tell us if that kind of support is planned in the near future?

Thank you
Pablo

### Devices that reproduce the issue

STB AndroidTV - Android 14

### Reproducible in the demo app?

Yes

### Reproduction steps

Try to play the following URL:

https://test-kr-sport.akamaized.net/live/linea04-une/test-key-rotation.isml/index.mpd

### Expected result

playback starts 

### Actual result

playback does not start, with the mentioned error

### Media

https://test-kr-sport.akamaized.net/live/linea04-une/test-key-rotation.isml/index.mpd

### Bug Report

- [x] You will email the zip file produced by `adb bugreport` to android-media-github@google.com after filing this issue.

marcbaechinger:
Thanks for reporting. I can confirm the test stream above can't be played. I can't get the error message you report but I see that the DRM is not supported on the device I try:

```
10:19:18.635  D  tracks [eventTime=0.76, mediaPos=32.64, window=0, period=0
10:19:18.635  D    group [ id=2
10:19:18.635  D      [ ] Track:0, id=video=300000, mimeType=video/avc, container=video/mp4, bitrate=300000, codecs=avc1.4D4015, drm=[widevine,cenc], res=426x240, fps=25.0, roleFlags=[main], supported=NO_UNSUPPORTED_DRM
10:19:18.635  D      [ ] Track:1, id=video=700000, mimeType=video/avc, container=video/mp4, bitrate=700000, codecs=avc1.4D401E, drm=[widevine,cenc], res=640x360, fps=25.0, roleFlags=[main], supported=NO_UNSUPPORTED_DRM
10:19:18.635  D      [ ] Track:2, id=video=1000000, mimeType=video/avc, container=video/mp4, bitrate=1000000, codecs=avc1.64001E, drm=[widevine,cenc], res=764x432, fps=25.0, roleFlags=[main], supported=NO_UNSUPPORTED_DRM
[...]

PR #3243, opened 2026-05-27
Support variable length `sgpd` boxes in fMPEG

dimitry-unified-streaming:
For `sgpd` boxes with version >= 1, `the default_length` field is zero if the length of the following sample group entries is variable.

Since the `FragmentedMp4Extractor` only supports one entry, skip the `description_length` field if applicable (e.g. when version >= 1 and `default_length` == 0), instead of throwing an error.

For parsing the following CencSampleEncryptionInformationGroupEntry, this should make no difference.

Issue: androidx/media#3177


libraries/extractor/src/main/java/androidx/media3/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -1532,18 +1532,17 @@ private static void parseSampleGroups(
     sgpd.setPosition(Mp4Box.HEADER_SIZE);
     int sgpdVersion = BoxParser.parseFullBoxVersion(sgpd.readInt());
     sgpd.skipBytes(4); // grouping_type == seig.
-    if (sgpdVersion == 1) {
-      if (sgpd.readUnsignedInt() == 0) {
-        throw ParserException.createForUnsupportedContainerFeature(
-            "Variable length description in sgpd found (unsupported)");
-      }
-    } else if (sgpdVersion >= 2) {
+    long default_length = sgpdVersion >= 1 ? sgpd.readUnsignedInt() : 0;
```
icbaker: i don't think the `>= 1` check here is correct - I think it should be `== 1`.

Looking at section 8.9.3.2 of ISO 14496-12:2015 it seems `default_length` is present at v1 **only**, and the **same 4 bytes** are `default_sample_description_index` on v2+:

```
if (version==1) { unsigned int(32) default_length; }
if (version>=2) {
  unsigned int(32) default_sample_description_index;
}
```

This also matches the previous code.

Same below on the condition gating the `description_length` skip.

dimitry-unified-streaming: I have the 2022 and 2024 DIS versions of ISO 14496-12 here, and those both have the following:

```
aligned(8) class SampleGroupDescriptionBox ()
  extends FullBox('sgpd', version, flags){
  unsigned int(32) grouping_type;
  if (version>=1) { unsigned int(32) default_length; }
  if (version>=2) {
    unsigned int(32) default_group_description_index;
  }
  unsigned int(32) entry_count;
  int i;
  for (i = 1 ; i <= entry_count ; i++){
    if (version>=1) {
     if (default_length==0) {
       unsigned int(32) description_length;
      }
    }
    SampleGroupDescriptionEntry (grouping_type);
    // an instance of a class derived from SampleGroupDescriptionEntry
    // that is appropriate and permitted for the media type
  }
}
```

I.e. `default_length` exists for `version>=1`, and `default_group_description_index` exists for `version>=2`. This is also how we write `sgpd` boxes in our own software.

icbaker: Interesting, thanks - aren't these two versions of the spec binary-incompatible with each other? My understanding was that this was generally avoided in these sort of specs. Was this an (unacknowledged?) bug in the 2012 to 2020 versions of the spec?

Given this is quite a fiddly change, and there's a risk that it's quite hard to find this PR comment thread in future when someone is going through the blame layer, I'm going to send a change that pretty much **only** changes `== 1` to `>= 1` with an explanation - and then we can rebase this PR on top - hope that's OK.

libraries/extractor/src/test/java/androidx/media3/extractor/mp4/FragmentedMp4ExtractorNonParameterizedTest.java
```diff
@@ -126,6 +126,22 @@ public void extract_h265WithoutGopParsingFlags() throws Exception {
         "extractordumps/mp4/fragmented_captions_h265.mp4_without_gop_parsing_flags");
   }
 
+  @Test
+  public void extract_h264WithVariableLengthSgpdBox() throws Exception {
```
icbaker: Does this test fail for you without the fix? It seems to pass for me

To show this, I checked out this PR, then ran:

```
$ git checkout HEAD^ -- libraries/extractor/src/main/java/androidx/media3/extractor/mp4/FragmentedMp4Extractor.java
```

And then ran the tests in this file, and they all passed.

dimitry-unified-streaming: Hm, I see what is happening; the `parseSampleGroups()` function scans only for `seig` grouping types, not `roll` ones. When it doesn't find any `seig` grouping type, it does an early return, and doesn't hit the `sgpd` box parsing code at all.

I had troubles getting any test to work with `seig` boxes, which is why the `sample_fragmented_variable_length_sgpd.mp4` file has another grouping type `roll`, e.g. MP4Box output:

```xml
<SampleGroupBox Size="28" Type="sbgp" Version="0" Flags="0" Specification="p12" Container="stbl traf" grouping_type="roll">
  <SampleGroupBoxEntry sample_count="94" group_description_index="1" group_description_in_traf="1"/>
</SampleGroupBox>
<SampleGroupDescriptionBox Size="30" Type="sgpd" Version="1" Flags="0" Specification="p12" Container="stbl traf" grouping_type="roll" default_length="0">
  <RollRecoveryEntry roll_distance="-1"/>
</SampleGroupDescriptionBox>
```

I'll try this again with a file with a `seig` box, but previously I got a `deduplicateConsecutiveFormats=false so TrackOutput must receive at least one sampleMetadata() call between format() calls.` error on that. I had no idea what to do to work around that other error. Maybe I have to supply some DRM metadata somewhere to make such a test work?

icbaker:
> I'll try this again with a file with a seig box, but previously I got a `deduplicateConsecutiveFormats=false so TrackOutput must receive at least one sampleMetadata() call between format() calls.` error on that. I had no idea what to do to work around that other error. 

If you get stuck on this again, please can you add the file that causes the error to the PR so I can have a play with it too?

dimitry-unified-streaming:
I replaced the `sample_fragmented_variable_length_sgpd.mp4` file with a file with an actual `seig` box. If you check this with e.g. `MP4Box -diso`, you will see:

```xml
<SampleGroupBox Size="28" Type="sbgp" Version="0" Flags="0" Specification="p12" Container="stbl traf" grouping_type="seig">
  <SampleGroupBoxEntry sample_count="48" group_description_index="1" group_description_in_traf="1"/>
</SampleGroupBox>
<SampleGroupDescriptionBox Size="65" Type="sgpd" Version="1" Flags="0" Specification="p12" Container="stbl traf" grouping_type="seig" default_length="0">
  <CENCSampleEncryptionGroupEntry IsEncrypted="1" IV_size="0" KID="0x804EA17317113C8B955B35996E5E3D92" constant_IV_size="16" constant_IV="0xDCC29B585406F0F0031C7D01C020C321"/>
</SampleGroupDescriptionBox>
```

However, as I indicated, this file leads to a different exception, which I do not know how to solve or work around:

```

deduplicateConsecutiveFormats=false so TrackOutput must receive at least one sampleMetadata() call between format() calls.
java.lang.IllegalStateException: deduplicateConsecutiveFormats=false so TrackOutput must receive at least one sampleMetadata() call between format() calls.
	at com.google.common.base.Preconditions.checkState(Preconditions.java:513)
	at androidx.media3.test.utils.FakeTrackOutput.format(FakeTrackOutput.java:90)
	at androidx.media3.extractor.mp4.FragmentedMp4Extractor$TrackBundle.updateDrmInitData(FragmentedMp4Extractor.java:2372)
	at androidx.media3.extractor.mp4.FragmentedMp4Extractor.onMoofContainerAtomRead(FragmentedMp4Extractor.java:920)
	at androidx.media3.extractor.mp4.FragmentedMp4Extractor.onContainerAtomRead(FragmentedMp4Extractor.java:789)
	at androidx.media3.extractor.mp4.FragmentedMp4Extractor.processAtomEnded(FragmentedMp4Extractor.java:758)
	at androidx.media3.extractor.mp4.FragmentedMp4Extractor.readAtomPayload(FragmentedMp4Extractor.java:753)
	at androidx.media3.extractor.mp4.FragmentedMp4Extractor.read(FragmentedMp4Extractor.java:586)
	at androidx.media3.test.utils.TestUtil.extractAllSamplesFromByteArray(TestUtil.java:614)
	at androidx.media3.test.utils.TestUtil.extractAllSamplesFromFile(TestUtil.java:577)
	at androidx.media3.extractor.mp4.FragmentedMp4ExtractorNonParameterizedTest.extract_h264WithVariableLengthSgpdBox(FragmentedMp4ExtractorNonParameterizedTest.java:134)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:524)
	at org.robolectric.internal.SandboxTestRunner.executeInSandbox(SandboxTestRunner.java:494)
	at org.robolectric.internal.SandboxTestRunner.access$900(SandboxTestRunner.java:67)
	at org.robolectric.internal.SandboxTestRunner$7.evaluate(SandboxTestRunner.java:442)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.robolectric.internal.SandboxTestRunner.access$600(SandboxTestRunner.java:67)
	at org.robolectric.internal.SandboxTestRunner$6.evaluate(SandboxTestRunner.java:333)
	at org.robolectric.internal.SandboxTestRunner$3.evaluate(SandboxTestRunner.java:233)
	at org.robolectric.internal.SandboxTestRunner$5.lambda$evaluate$0(SandboxTestRunner.java:317)
	at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:101)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
```

icbaker:
Thanks - I had a play with the file you provided and concluded that the duplicate format is expected because the `sgpd` is coming from within the `moof` and not the `moov` - which triggers a second format to be emitted when it's found. I've updated the test to allow this duplicate format for this specific sample file.

icbaker:

dimitry-unified-streaming:
Thanks for the updates and rebasing. I have tested your updates here locally, and they seem to all be successful.

PR #3247, opened 2026-05-28
fix(ttml): TTML/IMSC subtitles not shown when cue timing is on a <div> wrapping the <p>

### LINKED ISSUE #3246: TTML/IMSC subtitles not shown when cue timing is on a <div> wrapping the <p>
### Version

Media3 main branch

### More version details

Reproduced on main at commit 18ae2a419f. Also affects released versions that use the extraction-time subtitle path (TtmlParser / application/x-media3-cues).

Any device — the problem is in TTML parsing, not device-specific. Observed on a physical Android phone running the Media3 demo app and in a small native Media3 sample app.

Not applicable (parsing issue, independent of device).Not applicable (parsing issue, independent of device).

Yes

### Reproduction steps

1Take a DASH or HLS stream (or a standalone sidecar) whose TTML/IMSC subtitle carries the cue timing on a <div> that wraps the <p>, instead of on the <p> itself. A minimal document that reproduces it:

```
<?xml version="1.0" encoding="utf-8"?>
<tt xmlns="http://www.w3.org/ns/ttml"
    xmlns:ttp="http://www.w3.org/ns/ttml#parameter"
    ttp:contentProfiles="http://www.w3.org/ns/ttml/profile/imsc1.1/text"
    xml:lang="nl">
  <body>
    <div>
      <!-- timing is on the <div>, the <p> has no begin/end -->
      <div xml:id="d0" begin="00:01:09.583" end="00:01:11.125">
        <p>Lara.</p>
      </div>
      <div xml:id="d1" begin="00:01:27.208" end="00:01:28.833">
        <p>Stil.</p>
      </div>
    </div>
  </body>
</tt>
```

2 Play it in the demo app and enable the subtitle track.
3 Observe the subtitles (or attach an onCues listener / Player.Listener.onCues).

For comparison, the same content with the timing moved onto the <p> renders correctly:
```
<div>
  <p begin="00:01:09.583" end="00:01:11.125">Lara.</p>
  <p begin="00:01:27.208" end="00:01:28.833">Stil.</p>
</div>
```

### Spec references
- TTML2 §8.1.4 — <div> default timeContainer is par: https://www.w3.org/TR/ttml2/
- TTML2 §12.2.1 — absent begin defaults to 0s relative to the nearest time container.
- IMSC1.1 §6 feature table — #nested-div permitted in the Text profile: https://www.w3.org/TR/ttml-imsc1.1/

### Expected result

The cues are shown at their interval, e.g. "Lara." from 00:01:09.583 to 00:01:11.125 and "Stil." from 00:01:27.208 to 00:01:28.833 — the same as when the timing is on the <p>.

This is what the spec requires. A <div> is a parallel time container by default (TTML2 §8.1.4), and a child with no begin defaults to begin = 0s relative to that container (TTML2 §12.2.1, citing SMIL 3.0 §5.4.3). So the <p> should inherit the wrapping <div>'s start time. The construct is valid IMSC1.1 Text profile (nested <div> is permitted — IMSC1.1 §6 feature table, #nested-div).

### Actual result

No subtitles are shown.

When testing in a reference Media3 app with an onCues listener:

On the default extraction-time path (MediaItem parsing, MIME application/x-media3-cues), onCues is called with 0 cues for the whole track.

After forcing the legacy TtmlDecoder path, onCues does fire with the text, but every cue is active from the start of playback (out of sync) instead of at its interval:
```
onCues @ 1935063ms — 1 cue: "Pardon, c'est la première fois que je suis en cavale."
onCues @ 1944000ms — 1 cue: "Tu as pris la main jusqu'ici."
```

Root cause: `TtmlParser.parseNode` only inherits end from a timed parent, not begin. When the <p> has no begin, its startTimeUs is left unset. The extraction-time path (LegacySubtitleUtil#toCuesWithTiming) then never registers a start event for the cue, so no CuesWithTiming is emitted; the legacy path (TtmlNode#isActive) treats an unset start as "active from time 0", which is the out-of-sync behaviour.

### Media

See the inline TTML above (self-contained, no DRM). Real-world example: AWS MediaConvert IMSC (application/ttml+xml) output for some titles emits the <div begin/end>-wraps-<p> shape; other titles emit timing directly on <p> and play fine.

Stripped reference TTML we used to reproduce the issue:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<tt xmlns="http://www.w3.org/ns/ttml"
  xmlns:ttm="http://www.w3.org/ns/ttml#metadata"
  xmlns:ttp="http://www.w3.org/ns/ttml#parameter"
  xmlns:tts="http://www.w3.org/ns/ttml#styling"
  xmlns:ittm="http://www.w3.org/ns/ttml/profile/imsc1#metadata"
  xmlns:ittp="http://www.w3.org/ns/ttml/profile/imsc1#parameter"
  xmlns:itts="http://www.w3.org/ns/ttml/profile/imsc1#styling"
  xmlns:smpte="http://www.smpte-ra.org/schemas/2052-1/2010/smpte-tt"
  xmlns:ebutts="urn:ebu:tt:style"
  xmlns:ebuttm="urn:ebu:tt:metadata" xml:lang="nl" ttp:contentProfiles="http://www.w3.org/ns/ttml/profile/imsc1.1/text">
  <head>
    <styling>
      <style xml:id="backgroundStyle" tts:fontFamily="proportionalSansSerif" tts:fontSize="18px" tts:textAlign="center" tts:origin="0% 66%" tts:extent="100% 33%" tts:backgroundColor="rgba(0,0,0,0)" tts:displayAlign="center"/>
      <style xml:id="speakerStyle" style="backgroundStyle" tts:color="white" tts:textOutline="black 1px" tts:backgroundColor="transparent"/>
      <style xml:id="textStyle" style="speakerStyle" tts:color="white" tts:textOutline="none" tts:backgroundColor="black"/>
      <style xml:id="style.center" tts:color="#ffffff" tts:fontFamily="default" tts:fontSize="100%" tts:opacity="1" tts:textAlign="center"/>
    </styling>
    <layout>
      <region xml:id="full" tts:origin="0% 0%" tts:extent="100% 100%" tts:zIndex="1"/>
      <region xml:id="speaker" style="speakerStyle" tts:zIndex="1"/>
      <region xml:id="background" style="backgroundStyle" tts:zIndex="0"/>
      <region tts:origin="17.583% 73.414%" tts:extent="64.844% 16.667%" tts:displayAlign="center" tts:textAlign="center" xml:id="region.bottomCenter"/>
    </layout>
  </head>
  <body>
    <div>

      <!-- text-based TTML -->
      <div xml:id="d0" style="style.center" begin="00:01:09.583" end="00:01:11.125">
        <p region="region.bottomCenter">Lorum.</p>
      </div>

      <div style="style.center" begin="00:01:27.208" end="00:01:28.833">
        <p region="region.bottomCenter">Ipsum.</p>
      </div>
      <div style="style.center" begin="00:02:46.333" end="00:02:48.500">
        <p region="region.bottomCenter">Demo line!</p>
      </div>
      <div style="style.center" begin="00:02:48.583" end="00:02:49.625">
        <p region="region.bottomCenter">Closing statement.</p>
      </div>
    </div>
  </body>
</tt>
```

A small fix in `TtmlParser.parseNode` (inherit the parent's resolved start when the child's own begin is absent) resolves both symptoms. PR to follow.

### Bug Report

ivarb:
Inherit `begin` from a timed parent in TtmlParser

Fixes #3246

## Summary

TTML/IMSC subtitles are not shown when the cue timing is carried on a `<div>` that wraps the
`<p>` (the `<p>` itself has no `begin`/`end`). `TtmlParser.parseNode` inherits `end` from a
timed parent but not `begin`, so the `<p>`'s `startTimeUs` is left unset and the cue is dropped
(extraction-time path) or rendered from t=0 (legacy decoder path).

This change inherits the parent's resolved start time when the child's own `begin` is absent,
matching the TTML2 default-`begin`-of-0s rule.

## Why this is a bug (spec)

* `<div>` is a parallel time container by default (TTML2 §8.1.4).
* A child element with no `begin` defaults to `begin = 0s` relative to its nearest time
  container (TTML2 §12.2.1, citing SMIL 3.0 §5.4.3) — i.e. it inherits the container's start.
* Nested `<div>` is valid in the IMSC1.1 Text profile (IMSC1.1 §6 feature table, `#nested-div`).

So a `<p>` inside `<div begin="..." end="...">` must present at the `<div>`'s interval. This
shape is produced by real authoring tools (e.g. AWS MediaConvert IMSC output).

## Root cause

In `TtmlParser.parseNode`, after the parent-offset block:

* `end` is inherited from `parent.endTimeUs` when unset, but
* `start` is **not** inherited from `parent.startTimeUs` when unset.

Consequences with `startTimeUs` left unset:

* `LegacySubtitleUtil#toCuesWithTiming` (extraction-time path) registers no start event for the
  cue → **0 cues emitted**.
* `TtmlNode#isActive` returns `true` for an unset start → cue is **active from t=0** (out of
  sync) on the legacy `TtmlDecoder` path.

## The change

`libraries/extractor/.../ttml/TtmlParser.java` — add, after the parent-offset block and before
the `dur`→`end` computation:

```java
if (startTime == C.TIME_UNSET && parent != null && parent.startTimeUs != C.TIME_UNSET) {
  startTime = parent.startTimeUs;
}
```

Placement matters: resolving `start` before the `dur`→`end` step means a `<p dur="...">` inside
a timed `<div>` still computes `end = start + dur` correctly. Nested timed `<div>`s continue to
compound additively, and a plain `<p begin/end>` is unaffected.

## Testing

`libraries/extractor/.../ttml/TtmlParserTest.java` + one new fixture
`libraries/test_data/.../media/ttml/inherit_begin_from_div.xml`.

A single fixture exercises every in-scope `<div begin/end>`-wraps-`<p>` variant; the test
`inheritBeginFromDiv_allCues` asserts the full ordered cue list:

| Cue | Interval | Variant |
|---|---|---|
| `Lara.` | [69.583s, 71.125s) | timed div wraps untimed p |
| `Stil.` | [72.000s, 74.500s) | sibling timed div wraps untimed p |
| `offset` | [122.000s, 125.000s) | p own begin/end as offsets within the timed div |
| `dur` | [180.000s, 184.000s) | p with only dur inherits div start, end = start + dur |
| `nested` | [242.000s, 245.000s) | nested timed divs compound additively |
| `bare` | [300.000s, 302.000s) | plain `<p begin/end>` (regression guard) |

Full `androidx.media3.extractor.text.ttml.*` unit tests pass (TtmlParserTest, TextEmphasisTest,
TtmlRenderUtilTest, TtmlStyleTest) — 0 failures.

```shell
./gradlew :lib-extractor:testDebugUnitTest --tests "androidx.media3.extractor.text.ttml.*"
```

## Out of scope (noted for maintainers)

* `timeContainer="seq"` resolution is not implemented (still treated as `par`). Out of scope
  for this fix; EBU-TT-D — of which IMSC Text is a syntactic superset — forbids `seq`.
* A `<p dur="...">` with no `begin` on any ancestor remains a pre-existing edge case (unset
  start). Not addressed here to keep the change focused on the reported `<div>`-wraps-`<p>` case.

## Checklist

- [x] CLA signed
- [x] "Allow edits from maintainers" enabled on the PR
- [x] `google-java-format` applied to the diff

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/text/ttml/TtmlParser.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/text/ttml/TtmlParser.java
@@ -738,6 +738,13 @@ public final class TtmlParser implements SubtitleParser {
     if (parent != null && parent.startTimeUs != C.TIME_UNSET) {
       if (startTime != C.TIME_UNSET) {
         startTime += parent.startTimeUs;
+      } else {
+        // Per TTML2 §12.2.1 and SMIL 3.0 §5.4.3, when `begin` is absent on a timed element
+        // inside a parallel or sequential time container, the implicit value is 0s, which
+        // resolves to the parent's start time. Without this, an element whose timing is
+        // carried by a wrapping ancestor (e.g. a `<p>` inside a `<div begin="…">`) would
+        // keep an unset startTimeUs and never be emitted as a cue.
+        startTime = parent.startTimeUs;
       }
       if (endTime != C.TIME_UNSET) {
         endTime += parent.startTimeUs;
--- a/libraries/test_data/src/test/assets/media/ttml/inherit_begin_from_div.xml
+++ b/libraries/test_data/src/test/assets/media/ttml/inherit_begin_from_div.xml
@@ -1,0 +1,44 @@
+<tt xmlns="http://www.w3.org/ns/ttml"
+    xmlns:ttm="http://www.w3.org/ns/ttml#metadata"
+    xmlns:ttp="http://www.w3.org/ns/ttml#parameter"
+    xmlns:tts="http://www.w3.org/ns/ttml#styling"
+    ttp:contentProfiles="http://www.w3.org/ns/ttml/profile/imsc1.1/text"
+    xml:lang="en">
+  <head>
+    <styling>
+      <style xml:id="style.center" tts:color="white" tts:textAlign="center"/>
+    </styling>
+    <layout>
+      <region xml:id="region.bottomCenter" tts:origin="10% 80%" tts:extent="80% 15%"/>
+    </layout>
+  </head>
+  <body>
+    <div>
+      <!-- <p> with no timing of its own inside a timed <div>: it uses the div's begin/end -->
+      <div xml:id="d0" style="style.center" begin="00:01:09.583" end="00:01:11.125">
+        <p region="region.bottomCenter">Lara.</p>
+      </div>
+      <div xml:id="d1" style="style.center" begin="00:01:12.000" end="00:01:14.500">
+        <p region="region.bottomCenter">Stil.</p>
+      </div>
+      <!-- <p> with its own begin/end inside a timed <div>: its times are relative to the div -->
+      <div begin="00:02:00.000" end="00:02:30.000">
+        <p begin="00:00:02.000" end="00:00:05.000">offset</p>
+      </div>
+      <!-- <p> with only dur inside a timed <div>: end = the div's start + dur -->
+      <div begin="00:03:00.000" end="00:03:30.000">
+        <p dur="00:00:04.000">dur</p>
+      </div>
+    </div>
+    <!-- nested timed <div>s: the begin offsets add up -->
+    <div begin="00:04:00.000">
+      <div begin="00:00:02.000" end="00:00:05.000">
+        <p>nested</p>
+      </div>
+    </div>
+    <!-- <p> with its own begin/end under an untimed <div>: timed directly, nothing to inherit -->
+    <div>
+      <p begin="00:05:00.000" end="00:05:02.000">bare</p>
+    </div>
+  </body>
+</tt>
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -144,6 +144,8 @@
 *   Text:
     *   TTML: Fallback to `displayAlign` from `style` for regions
         ([#2559](https://github.com/androidx/media/issues/2559)).
+    *   TTML: Inherit cue start time from parent of `<p>` elements
+        ([#3246](https://github.com/androidx/media/issues/3246)).
 *   Metadata:
     *   Add `MediaMetadata.discSubtitle` field and parse it from ID3v2.4 `TSST`
         and Vorbis `DISCSUBTITLE` data.
--- a/libraries/extractor/src/test/java/androidx/media3/extractor/text/ttml/TtmlParserTest.java
+++ b/libraries/extractor/src/test/java/androidx/media3/extractor/text/ttml/TtmlParserTest.java
@@ -44,6 +44,8 @@ public final class TtmlParserTest {
 
   private static final String SIMPLE_TTML_FILE = "media/ttml/simple.xml";
   private static final String OVERLAPPING_TIMES_TTML_FILE = "media/ttml/overlapping_times.xml";
+  private static final String INHERIT_BEGIN_FROM_DIV_TTML_FILE =
+      "media/ttml/inherit_begin_from_div.xml";
   private static final String INLINE_ATTRIBUTES_TTML_FILE =
       "media/ttml/inline_style_attributes.xml";
   private static final String INHERIT_STYLE_TTML_FILE = "media/ttml/inherit_style.xml";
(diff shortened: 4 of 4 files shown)


libraries/test_data/src/test/assets/media/ttml/inherit_begin_from_div.xml
```diff
@@ -0,0 +1,44 @@
...
+        <p dur="00:00:04.000">dur</p>
+      </div>
+    </div>
+    <!-- nested timed <div>s compound additively -->
+    <div begin="00:04:00.000">
+      <div begin="00:00:02.000" end="00:00:05.000">
+        <p>nested</p>
+      </div>
+    </div>
+    <!-- bare <p begin/end> under an untimed <div> keeps pre-fix behaviour -->
```
icbaker: this comment only makes sense in the context of this change (pre fix) but it won't make sense to a reader who just looks at this file in a years time (without this change as context).

Please can you re-word it to be more 'stateless'.

libraries/extractor/src/main/java/androidx/media3/extractor/text/ttml/TtmlParser.java
icbaker, line 748: would it make more sense to put the logic in an `else` here?

Rather than repeating the check for `parent != null && parent.startTimeUs != C.TIME_UNSET`

ivarb: Hi @icbaker , thanks for the quick review. indeed, makes sense and easier to read as well. Changed.

libraries/test_data/src/test/assets/media/ttml/inherit_begin_from_div.xml
ivarb: Agreed. Adjusted the fixture comments to be time agnostic and describe the intended behaviour. Hope this feels better.

icbaker:

ivarb:
> I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks!

Thanks @icbaker. Will do

PR #3256, opened 2026-06-03
fix(dash): ignore presentationTimeOffset for standalone text

clown6613:
## Summary

Standalone text Representations (TTML / WebVTT / SubRip etc. that are not muxed into MP4) currently apply the manifest's `@presentationTimeOffset`, causing subtitle timings to be shifted by the offset.

[DASH-IF IOP Guidelines, "Standalone Text Timing"](https://dashif.org/Guidelines-TimingModel/#standalone-text-timing) state:

> @presentationTimeOffset SHALL NOT be present and SHALL be ignored by clients if present.

This PR makes the parser ignore `@presentationTimeOffset` for standalone text Representations, while continuing to apply it to muxed text (in `application/mp4`) and other media types.

## Changes

- `SegmentBase`: add abstract `copyWithPresentationTimeOffset(long)` and implement it in `SingleSegmentBase` / `SegmentList` / `SegmentTemplate`. Two `MultiSegmentBase` fields needed to be lowered from `private` to package-private so subclass copies can pass them through.
- `DashManifestParser.buildRepresentation`: when the Representation's `containerMimeType` is a text type (i.e. not `application/mp4`), replace its `SegmentBase` with one whose `presentationTimeOffset` is `0`. This zeros out both `Representation.presentationTimeOffsetUs` and the PTO subtraction in `SegmentBase.getSegmentTimeUs`, making the behavior compliant.
- Muxed text in MP4 and other media types are unaffected.

## Test plan

Added `DashManifestParserTest.parseMediaPresentationDescription_standaloneTextIgnoresPresentationTimeOffset` with a manifest containing three Representations sharing the same period and non-zero `@presentationTimeOffset`:

- [x] Standalone TTML (`application/ttml+xml`, PTO=2000/1000s) → `presentationTimeOffsetUs == 0`, first segment time = `@t` directly (2,000,000us).
- [x] Standalone WebVTT (`text/vtt`, PTO=3000/1000s) → `presentationTimeOffsetUs == 0`, first segment time = 3,000,000us.
- [x] Muxed TTML in MP4 (`application/mp4` + `stpp.ttml.im1t`, PTO=5000/1000s) → `presentationTimeOffsetUs == 5,000,000us`, first segment time = `@t - PTO = 0`.
- [x] Full `:lib-exoplayer-dash:testDebugUnitTest` suite passes with no regressions.

### DIFF
--- a/libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/manifest/DashManifestParser.java
+++ b/libraries/exoplayer_dash/src/main/java/androidx/media3/exoplayer/dash/manifest/DashManifestParser.java
@@ -977,10 +977,9 @@ public class DashManifestParser extends DefaultHandler
     }
     ArrayList<Descriptor> inbandEventStreams = representationInfo.inbandEventStreams;
     inbandEventStreams.addAll(extraInbandEventStreams);
-    Format format = formatBuilder.build();
     return Representation.newInstance(
         representationInfo.revisionId,
-        format,
+        formatBuilder.build(),
         representationInfo.baseUrls,
         representationInfo.segmentBase,
         inbandEventStreams,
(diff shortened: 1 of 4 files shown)


tonihei:
Thanks for the contribution. I will start the internal merge process!

tonihei:
One thing to note though is that the spec not only says "SHALL be ignored " but also  "@presentationTimeOffset SHALL NOT be present ", so the manifest is arguably also malformed according to the IOP guidelines.

PR #3257, opened 2026-06-03
EAC3-JOC MIME type should not be taken as EAC3 on Pixel Device

ybai001:
Pixel devices integrate an EAC3 decoder, but it does not support EAC3-JOC stream decoding. If there is a manifest file includes one AAC 2.0 track and one EAC3-JOC track.

- Without this PR, EAC3-JOC track is selected since channel count DD+JOC > AAC. Then playback failed.
- With this PR, AAC is the only MIME type that DUT can support so that AAC track is selected. Playback is OK.

Note: Pixel devices integrate EAC3 decoder but cannot handle EAC3-JOC stream. -- This is a product decision rather than technical limitation.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecUtil.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecUtil.java
@@ -378,7 +378,9 @@ public final class MediaCodecUtil {
   public static String getAlternativeCodecMimeType(Format format) {
     if (MimeTypes.AUDIO_E_AC3_JOC.equals(format.sampleMimeType)) {
       // E-AC3 decoders can decode JOC streams, but in 2-D rather than 3-D.
-      return MimeTypes.AUDIO_E_AC3;
+      // Some devices (e.g. Pixel) integrate an EAC3 decoder that does not support EAC3-JOC
+      // stream decoding.
+      return supportsEac3JocFallbackDecoding() ? MimeTypes.AUDIO_E_AC3 : null;
     }
     if (MimeTypes.AUDIO_DTS_HD.equals(format.sampleMimeType)
         || MimeTypes.AUDIO_DTS_UHD_P2.equals(format.sampleMimeType)) {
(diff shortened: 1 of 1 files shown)


microkatz:
@ybai001 

Thank you for reporting this issue! I think a better solution would be to isolate this "disabling" to the specific EAC3 decoder rather than the device? Would you be able to provide the decoder information?

ybai001:
> @ybai001
> 
> Thank you for reporting this issue! I think a better solution would be to isolate this "disabling" to the specific EAC3 decoder rather than the device? Would you be able to provide the decoder information?

Hi, microkatz,

Thanks for your reply. The point is the EAC3 decoder name is SAME for both Google and other OEMs (“c2.dolby.eac3.decoder”).

For Google release, we have
`        <MediaCodec name="c2.dolby.eac3.decoder" >`
`            <Type name="audio/ac3">`
`                <Alias name="OMX.dolby.ac3.decoder" />`
`                <Limit name="channel-count" max="6" />`
`                <Limit name="sample-rate" ranges="32000,44100,48000" />`
`                <Limit name="bitrate" range="32000-640000" />`
`            </Type>`
`            <Type name="audio/eac3">`
`                <Alias name="OMX.dolby.eac3.decoder" />`
`                <Limit name="channel-count" max="8" />`
`                <Limit name="sample-rate" ranges="32000,44100,48000" />`
`                <Limit name="bitrate" range="32000-6144000" />`
`            </Type>`
`            <Attribute name="software-codec" />`
`        </MediaCodec>`
        
But for other OEMs,
`        <MediaCodec name="c2.dolby.eac3.decoder" >`
`            <Type name="audio/ac3">`
`                <Alias name="OMX.dolby.ac3.decoder" />`
`                <Limit name="channel-count" max="6" />`
`                <Limit name="sample-rate" ranges="32000,44100,48000" />`
`                <Limit name="bitrate" range="32000-640000" />`
`            </Type>`
`            <Type name="audio/eac3">`
`                <Alias name="OMX.dolby.eac3.decoder" />`
`                <Limit name="channel-count" max="8" />`
`                <Limit name="sample-rate" ranges="32000,44100,48000" />`
`                <Limit name="bitrate" range="32000-6144000" />`
`            </Type>`
`            <Type name="audio/eac3-joc">`
`                <Alias name="OMX.dolby.eac3-joc.decoder" />`
`                <Limit name="channel-count" max="16" />`
`                <Limit name="sample-rate" ranges="48000" />`
`                <Limit name="bitrate" range="32000-6144000" />`
`            </Type>`
`            <Attribute name="software-codec" />`
`        </MediaCodec>`

So I'm afraid that we cannot isolate this "disabling" to the specific EAC3 decoder.

microkatz:
@ybai001 

Thank you for the quick response and noted! I see the understanding of the product decision and I will work to submit your suggestion.

ybai001:
@microkatz, I noticed that you changed this PR a little. In supportsEac3JocFallbackDecoding(), you simplified the implementation to check Build.MANUFACTURER only. Could you share your thought about this change?
**This is OK now**. But there is a potential risk. For example, ChromeBook also integrates DD+ decoder and that decoder can decode DD+JOC. Currently, Build.MANUFACTURER properties on these ChromeBooks are NOT Google so that everything is OK. If Google plan to produce ChromeBook by itself in future, there will be a problem.

microkatz:
Discussed outside of thread. The scenarios are as follows:

The [fix](https://github.com/androidx/media/commit/53ac10154ec1c6085627bf2dca8d224eab7bdf65) is adequate for the moment as there is not a way to tell if a decoder cannot handle fallback support of EAC3-JOC to EAC3.

If that changes in the future we can revisit this issue. Closing the PR.

PR #3259, opened 2026-06-03
Fix HLS download failure when EXT-X-DEFINE variables are referenced in media playlist segment URLs

### LINKED ISSUE #3258: HLS download fails when EXT-X-DEFINE variables from multivariant playlist are used in media playlist segment URLs
### Version

Media3 main branch

### More version details

Media3 1.10.1 (and all prior versions)

  All devices

_No response_

Yes

### Reproduction steps

1. Create a multivariant playlist that defines a variable:
  
  **master.m3u8:**
  
  #EXTM3U
  #EXT-X-DEFINE:NAME="cdnPrefix",VALUE="https://cdn.example.com/ (https://cdn.example.com/)"
  #EXT-X-STREAM-INF:BANDWIDTH=500000
  media.m3u8
  
  **media.m3u8:**
  
  #EXTM3U
  #EXT-X-TARGETDURATION:10
  #EXT-X-VERSION:8
  #EXTINF:10,
  {$cdnPrefix}segment001.ts
  #EXTINF:10,
  {$cdnPrefix}segment002.ts
  #EXT-X-ENDLIST
  
  2. Host these files on any HTTP server (or use `FakeDataSource` in a test)
  3. Create a `DownloadRequest` for `master.m3u8` and add it to `DownloadManager`
  4. Observe that segment requests go to the literal URL `{$cdnPrefix}segment001.ts` instead of `https://cdn.example.com/segment001.ts`
  
  ### Note
  Playback works correctly because `HlsPlaylistTracker` passes the multivariant playlist to the parser. The download path (`HlsDownloader` → `SegmentDownloader`) uses a single `HlsPlaylistParser()` that never
  stores the parsed multivariant playlist, so variable definitions are unavailable when media playlists are parsed.

### Expected result

Variables defined in the multivariant playlist should resolve in media playlist segment URLs during download, matching the playback behavior.

### Actual result

`HlsDownloader` uses a single `HlsPlaylistParser()` instance (constructed with `HlsMultivariantPlaylist.EMPTY`). After parsing the multivariant playlist, the result
  is not stored back into the parser. When media playlists are subsequently parsed by the same instance, `multivariantPlaylist` is still `EMPTY`, so
  `variableDefinitions` is unavailable and `{$variable}` references pass through unsubstituted.
  
  This causes:
  - Download failures (HTTP 404/400 on malformed URLs)
  - `IndexOutOfBoundsException` in `HlsMediaPeriod.getStreamKeys()` when `DownloadHelper.getDownloadRequest()` is called
  
  Stack trace:
  
  java.lang.IndexOutOfBoundsException: index (0) must be less than size (0)
      at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:1372)
      at com.google.common.collect.RegularImmutableList.get(RegularImmutableList.java:84)
      at androidx.media3.exoplayer.hls.HlsMediaPeriod.getStreamKeys(HlsMediaPeriod.java:289)
      at androidx.media3.exoplayer.offline.DownloadHelper.getDownloadRequestBuilder(DownloadHelper.java:939)
      at androidx.media3.exoplayer.offline.DownloadHelper.getDownloadRequest(DownloadHelper.java:891)

 ## Analysis
  
  - **Playback works** because `HlsPlaylistTracker` creates `HlsPlaylistParser(multivariantPlaylist, ...)` with the real multivariant playlist.
  - **Download fails** because `HlsDownloader` (via `SegmentDownloader`) uses the same parser instance for both master and media playlists, but the parser never stores
  the parsed multivariant playlist for reuse.
  
  The fix is to make `HlsPlaylistParser` store the multivariant playlist after parsing it in `parse()`, so subsequent media playlist parses on the same instance
  inherit the variable definitions.

### Media

No DRM

### Bug Report

prashant9934:
PR Description:
  
  ## Summary
  
  Fixes HLS download failures for streams where the multivariant playlist defines
  variables via `#EXT-X-DEFINE:NAME=...,VALUE=...` and media playlists reference
  them in segment URLs without an explicit `#EXT-X-DEFINE:IMPORT=...` tag.
  
  Fixes #3258
  
  ## Problem
  
  `HlsDownloader` (via `SegmentDownloader`) uses a single `HlsPlaylistParser()`
  instance constructed with `HlsMultivariantPlaylist.EMPTY`. This instance first
  parses the multivariant playlist, then parses each media playlist. However, the
  parsed multivariant playlist was never stored back into the parser instance.
  
  When media playlists are subsequently parsed, `multivariantPlaylist` is still
  `EMPTY`, so:
  - `variableDefinitions` is unavailable during `replaceVariableReferences()`
  - `{$variable}` references in segment URLs pass through unsubstituted
  - Downloads fail with HTTP errors on malformed URLs
  - `HlsMediaPeriod.getStreamKeys()` crashes with `IndexOutOfBoundsException`
  
  During **playback** this works correctly because `HlsPlaylistTracker` constructs
  the parser with the real multivariant playlist.
  
  ## Fix
  
  **`HlsPlaylistParser.java`:**
  - Make `multivariantPlaylist` field non-final
  - After parsing a multivariant playlist in `parse()`, store it in the instance
  - Subsequent `parseMediaPlaylist()` calls on the same instance now have access
    to the correct `variableDefinitions` for variable substitution
  
  **`HlsMediaPeriod.java`:**
  - Add bounds checks in `getStreamKeys()` to prevent `IndexOutOfBoundsException`
    when `redundantGroupIndicesPerWrapper` has an empty array or when
    `renditionRedundantGroupIndex` exceeds the redundant groups list size
  
  ## Testing
  
  Tested with an HLS stream using:
  
  #EXT-X-DEFINE:NAME="awsContentSegmentPrefix",VALUE="https://cdn.example.com/"
  
  in the multivariant playlist, with media playlist segment URLs referencing
  `{$awsContentSegmentPrefix}` directly. Download completes successfully with
  variables resolved. Existing playback behavior is unchanged

### DIFF
--- a/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
+++ b/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
@@ -233,7 +233,7 @@ public final class HlsDownloader extends SegmentDownloader<HlsPlaylist> {
     // When the multivariant playlist defines variables, create a parser that propagates
     // them to child manifests for IMPORT resolution. Otherwise use the default path.
     boolean hasVariables = !multivariantPlaylist.variableDefinitions.isEmpty();
-    @Nullable HlsPlaylistParser childManifestParser = hasVariables
+    @Nullable HlsPlaylistParser mediaPlaylistParser = hasVariables
         ? new HlsPlaylistParser(multivariantPlaylist, /* previousMediaPlaylist= */ null)
         : null;
 
@@ -243,11 +243,9 @@ public final class HlsDownloader extends SegmentDownloader<HlsPlaylist> {
       segments.add(new Segment(/* startTimeUs= */ 0, mediaPlaylistDataSpec));
       HlsMediaPlaylist mediaPlaylist;
       try {
-        if (childManifestParser != null) {
-          mediaPlaylist =
-              (HlsMediaPlaylist)
-                  ParsingLoadable.load(
-                      dataSource, childManifestParser, mediaPlaylistDataSpec, C.DATA_TYPE_MANIFEST);
+        if (mediaPlaylistParser != null) {
+          mediaPlaylist = (HlsMediaPlaylist) getManifest(dataSource, mediaPlaylistDataSpec,
+              removing, mediaPlaylistParser);
         } else {
           mediaPlaylist = (HlsMediaPlaylist) getManifest(dataSource, mediaPlaylistDataSpec,
               removing);
--- a/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
+++ b/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
@@ -29,7 +29,6 @@ import androidx.media3.exoplayer.hls.playlist.HlsMultivariantPlaylist;
 import androidx.media3.exoplayer.hls.playlist.HlsPlaylist;
 import androidx.media3.exoplayer.hls.playlist.HlsPlaylistParser;
 import androidx.media3.exoplayer.offline.SegmentDownloader;
-import androidx.media3.exoplayer.upstream.ParsingLoadable;
 import androidx.media3.exoplayer.upstream.ParsingLoadable.Parser;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.IOException;
--- a/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
+++ b/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
@@ -29,6 +29,7 @@ import androidx.media3.exoplayer.hls.playlist.HlsMultivariantPlaylist;
 import androidx.media3.exoplayer.hls.playlist.HlsPlaylist;
 import androidx.media3.exoplayer.hls.playlist.HlsPlaylistParser;
 import androidx.media3.exoplayer.offline.SegmentDownloader;
+import androidx.media3.exoplayer.upstream.ParsingLoadable;
 import androidx.media3.exoplayer.upstream.ParsingLoadable.Parser;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.IOException;
@@ -220,13 +221,21 @@ public final class HlsDownloader extends SegmentDownloader<HlsPlaylist> {
   protected List<Segment> getSegments(DataSource dataSource, HlsPlaylist manifest, boolean removing)
       throws IOException, InterruptedException {
     ArrayList<DataSpec> mediaPlaylistDataSpecs = new ArrayList<>();
+    HlsMultivariantPlaylist multivariantPlaylist = HlsMultivariantPlaylist.EMPTY;
     if (manifest instanceof HlsMultivariantPlaylist) {
-      HlsMultivariantPlaylist multivariantPlaylist = (HlsMultivariantPlaylist) manifest;
+      multivariantPlaylist = (HlsMultivariantPlaylist) manifest;
       addMediaPlaylistDataSpecs(multivariantPlaylist.mediaPlaylistUrls, mediaPlaylistDataSpecs);
     } else {
       mediaPlaylistDataSpecs.add(
           SegmentDownloader.getCompressibleDataSpec(Uri.parse(manifest.baseUri)));
     }
+
+    // When the multivariant playlist defines variables, create a parser that propagates
+    // them to child manifests for IMPORT resolution. Otherwise use the default path.
+    boolean hasVariables = !multivariantPlaylist.variableDefinitions.isEmpty();
+    @Nullable HlsPlaylistParser childManifestParser = hasVariables
+        ? new HlsPlaylistParser(multivariantPlaylist, /* previousMediaPlaylist= */ null)
+        : null;
 
     ArrayList<Segment> segments = new ArrayList<>();
     HashSet<Uri> seenEncryptionKeyUris = new HashSet<>();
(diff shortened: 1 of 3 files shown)


marcbaechinger:
Thanks. Please can you file the CLA? We can't process this before the CLA requirements.

As you are saying that the problem manifests in download use cases only. Would you be able to craft a unit test that demonstrates the failure and verifies the fix? That's what we have to do anyway internally when processing your fix, and if you are able to provide this, this may make taking the pull a bit quicker. Thanks for considering adding such a unit test. We'd need the CLA in any case.

libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
```diff
@@ -220,21 +221,37 @@ private HlsDownloader(
...
+
     ArrayList<Segment> segments = new ArrayList<>();
     HashSet<Uri> seenEncryptionKeyUris = new HashSet<>();
     for (DataSpec mediaPlaylistDataSpec : mediaPlaylistDataSpecs) {
       segments.add(new Segment(/* startTimeUs= */ 0, mediaPlaylistDataSpec));
       HlsMediaPlaylist mediaPlaylist;
       try {
-        mediaPlaylist = (HlsMediaPlaylist) getManifest(dataSource, mediaPlaylistDataSpec, removing);
+        if (childManifestParser != null) {
+          mediaPlaylist =
```
marcbaechinger: I think we should not load here. When looking at `SegmentDownloader.getManifest` (from which HlsDownaloder) extends, we see that internally this is processed through `execute`. This method handles quite some cases that need to be handled properly and also integrates with the `PriorityTaskManager` to do some priority management.

I think instead we should overload `SegmentDownloader.getManifest` and add an additional parameter that takes a `Parser<M>`. We can then in `HlsSegmentParser` pass down either the `childManfestParser` or the original parser that we pass down to `super` in the constructor. After calling `super` we cab assign it to a new field `HlsPlaylistParser hlsPlaylistParser` of `HlsDownloader` and we can then use it like:

```
Parser<HlsPlaylist> childManifestParser =
        hasVariables
            ? new HlsPlaylistParser(multivariantPlaylist, /* previousMediaPlaylist= */ null)
            : hlsPlaylistParser;
```

Instead of the `if (childManfestParser != null) {` we can then instead just do:

```
// calling the new overload method in the super class `SegementDownloader`
mediaPlaylist =
            (HlsMediaPlaylist)
                getManifest(dataSource, childManifestParser, mediaPlaylistDataSpec, removing);
```
This way we take advantage of the harness in `execute` in the same way as before.

libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
```diff
@@ -220,21 +221,37 @@ private HlsDownloader(
...
       addMediaPlaylistDataSpecs(multivariantPlaylist.mediaPlaylistUrls, mediaPlaylistDataSpecs);
     } else {
       mediaPlaylistDataSpecs.add(
           SegmentDownloader.getCompressibleDataSpec(Uri.parse(manifest.baseUri)));
     }
 
+    // When the multivariant playlist defines variables, create a parser that propagates
+    // them to child manifests for IMPORT resolution. Otherwise use the default path.
+    boolean hasVariables = !multivariantPlaylist.variableDefinitions.isEmpty();
+    @Nullable HlsPlaylistParser childManifestParser = hasVariables
```
marcbaechinger: rename from `childManifestParser` to `mediaPlaylistParser` ?

marcbaechinger:
Many thanks for the fix!

I've added some comment.

First though, the change is based on the 'release' branch. We would need that from the 'main' branch.

I suggested some changes in the comments. If you want to apply these and rebase the change onto main, please let me know. This way you get the credits in the commit logs attributed to your user. 

Otherwise, I have this change ready in a local change where I veryfied my suggestions for change. So quickest would be I just apply this change and send it for internal review. 

Please let me know what you prefer and give me a go to continue on my change or make a new PR and let me know when you are ready for review with the changes from above applied. 

Thanks again for the change and the unit tests!

prashant9934:
> Many thanks for the fix!
> 
> I've added some comment.
> 
> First though, the change is based on the 'release' branch. We would need that from the 'main' branch.
> 
> I suggested some changes in the comments. If you want to apply these and rebase the change onto main, please let me know. This way you get the credits in the commit logs attributed to your user.
> 
> Otherwise, I have this change ready in a local change where I veryfied my suggestions for change. So quickest would be I just apply this change and send it for internal review.
> 
> Please let me know what you prefer and give me a go to continue on my change or make a new PR and let me know when you are ready for review with the changes from above applied.
> 
> Thanks again for the change and the unit tests!



> Many thanks for the fix!
> 
> I've added some comment.
> 
> First though, the change is based on the 'release' branch. We would need that from the 'main' branch.
> 
> I suggested some changes in the comments. If you want to apply these and rebase the change onto main, please let me know. This way you get the credits in the commit logs attributed to your user.
> 
> Otherwise, I have this change ready in a local change where I veryfied my suggestions for change. So quickest would be I just apply this change and send it for internal review.
> 
> Please let me know what you prefer and give me a go to continue on my change or make a new PR and let me know when you are ready for review with the changes from above applied.
> 
> Thanks again for the change and the unit tests!

Thanks for the review and the suggestions! Please go ahead and apply your change with the suggested modifications and send it for internal review. Happy to have it land quickly — attribution isn't a concern for me.
  
Let me know if there's anything else I can help with!

libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
prashant9934: updated review changes

libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/offline/HlsDownloader.java
prashant9934: updated the changes

PR #3271, opened 2026-06-11
Fix some dodgy math

nift4:
I was expanding SilenceSkippingAudioProcessor to cover different sample formats and noticed a couple instances of dodgy math.

- SilenceSkippingAudioProcessor was not considering `bytesPerFrame` when calculating `maybeSilenceBufferSize` which in practice led to `minimumSilenceDurationUs` being divided by 4 (channel count 2 and sample format 2). The test `skipInNoisySignalWithShortSilences_skipsNothing` should've caught this but didn't, due to a bug in the test:
- In `Pcm16BitAudioBuilder`, `appendFrames` was only generating half of the supposed millisecond amounts (but due to the loop in the caller, this wasn't noticed as the output was still correct size, just with wrong amount of silence/noise between each change). This is because the loop was accepting a frame count but did `i += channelCount`, that is, treating it as sample count. Thus, `skipInNoisySignalWithShortSilences_skipsNothing` only generated 15ms of silence and noise, which is below the threshold of 25ms (which is the intended default 100ms but divided by 4 due to aforementioned bug), and thus passed.
- In `modifyVolume`, the volume percentage was calculated from the byte index, which could lead to the left and right channel having a different percentage or prevent reaching 100% of volume in edge cases.
- In PcmAudioUtil, 8-bit PCM is treated as signed, but Android defines 8-bit PCM to be unsigned.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -69,6 +69,12 @@
         `MetadataRetriever`.
 *   Inspector Frame:
 *   Audio:
+    *   Fix buffer size calculation in `SilenceSkippingAudioProcessor` so that
+        the minimum silence duration is not incorrectly scaled down by the frame
+        size ([#3271](https://github.com/androidx/media/pull/3271)).
+    *   Fix 8-bit PCM handling in `PcmAudioUtil` to treat samples as unsigned as
+        defined by Android
+        ([#3271](https://github.com/androidx/media/pull/3271)).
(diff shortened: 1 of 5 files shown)


nift4:
Hi @microkatz, please can you check this PR? Thanks!

microkatz:

nift4:
Thanks!

PR #3276, opened 2026-06-16
Refactor audio sample calculation for EAC3

minlinbroadcom:
Calculate the number of audio samples based on frame size and number of frames for EAC3.

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/Ac3Util.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/Ac3Util.java
@@ -537,10 +537,10 @@ public final class Ac3Util {
     // Parse the bitstream ID for AC-3 and E-AC-3 (see subsections 4.3, E.1.2 and E.1.3.1.6).
     boolean isEac3 = ((buffer.get(buffer.position() + 5) & 0xF8) >> 3) > 10;
     if (isEac3) {
-      // An EC-3 access unit may contain multiple syncframes. Per ETSI TS 102 366 Annex F,
-      // dependent substreams (strmtyp=1) cover the same presentation time as the preceding
-      // independent substream, so only audio blocks from independent substreams count toward the
-      // total sample duration.
+      // Per ETSI TS 102 366 Annex F, an EC-3 sample represents six audio blocks (1536 PCM
+      // samples). Additional independent substreams (substreamid > 0) and dependent substreams
+      // (strmtyp=1) are concurrent services that do not extend the presentation duration. Count
+      // only audio blocks from the primary independent substream (strmtyp=0, substreamid=0).
       int totalAudioBlocks = 0;
       int pos = buffer.position();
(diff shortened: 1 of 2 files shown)


FongMi:
For #3276, the sample count calculation can overcount when an EC-3 sample contains multiple syncframes for different substreams.

In `Ac3Util.java`, the new code derives `numFrames` from `buffer.remaining() / firstFrameBytes` and returns `numFrames * samplesPerFrame`. However, in ISO BMFF, an E-AC-3 sample may contain multiple syncframes for different substreams or shorter block frames, and those syncframes do not necessarily represent consecutive presentation time. Per ETSI TS 102 366 Annex F, the EC-3 sample duration is still six audio blocks, i.e. 1536 PCM samples.

For example, a sample with two substreams and three audio blocks per syncframe may contain four syncframes. This implementation would return `4 * 768 = 3072`, but the correct sample duration is still 1536. I think this should either be reverted or implemented as MP4/EC-3-specific logic, rather than changing a utility method whose contract is to parse one syncframe and relying on `buffer.remaining()`.

FongMi:
Thanks for the update. This fixes the dependent-substream overcount, but I think the duration can still be overcounted when the EC-3 bitstream contains multiple independent substreams.

The current code adds audio blocks for every syncframe whose `strmtyp` is not TYPE1. That means it also adds blocks from independent substreams with non-zero `substreamid`. Per ETSI TS 102 366 Annex F.2, an AC-3/E-AC-3 sample still represents six audio blocks / 1536 PCM samples; additional independent substreams are concurrent services/substreams, not additional presentation time. Annex F.6 also defines `num_ind_sub` and associated services, so summing all independent substreams can make the sample duration too large.

I think this should count only one presentation timeline, for example independent substream 0, or otherwise keep the MP4 EC-3 sample duration fixed at 1536 samples. It would also be good to add tests for an access unit with dependent substreams and one with multiple independent substreams.

FongMi:
https://github.com/FongMi/media/commit/5e43a057713a2d982edc4347b5c9d079034ee954

minlinbroadcom:
@FongMi Thanks again for the feedback! I've merged the latest commits from  https://github.com/FongMi/media/commit/5e43a057713a2d982edc4347b5c9d079034ee954. Could you please take another look when you have a chance?

FongMi:
The updated implementation looks correct to me now. It counts E-AC3 access units per substream timeline, skips dependent substreams, and uses the longest independent substream duration, so it avoids the previous overcounting cases.

One small follow-up: the public javadoc for parseAc3SyncframeAudioSampleCount still says the input is a syncframe, but the implementation now also handles E-AC3 access units containing multiple syncframes. It may be worth updating the wording to "syncframe or access unit" and briefly mentioning that E-AC3 access units use the longest non-dependent substream timeline.

I also verified Ac3UtilTest locally and it passes.

FongMi:
Good luck, I hope it gets merged.

PR #3280, opened 2026-06-18
support Dolby Vision in HLS/TS streams

CharmingCheung:
Route HEVC-based Dolby Vision elementary streams in MPEG-TS to the Dolby Vision decoder instead of decoding the HEVC base layer as plain HEVC, which produced a green/magenta tint for non-cross-compatible profiles (e.g. profile 5).

- TsExtractor: detect the "DOVI" registration descriptor and the 0xB0 Dolby Vision video descriptor in the PMT, parsing profile/level via DolbyVisionConfig and carrying it through EsInfo.
- DefaultTsPayloadReaderFactory: pass the DV config to H265Reader.
- H265Reader: when DV is signalled, emit a video/dolby-vision Format with the dvhe codec string; for profile 5 (proprietary IPT-PQ base layer whose SPS omits colour signalling) synthesize BT.2020/PQ/limited colour info, leaving cross-compatible profiles to use their SPS VUI.
- H265Reader/NalUnitUtil: keep Dolby Vision RPU (NAL type 62) and enhancement-layer (type 63) NAL units within the current access unit so the per-frame RPU metadata reaches the decoder instead of being dropped at the sample boundary.

Verified on a OnePlus 12 with DV profile 5 and profile 8.1 HLS streams: both select c2.qti.dv.decoder and render correctly.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -98,6 +98,13 @@
         custom format ordering and ABR selection priority beyond bitrate-only
         ordering.
 *   Extractors:
+    *   MPEG-TS: Add Dolby Vision support in HLS/TS streams. The extractor now
+        detects the `DOVI` registration descriptor and the `0xB0` Dolby Vision
+        video descriptor in the PMT, and routes HEVC-based Dolby Vision
+        elementary streams to the Dolby Vision decoder with the correct profile,
+        level, and colour info. This fixes a green/magenta tint on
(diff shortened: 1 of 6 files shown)


osscv:
Yes, verified works, hope it gets merged.
Tested on Sony Android TV, Xiaomi 13.

I also created a custom version (which DolbyVisionExtractorsFactory.kt, DolbyVisionTsPayloadReaderFactory.kt
and etc hahha as we discussed before this)

PR #3309, opened 2026-06-29
Update DefaultAudioSink.java

AndresBC-123:
Custom audio processors are allowed on any type of audio.


icbaker:
Please target this change against the `main` branch.

PR #3312, opened 2026-07-02
Add muxer support for Dolby Atmos

rexjin-meta:
## Summary

  Add support for muxing Dolby Atmos (E-AC-3 JOC) tracks in Mp4Muxer.

##  Problem

Mp4Muxer and FragmentedMp4Muxer did not support AUDIO_E_AC3 or AUDIO_E_AC3_JOC MIME types. Attempting to mux an E-AC-3 or Dolby Atmos track would fail at the supported-format check, and even if bypassed, the codec-specific box  (dec3) and fourcc (ec-3) were not implemented. This made stream-copy of Dolby Atmos audio from one MP4 container to another impossible using the in-app muxer.

##  Changes

  - Boxes.java: Added dec3Box() which wraps the raw EC3SpecificBox payload from format.initializationData (csd-0) verbatim into a dec3 box. Added AUDIO_E_AC3 / AUDIO_E_AC3_JOC cases in codecSpecificBox() and codecSpecificFourcc()
  (fourcc: ec-3).
  - Mp4Muxer.java: Added AUDIO_E_AC3 and AUDIO_E_AC3_JOC to SUPPORTED_AUDIO_SAMPLE_MIME_TYPES.
  - FragmentedMp4Muxer.java: Same addition to SUPPORTED_AUDIO_SAMPLE_MIME_TYPES.
  - BoxesTest.java: Added 3 unit tests — dec3 payload wrapping for E-AC3, dec3 payload wrapping for E-AC3 JOC, and a guard test verifying IllegalArgumentException when csd-0 is missing.

##  Testing

  - ./gradlew :lib-muxer:compileDebugJavaWithJavac — clean compile.
  - ./gradlew :lib-muxer:test — all existing tests pass; 3 new unit tests added and passing.
  - No device testing yet; the new code path is only exercised during stream-copy (no encoding/decoding involved).

##  Notes

  - This is a stream-copy only implementation. The dec3 payload must be populated in format.initializationData by the caller (e.g. copied verbatim from the source track's EC3SpecificBox). There is no encoding-side support.
  - The dec3 payload is carried verbatim with no parsing or validation beyond the non-empty check — consistent with how av1C and dOps are handled.
  - Follow-up work: end-to-end test with a real Atmos source file via Mp4MuxerEndToEndParameterizedTest once test assets are available.

PR #3317, opened 2026-07-02
Adds E-AC-3 / Dolby Atmos (E-AC-3 JOC) support across the MP4 extractor and muxer.

rexjin-meta:
Adds E-AC-3 / Dolby Atmos (E-AC-3 JOC) support across the MP4 extractor and muxer.
- New Mp4FormatSpecificMetadataEntry (Metadata.Entry, media3-container) carries a raw box payload for transmuxing.
- Extractor attaches the dec3 payload to Format.metadata (not initializationData), so the E-AC-3 decode path is unchanged.
- Muxer writes the dec3 box / ec-3 fourcc from that metadata entry; AUDIO_E_AC3 / AUDIO_E_AC3_JOC added to Mp4Muxer + FragmentedMp4Muxer supported types, so Transformer can transmux Atmos.
- Tests: BoxesTest + end-to-end transmux (Mp4MuxerEndToEndTest, FragmentedMp4MuxerEndToEndTest); extractor golden dumps regenerated.

Let's go Dolby!

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -24,6 +24,9 @@
         Custom `AssetLoader` implementations must implement these new methods.
 *   Track Selection:
 *   Extractors:
+    *   MP4: Preserve E-AC-3 and E-AC-3 JOC (Dolby Atmos) format metadata during
+        demuxing to enable lossless stream-copying
+        ([#3317](https://github.com/androidx/media/pull/3317)).
 *   Inspector:
 *   Inspector Frame:
 *   Audio:
@@ -40,6 +43,9 @@
 *   Muxers:
     *   Fix crash in `Mp4Writer` when writing an EOS sample before any other
         samples are written.
+    *   MP4: Add support for writing E-AC-3 and E-AC-3 JOC (Dolby Atmos) audio
+        tracks in `Mp4Muxer` and `FragmentedMp4Muxer`
+        ([#3317](https://github.com/androidx/media/pull/3317)).
 *   IMA extension:
 *   Session:
 *   UI:
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -24,9 +24,6 @@
         Custom `AssetLoader` implementations must implement these new methods.
 *   Track Selection:
 *   Extractors:
-    *   MP4: Preserve E-AC-3 and E-AC-3 JOC (Dolby Atmos) format metadata during
-        demuxing to enable lossless stream-copying
-        ([#3317](https://github.com/androidx/media/pull/3317)).
 *   Inspector:
 *   Inspector Frame:
 *   Audio:
@@ -43,9 +40,6 @@
 *   Muxers:
     *   Fix crash in `Mp4Writer` when writing an EOS sample before any other
         samples are written.
-    *   MP4: Add support for writing E-AC-3 and E-AC-3 JOC (Dolby Atmos) audio
-        tracks in `Mp4Muxer` and `FragmentedMp4Muxer`
-        ([#3317](https://github.com/androidx/media/pull/3317)).
 *   IMA extension:
 *   Session:
 *   UI:
--- a/libraries/container/src/main/java/androidx/media3/container/FormatSpecificTransmuxingData.java
+++ b/libraries/container/src/main/java/androidx/media3/container/FormatSpecificTransmuxingData.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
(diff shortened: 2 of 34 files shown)


FongMi:
**Findings**
1. **P1 `Mp4Muxer.java:417` / `FragmentedMp4Muxer.java:184`**  
   Adding `AUDIO_E_AC3` / `AUDIO_E_AC3_JOC` to the supported list causes the default Transformer to directly transmux E-AC-3. However, `dec3Box()` requires the raw `dec3` payload in `Format.initializationData`, while the existing MP4 extractor parses `dec3` into Format fields without preserving initialization data. For common E-AC-3/Atmos MP4 or TS inputs, this may change behavior from transcoding to AAC into direct copy, then fail later while muxing `moov` because `csd-0` is missing. I’d recommend either preserving raw `dec3` in `BoxParser` in the same PR, or not adding these MIME types to the factory supported list yet.

2. **P3 `Boxes.java:1658`**  
   `dec3Box()` only checks that `initializationData` is non-empty, but does not check that `csd-0` itself has bytes. An empty payload would produce an empty `dec3` box, which is not a valid EC3SpecificBox. This should mirror `esdsBox()` / `hvcCBox()` with `checkArgument(csd0.length > 0, ...)`, plus a test for an empty byte array.

rexjin-meta:
Hey @FongMi and @snow2405 , Thanks for the reviews — addressed both comments:

- P1 (Transformer regression): Dropped AUDIO_E_AC3 / AUDIO_E_AC3_JOC from SUPPORTED_AUDIO_SAMPLE_MIME_TYPES in both Mp4Muxer and FragmentedMp4Muxer, so the default Transformer keeps transcoding E-AC-3 to AAC — no behavior change. The muxer retains the dec3/ec-3 write capability for direct callers that supply csd-0. Added a comment on dec3Box() explaining the deferral.

- P3 (empty csd-0): Added checkArgument(csd0.length > 0, "csd-0 is empty for dec3 box."), mirroring esdsBox() / hvcCBox(), plus a test for the empty-byte[] case.

Also added a unit test asserting the ec-3 fourcc on the audio sample entry.

This PR will laser focus on enabling muxer to write atmos. 
Follow-up (separate PR): preserve the raw dec3 payload in BoxParser, then add E-AC-3 to the supported list and add an end-to-end mux test (with the audio_sample_entry_box_eac3 golden dump). This staging keeps the current PR free of the extractor change and any Transformer impact.

rexjin-meta:
cc: @andrewlewis  This PR is the new attempt for my previously closed PR: https://github.com/androidx/media/pull/3312#issuecomment-4864036707

rexjin-meta:
Updated the PR to fix the extractor and muxer together, as suggested.

Extractor: BoxParser now preserves the raw dec3 (EC3SpecificBox) payload in Format.initializationData when parsing the MP4 dec3 box — mirroring the existing av1C/dOps handling (box body without header). This is the piece that was missing.

Muxer: implements the dec3 box + ec-3 fourcc and adds AUDIO_E_AC3 / AUDIO_E_AC3_JOC to the supported audio MIME types. With the extractor change in place, the default Transformer now correctly stream-copies E-AC-3 / Atmos (instead of transcoding), and it no longer risks the missing-csd-0 failure — the two changes are consistent across extractor, muxer, and transformer.

Tests: added an end-to-end Mp4MuxerEndToEndParameterizedTest pass for sample_eac3.mp4 and sample_eac3joc.mp4; regenerated the extractor golden dumps (they now include the dec3 initializationData); BoxesTest covers dec3 wrapping, the ec-3 fourcc, and the  missing/empty csd-0 guards.

rexjin-meta:
Reworked per your suggestion. Added Mp4FormatSpecificMetadataEntry (a Metadata.Entry in media3-container, boxType + raw data). The extractor now attaches the dec3 payload to Format.metadata instead of initializationData, so the E-AC-3 decode path is unchanged; the muxer reads it back from Format.metadata to write the dec3 box.

Tests: BoxesTest covers dec3 wrapping, the ec-3 fourcc, and missing/empty-payload guards; Mp4MuxerEndToEndTest +  FragmentedMp4MuxerEndToEndTest do a full Mp4Extractor → mux → re-extract round trip for  mple_eac3(.mp4|joc.mp4). Extractor golden dumps regenerated.

Made the entry general (boxType + data) rather than E-AC-3-specific — open to either.

snow2405:
Hey @rexjin-meta 

osscv:
It looks like your method is similar to mine. I solved it using the same approach, hahha but mine code is complex, great work @rexjin-meta 

hopefully they appove this in the stable version soon so i can migrate to urs haha

Issue #3327, opened 2026-07-09
PsExtractor does not strip the private_stream_1 sub-stream header, corrupting AC-3/DTS/LPCM audio in MPEG-PS


TsaiHao:
### Version

Media3 main branch

### More version details

Reproduced on `main` (HEAD as of 2026-07-09) and confirmed present in all releases through 1.10.1. The relevant code (`PsExtractor.PesReader.consume()`) is unchanged across these versions.

### Devices that reproduce the issue

Reproduced on `main` (HEAD as of 2026-07-09) and confirmed present in all releases through 1.10.1. The relevant code (`PsExtractor.PesReader.consume()`) is unchanged across these versions.

### Reproducible in the demo app?

Yes

### Reproduction steps

**Root cause:** In an MPEG Program Stream, AC-3 (also DTS and LPCM) audio is carried in `private_stream_1` (PES stream id `0xBD`). In DVD-style content, each such PES packet begins with a 4-byte sub-stream header immediately after the PES header:

```
sub_stream_id (1 byte)              // 0x80–0x87 = AC-3, 0x88–0x8F = DTS, 0xA0–0xA7 = LPCM
number_of_frame_headers (1 byte)
first_access_unit_pointer (2 bytes)
```

`PsExtractor.PesReader.consume()` strips the PES header + PTS/DTS extension but **does not strip this 4-byte sub-stream header** — it forwards it straight into `Ac3Reader` as if it were AC-3 elementary data:

https://github.com/androidx/media/blob/main/libraries/extractor/src/main/java/androidx/media3/extractor/ts/PsExtractor.java#L362-L370

Because AC-3 sync frames do not align to PES packet boundaries, the 4 stray bytes land **inside** the AC-3 frame that spans each PES boundary, corrupting that frame's bitstream. `Ac3Reader` re-syncs on the next `0x0B77` syncword, so the corruption is invisible to a frame-count/size check (every emitted frame still starts with a valid syncword) but every frame straddling a boundary is corrupt. ffmpeg's `mpegps` demuxer strips these bytes, which is why VLC/ffmpeg play the same content correctly. `first_access_unit_pointer` (which a correct demuxer uses to realign to the first frame that starts within the packet) is also unused.

**Minimal desktop repro** (no device needed) — feed the attached `sample_ac3_ps.ps` through `PsExtractor` with `FakeExtractorInput`/`FakeExtractorOutput`, concatenate the AC-3 track's samples, and decode with ffmpeg:

```java
[...]

PR #3328, opened 2026-07-10
Strip DVD private_stream_1 sub-stream header in PsExtractor

TsaiHao:
In an MPEG Program Stream, AC-3 (and DTS/LPCM) audio is carried in private_stream_1 (PES stream id 0xBD). In DVD-style content each such PES packet begins with a 4-byte sub-stream header immediately after the PES header: sub_stream_id (0x80-0x87 AC-3, 0x88-0x8F DTS, 0xA0-0xA7 LPCM), number_of_frame_headers and first_access_unit_pointer.

`PsExtractor.PesReader.consume()` stripped only the PES header and PTS/DTS extension, forwarding the sub-stream header straight into the audio reader as if it were elementary data. Because audio sync frames do not align to PES packet boundaries, these 4 bytes land inside the frame spanning each boundary and corrupt it, producing garbled audio (and decoder errors such as Dolby MS12 "UDC Node: Error extracting BSI metadata (err 1021)") when playing MPEG-PS content that VLC/ffmpeg play correctly.

Skip the sub-stream header for private_stream_1 packets before handing the payload to the elementary stream reader, matching ffmpeg's mpegps demuxer.

Adds a regression test with a small MPEG-PS asset whose AC-3 frames straddle PES boundaries (the existing single-packet sample_ac3.ps cannot expose this).

Issue: #3327


kristinatrajkovski:
Thank you for contributing this PR and for the detailed fix! I'm going to send this for internal review now. You may see some more commits being added as I make changes in response to review feedback. Please refrain from pushing any more substantive changes as it will complicate the internal review - thanks.

PR #3330, opened 2026-07-11
Fix 14-bit DTS frame size calculation

Tolriq:
ETSI TS 102 114 V1.2.1 section 5.3 defines 14-bit DTS storage as one 14-bit payload word in the least significant bits of each 16-bit word. Convert FSIZE+1 to a whole number of 14-bit words before mapping those words to physical bytes.

The previous fsize * 16 / 14 calculation could produce an impossible odd physical size. For common DTS-CD headers with FSIZE+1 equal to 3585 it returned 4097 bytes, although consecutive sync words are 4096 bytes apart. DtsReader then consumed the first byte of the next sync word and failed to recognize subsequent frames.

Add direct coverage for little- and big-endian 14-bit headers and for an already-exact conversion.

Test: ./gradlew :lib-extractor:testDebugUnitTest --tests androidx.media3.extractor.DtsUtilTest

### DIFF
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/DtsUtil.java
@@ -408,8 +408,14 @@ public final class DtsUtil {
         fsize = (((data[5] & 0x03) << 12) | ((data[6] & 0xFF) << 4) | ((data[7] & 0xF0) >> 4)) + 1;
     }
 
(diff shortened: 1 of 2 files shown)


osscv:
Hi, do u have any test stream?

Tolriq:
Not directly usable here as they are dts in wav that is not supported by the current extractors. But since those dtsutils and reader are reusable I supposed the fix still have it's place upstream.

PR #3343, opened 2026-07-24
Fix issue where ForwardingAudioOutputProvider wasn't forwarding audio capabilities

nift4:


tonihei:
This isn't as straight-forward as you make it look like I'm afraid. The `AudioOutputProvider` has methods to check the format support internally based on its knowledge of the output path. The `AudioCapabilties` class is tighly coupled to the default `AudioTrack` output path and can't be easily moved to the stable API surface. That's the reason it's only available on the `AudioTrackAudioOutputProvider`. In fact, the method was only added there as a slightly hacky workaround to retrieve the available spatialization channel masks to configure IAMF output in the codec, and it has no other purpose. Given that, we'd need to find a more generic solution for the IAMF config problem instead of simply publishing the `AudioCapabilities` via the interface.

nift4:
Thanks for the explaination, I see. So it's only a short term solution, that makes sense.

nift4:
Are you interested in a PR taking a shot at a more generic API to query suggested spatialization, or is the problem it not being spec'd out yet?

tonihei:
We haven't thought this through yet to see what the best design would be exactly to be honest. If you have a suggestion for something that fits nicely into the interface, feel free to suggest it for discussion.

Issue #3135, opened 2026-03-24
[DASH] DashManifestParser unconditionally promotes supplementalCodecs to Dolby Vision, ignoring backward-compatible base codec


akhilesh-dubey:
### Version

Media3 main branch

### More version details

DashManifestParser.buildFormat() unconditionally promotes HEVC to Dolby Vision when supplementalCodecs is present, causing wrong track selection

### Devices that reproduce the issue

All Android Device

### Devices that do not reproduce the issue

NA

### Reproducible in the demo app?

Yes

### Reproduction steps

Play a DASH stream where the manifest contains both AVC and HEVC AdaptationSets, and the HEVC representations include `scte214:supplementalCodecs="dvh1.08.xx"` (Dolby Vision Profile 8):

```
<!-- AVC AdaptationSet -->
<AdaptationSet mimeType="video/mp4" codecs="avc1.640020">
  <Representation bandwidth="3911733" codecs="avc1.64001f" height="720" width="1280" />
  <Representation bandwidth="10246893" codecs="avc1.64002a" height="1080" width="1920" />
</AdaptationSet>

<!-- HEVC AdaptationSet with DV Profile 8 supplementalCodecs -->
<AdaptationSet mimeType="video/mp4" codecs="hvc1.2.4.L123">
  <Representation bandwidth="4013913" codecs="hvc1.2.4.L93" height="720" width="1280"
      scte214:supplementalCodecs="dvh1.08.02" scte214:supplementalProfiles="db1p" />
  <Representation bandwidth="8714193" codecs="hvc1.2.4.L123" height="1080" width="1920"
      scte214:supplementalCodecs="dvh1.08.05" scte214:supplementalProfiles="db1p" />
</AdaptationSet>
```

Observe that DefaultTrackSelector selects the AVC track instead of the HEVC track.

### Expected result

ExoPlayer should select the HEVC track. Per RFC 8216bis Section 4.4.6.2, SUPPLEMENTAL-CODECS (the HLS equivalent of SCTE-214's scte214:supplementalCodecs) describes:
> "media samples with both a backward-compatible base layer and a newer enhancement layer. The base layers are specified in the CODECS attribute and the enhancement layers are specified by the SUPPLEMENTAL-CODECS attribute."
The Dolby Vision Profiles and Levels specification further confirms that Profile 8 content:
> "Can be played by a decoder system that is unaware of Dolby Vision using only the base layer"
[...]

PR #3344, opened 2026-07-27
BugFix: Fix DV playback if SDR fallback is added to the manifest

ybai001:
The partial fix in aa5b318 moved HDR and codec score preferences into quality preferences and resolved fallback MIME types, but testing indicates there is still SDR track instead of Dolby Vision track selected on some TV devices.

Original issue: https://github.com/androidx/media/issues/3135
Related-commit: aa5b318406d40de7fdb31dabc3802b4e1624f710

This change fixes the remaining portion of #3135 not addressed by aa5b318. (The details can be found in inline comments.)

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/DefaultTrackSelector.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/DefaultTrackSelector.java
@@ -3937,7 +3937,7 @@ public class DefaultTrackSelector extends MappingTrackSelector
               == RendererCapabilities.HARDWARE_ACCELERATION_SUPPORTED;
       this.resolvedMimeType = resolvedMimeType;
       codecPreferenceScore = getVideoCodecPreferenceScore(resolvedMimeType);
-      isHdr = usesPrimaryDecoder && ColorInfo.isTransferHdr(format.colorInfo);
+      isHdr = usesPrimaryOrFallbackDecoder && ColorInfo.isTransferHdr(format.colorInfo);
       selectionEligibility = evaluateSelectionEligibility(formatSupport, requiredAdaptiveSupport);
     }
 
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/video/MediaCodecVideoRenderer.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/video/MediaCodecVideoRenderer.java
@@ -940,8 +940,13 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer
         return alternativeDecoderInfos;
       }
     }
-    return MediaCodecUtil.getDecoderInfosSoftMatch(
-        mediaCodecSelector, format, requiresSecureDecoder, requiresTunnelingDecoder);
+    if (MimeTypes.VIDEO_DOLBY_VISION.equals(format.sampleMimeType)) {
+      return MediaCodecUtil.getDecoderInfosSoftMatchFilteredByFormatSupport(
+          context, mediaCodecSelector, format, requiresSecureDecoder, requiresTunnelingDecoder);
+    } else {
+      return MediaCodecUtil.getDecoderInfosSoftMatch(
+          mediaCodecSelector, format, requiresSecureDecoder, requiresTunnelingDecoder);
+    }
   }
 
   @RequiresApi(26)
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecUtil.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecUtil.java
@@ -223,6 +223,55 @@ public final class MediaCodecUtil {
   }
 
   /**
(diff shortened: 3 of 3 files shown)


libraries/exoplayer/src/main/java/androidx/media3/exoplayer/trackselection/DefaultTrackSelector.java
```diff
@@ -3937,7 +3937,7 @@ public VideoTrackInfo(
               == RendererCapabilities.HARDWARE_ACCELERATION_SUPPORTED;
       this.resolvedMimeType = resolvedMimeType;
       codecPreferenceScore = getVideoCodecPreferenceScore(resolvedMimeType);
-      isHdr = usesPrimaryDecoder && ColorInfo.isTransferHdr(format.colorInfo);
+      isHdr = usesPrimaryOrFallbackDecoder && ColorInfo.isTransferHdr(format.colorInfo);
```
ybai001, line 3940: The fallback decoder could be HDR too. For example,
- Dolby Vision profile 8.1 content includes a HDR10 compatible HEVC base layer
- Dolby Vision profile 8.4 content includes a HLG compatible HEVC base layer
- Dolby Vision profile 10.1 content includes a HDR10 compatible AV1 base layer

If the target device is not Dolby Vision licensed, HEVC/HEVC/AV1 decoder will be primary decoder. And the corresponding base layer content is still HDR content.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/video/MediaCodecVideoRenderer.java
```diff
@@ -940,8 +940,13 @@ private static List<MediaCodecInfo> getDecoderInfos(
         return alternativeDecoderInfos;
       }
     }
-    return MediaCodecUtil.getDecoderInfosSoftMatch(
-        mediaCodecSelector, format, requiresSecureDecoder, requiresTunnelingDecoder);
+    if (MimeTypes.VIDEO_DOLBY_VISION.equals(format.sampleMimeType)) {
+      return MediaCodecUtil.getDecoderInfosSoftMatchFilteredByFormatSupport(
```
ybai001, line 944: We replace getDecoderInfosSoftMatch() with getDecoderInfosSoftMatchFilteredByFormatSupport() to resolve below two issues:

1. Separate OMX Dolby Vision decoders per profile
On some old Dolby Vision licensed devices (e.g. Philips 55OLED910/12), a dedicated OMX Dolby Vision decoder is for each DV profile (e.g. three decoders to support profile 5/8.x/9.x respectively). Assume the played content includes two tracks in manifest: one is DV P8.4, the other is SDR. If we're fortunate, the decoder supporting DV Profile 8.X appears first in the decoder list. In that case, it is treated as the preferred decoder, and Dolby Vision is played back on the device. But, if the first decoder in the list supports a different Dolby Vision profile (e.g. DV P5) than the one being played, it gets marked as isPreferred = false. So it receives a lower score in the Track Selector than an SDR decoder. This leads to incorrect track selection, where SDR is preferred over Dolby Vision even though the device fully supports the Dolby Vision profile being played. 

 
2.  On the device, there is a single Dolby Vision decoder that supports both profile 5 and profile 8.x (HEVC based). The problem appears with Profile 10 (AV1 based). For example, consider a manifest containing P10.1 + SDR on a Dolby Vision-enabled device that supports P5, P8.x but not for P10.x. At the same time, this device also includes an AV1 decoder. With the current implementation, the SDR track will be selected. This happens because the Dolby Vision decoder is included in decoderInfos even though it does not support the requested Dolby Vision profile (P10.1), since the check is performed only against the MIME type. As a result, when such a decoder appears first in the list and is evaluated through isFormatSupported(), it reports that the format is not supported (i.e. it supports only p5 and p8.x). This triggers a decoder fallback to the AV1 decoder, but it also causes the decoder to be marked as isPreferred = false (which is wrong because the DV decoder shouldn't be on the list). For devices that do not support the requested Dolby Vision profile, the cross-compatible base layer decoder (AV1) should still be considered the preferred decoder - this is the same behavior that we can see on non Dolby Vision devices today.

libraries/exoplayer/src/main/java/androidx/media3/exoplayer/video/MediaCodecVideoRenderer.java
```diff
@@ -940,8 +940,13 @@ private static List<MediaCodecInfo> getDecoderInfos(
...
+          context, mediaCodecSelector, format, requiresSecureDecoder, requiresTunnelingDecoder);
+    } else {
```
ybai001, line 946: To minimize the risk, we just apply our new method MediaCodecUtil.getDecoderInfosSoftMatchFilteredByFormatSupport() to Dolby Vision only. In theory, we think this method should be applied to other formats too.

ybai001:
Hi, @microkatz, could you help to review this PR? I understood you were busy on v1.11.0 release in the past month and had already helped us to merge 2 PRs. If there is anything unclear, please feel free to let me know. Thanks in advance.

microkatz:
@ybai001 

Thanks for checking in. We're working on it. This CL is currently in review.

PR #3352, opened 2026-07-29
Recover from non-monotonic presentationTimeUs caused by naively-muxed B-frame content

### LINKED ISSUE #3347: c2.mtk.hevc.decoder (MediaTek MT8696 / Google TV Streamer 4K) delivers ~30% of frames 30-130ms late on specific x265-encoded HEVC content, causing sustained stutter; VLC on identical hardware/content is unaffected
### Version

Media3 1.10.1 (built and tested via the `demo` module directly from the `androidx/media` `release` branch, and independently confirmed in Plex, Jellyfin (Android TV), and Plezzy — all separate ExoPlayer/Media3 integrations, ruling out any single app's integration code as the cause).

### More version details

**Device:** Google TV Streamer 4K ("kirkwood"), Android 14, MediaTek MT8696 SoC (PowerVR Rogue GE9215 GPU).

**Symptom:** playing specific HEVC (H.265) Main10 content causes sustained, severe visible stutter — not a one-time hiccup, but continuous throughout playback for as long as the affected content plays.

**Root cause, precisely measured:** we instrumented `MediaCodecVideoRenderer` directly (custom build, logging every frame's `earlyUs` value at the point of the drop/render decision) and found that **~30% of decoded video frames from `c2.mtk.hevc.decoder` arrive between 31ms and 131ms late** relative to their presentation deadline (median ~70ms late; at 23.976fps one frame period is ~42ms, so this is roughly 1-3 frame periods behind schedule). This is a **bounded, repeating per-frame latency pattern** — not a runaway drift into seconds, and not correlated with the `popFromStashAndRegister` log line described below.

Two log signatures accompany playback, though we now believe only the first is the actual cause and the second is a red herring:
1. **The actual cause — decode/delivery latency**, directly measured via instrumented logging (see above). Dropped-frame sample (`earlyUs` in microseconds, negative = late): `-130838, -129795, -128975, -128942, -128582, ..., -46831, -46611, -41407, -37913, -30670`. Full distribution: 41 frames <-100ms late, 15 in [-100,-70)ms, 42 in [-70,-40)ms, 17 in [-40,0)ms (n=115 of 360 total frame outcomes in this test run, 31.9% — matches the originally-reported ~32% independently).
2. **A red herring we chased for a long time**: a `CCodecBufferChannel`/`CCodecBuffers` "stale input buffer" + `popFromStashAndRegister: at 999999916585us` (implausible future timestamp) log pair on decoder initialization. **This log pair appears on every single test we ran, including files/synthetic content that played back with zero drops** — so it cannot be the actual cause, despite superficially looking like the smoking gun early in our investigation. We mention it here so other reporters searching for this exact log text land on the correct diagnosis (the real cause is decode latency, not this log line) rather than chasing it as we initially did.

We confirmed **threshold-tuning does not fix this** — we patched `MediaCodecVideoRenderer.shouldDropOutputBuffer`'s threshold from the default `-30ms` up to `-500ms` (very lenient) and rebuilt/retested: on short (10-15s) test clips this produced zero logged `droppedFrames` events, which looked like a fix — but on a longer, representative 90-second real clip of the same content, **zero drops were logged yet the stutter was still clearly visible**. Loosening the threshold doesn't fix anything real: it just changes the failure mode from "drop the late frame" (visible as a skip) to "render the late frame anyway" (visible as a hold/judder) — a frame delivered 70-130ms late is visually disruptive either way at this frame rate. **This rules out any simple ExoPlayer-side threshold/config fix.**

**Critically, VLC on the exact same device plays the identical content with no visible stutter** — and this is not simply VLC "tolerating" lateness better: we checked VLC's own late-frame-drop logic (`src/video_output/video_output.c`), and its threshold (~one frame period, ~42ms, when frame rate is known) is actually **stricter** than the worst lateness we measured in ExoPlayer (131ms). If VLC's decode path were producing the same late frames, VLC's own drop logic would react to them too. We confirmed this directly: we installed VLC-Android and captured a full `adb logcat` spanning an entire 90-second real-content playback of the same underlying source movie — **zero "picture is too late to be displayed" warnings, zero drop-related log lines of any kind, for the full 90 seconds**, with visually smooth playback confirmed throughout. So **VLC's decode is not falling behind in the first place** — something about how VLC drives this decoder avoids the latency spike, rather than merely tolerating it.

We tested three concrete, source-motivated hypotheses for what VLC does differently, all disproven by direct on-device A/B testing against the same 90-second real-content clip:
1. **Force the synchronous `MediaCodecAdapter`** (`DefaultRenderersFactory.forceDisableMediaCodecAsynchronousQueueing()`), matching VLC's exclusive use of the Android NDK's synchronous `AMediaCodec_dequeueInputBuffer`/`dequeueOutputBuffer` polling API (VLC never uses the async callback API anywhere in its Android backend — confirmed by reading `modules/codec/omxil/mediacodec_ndk.c`/`mediacodec_jni.c`) — **no change**, identical drop pattern. (Both sync and async adapters sit on top of the same native `CCodecBufferChannel` plumbing, so this was always a long shot in retrospect.)
2. **Loosen the drop threshold** from -30ms to -500ms — eliminated logged drops on short clips but **stutter remained clearly visible on the 90s clip with zero drops logged**, proving this only changes which visual artifact appears (skip vs. hold), not the underlying lateness.
3. **Force `mediaFormat.setFeatureEnabled(CodecCapabilities.FEATURE_AdaptivePlayback, true)`** on the codec configuration, matching VLC's unconditional `feature-adaptive-playback=1` (confirmed via VLC's own log line `mediacodec configured for adaptative playback` — VLC sets this for essentially all Surface-based playback, while ExoPlayer only requests equivalent `KEY_MAX_WIDTH`/`KEY_MAX_HEIGHT` headroom for genuinely multi-format adaptive streams, not a single local file) — **no change**, identical drop pattern.

We're continuing to investigate (next candidate: input-buffer queuing depth/pipelining — VLC's decoder-output draining runs on a dedicated thread decoupled from input queueing, which may let more compressed frames be queued ahead of the decoder than ExoPlayer's tighter render-loop-coupled pacing) and will update this issue if we isolate the actual mechanism. Given three plausible, testable hypotheses have now been ruled out, it's also plausible the real difference is inside MediaTek's closed-source Codec2 HAL in a way that isn't reachable from app-level MediaCodec configuration at all — in which case this may need MediaTek's own engineers to resolve, and this report may be primarily useful as precise symptom documentation rather than pointing to a specific ExoPlayer code fix. Note that ExoPlayer already has a `c2.mtk.hevc.decoder`/`c2.mtk.avc.decoder`-specific workaround elsewhere in the codebase (`shouldReleaseCodecInsteadOfFlushing`), confirming this decoder family is already known to the maintainers as needing special-casing.

**Content dependency, precisely isolated:** the bug is tied to output from **x265 encoder builds around version 3.3** (~2020-era) specifically, not to file content, resolution, or container metadata:
- Ruled out: `hev1` vs `hvc1` FourCC tag, non-standard/cropped resolution + conformance-window offsets (reproduces identically on a *second*, unrelated file with a clean macroblock-aligned resolution needing zero cropping), GOP/keyframe interval, container timescale precision, SAO on/off, and general "visual complexity" (tested flat synthetic patterns, fractal motion, synthetic film-grain noise — none reproduced it).
- A byte-level VPS/SPS/PPS bitstream parse confirmed real bug-triggering files and non-triggering synthetic files were **byte-identical** on every SPS/PPS field checked (profile/tier/level, chroma format, bit depth, conformance window, VUI timing, PPS/VPS bytes).
- The isolating experiment: building **x265 from source at git tag `3.3`** and encoding real, non-copyrighted Creative Commons footage with it reproduces the bug; the same footage encoded with current x265 4.1 (even with closely matched encoder settings) does not. This points to something in the actual encoded coefficient/motion-vector/mode-decision data produced by that encoder generation — not any inspectable high-level setting — as what triggers the decoder's slow path.

Google TV Streamer 4K (MediaTek MT8696), Android 14.

Not tested on other devices. VLC (non-ExoPlayer) on the *same* device/content is unaffected — see analysis above.

**Yes** — confirmed directly in the official Media3 demo app (`demo` module, built from source, `androidx.media3.demo.main.PlayerActivity`, launched via `am start -a androidx.media3.demo.main.action.VIEW -d file://... --es mime_type video/mp4`), with no third-party app involved.

### Reproduction steps

1. Play the sample file below on a Google TV Streamer 4K via the Media3 demo app (or any ExoPlayer/Media3-based app).
2. Observe sustained visible stutter throughout playback.
3. `adb logcat` shows `EventLogger: droppedFrames` events recurring continuously (roughly every ~7s in batches, ~30% of frames total) for the duration of playback.
4. For comparison: play the same file in VLC on the same device — no stutter, no drops.

### Sample file

**https://github.com/2bitoperations/google-tv-streamer-hevc-decoder-repro** — a 15-second HEVC Main10 MP4 (`tos_x265_3.3_final.mp4`, ~2.3MB, 1916x816, `hvc1` tag, 24fps, AAC 5.1) that reliably reproduces the issue.

This sample is **entirely legally clean**: the source footage is Blender Foundation's *Tears of Steel*, licensed CC BY 3.0, encoded using x265 (open source) built from the official source tag `3.3`. No copyrighted commercial content is involved anywhere in the pipeline. The linked repo's README also documents the exact, fully reproducible command pipeline used to build the sample from public sources, in case it's useful for building further variants.

Full `adb logcat` capture and the raw per-frame `earlyUs` instrumentation log are available on request.

2bitoperations:
Fixes #3347

## Problem

On Google TV Streamer 4K (MediaTek MT8696, `c2.mtk.hevc.decoder`),
decoding specific HEVC Main10 content causes the codec to return
output buffers with non-monotonic `presentationTimeUs`, while the
order in which those buffers are released remains correct — confirmed
by comparing each buffer's release-order rank against its
timestamp-sorted rank: the two never diverge by more than a few
positions, and the timestamp values themselves are individually
correct for the frame they belong to.

Reproduced independently of ExoPlayer, using a minimal test harness
built directly on `AMediaCodec`/`AMediaExtractor` (no ExoPlayer, no
Java `MediaCodec`) against the same file: identical non-monotonic
timestamp pattern. This confirms the defect is in the decoder/HAL
output, not in ExoPlayer's handling of it.

`MediaCodecRenderer` processes output buffers in raw release order
but uses the codec-reported `presentationTimeUs` directly for the
render/drop decision in `MediaCodecVideoRenderer`. When a buffer's
timestamp is behind the previous one, it's treated as arriving too
late and dropped. Measured on affected content: ~30% of decoded video
frames dropped, sustained throughout playback, reproduced in the
Media3 demo app as well as Plex, Jellyfin, and other ExoPlayer-based
players on the same device/file.

## Fix

Re-derive each output buffer's `presentationTimeUs` from its release
order and the stream's known frame duration (from `Format.frameRate`),
instead of trusting the codec-reported value. Buffer release order is
left untouched — buffers are not held or reordered.

## Testing

Verified on the affected device (Google TV Streamer 4K) against three
files that reliably reproduce the issue on unpatched `main`,
including a 90-second real-world clip: dropped-frame count goes from
~30% to 0, and the fix was visually confirmed on-device — direct
observation of smooth playback, not just log metrics. HOWEVER, with this approach, video and audio lose sync.

An earlier approach that buffered and released output buffers in
sorted-timestamp order was also tried and rejected: it also eliminated
logged drops, but produced visible playback artifacts on-device. That
result is why this PR does not reorder buffers.

### DIFF
--- a/libraries/common/src/main/java/androidx/media3/common/Format.java
+++ b/libraries/common/src/main/java/androidx/media3/common/Format.java
@@ -172,6 +172,7 @@ public final class Format {
     @Nullable private DrmInitData drmInitData;
     private long subsampleOffsetUs;
     private boolean hasPrerollSamples;
+    private boolean hasReliablePresentationTimestamps;
 
     // Video specific.
 
@@ -218,6 +219,10 @@ public final class Format {
       maxInputSize = NO_VALUE;
       maxNumReorderSamples = NO_VALUE;
       subsampleOffsetUs = OFFSET_SAMPLE_RELATIVE;
+      // Default to true: nearly every extractor/format provides trustworthy per-sample
+      // presentation timestamps. Only set to false where the source is known not to (see
+      // Format#hasReliablePresentationTimestamps).
+      hasReliablePresentationTimestamps = true;
       // Video specific.
       width = NO_VALUE;
       height = NO_VALUE;
@@ -270,6 +275,7 @@ public final class Format {
       this.drmInitData = format.drmInitData;
       this.subsampleOffsetUs = format.subsampleOffsetUs;
       this.hasPrerollSamples = format.hasPrerollSamples;
+      this.hasReliablePresentationTimestamps = format.hasReliablePresentationTimestamps;
       // Video specific.
       this.width = format.width;
       this.height = format.height;
@@ -587,6 +593,20 @@ public final class Format {
     @CanIgnoreReturnValue
     public Builder setHasPrerollSamples(boolean hasPrerollSamples) {
       this.hasPrerollSamples = hasPrerollSamples;
+      return this;
+    }
+
+    /**
+     * Sets {@link Format#hasReliablePresentationTimestamps}. The default value is {@code true}.
+     *
+     * @param hasReliablePresentationTimestamps The {@link
+     *     Format#hasReliablePresentationTimestamps}.
+     * @return The builder.
+     */
+    @CanIgnoreReturnValue
+    public Builder setHasReliablePresentationTimestamps(
+        boolean hasReliablePresentationTimestamps) {
+      this.hasReliablePresentationTimestamps = hasReliablePresentationTimestamps;
       return this;
     }
 
@@ -1051,6 +1071,19 @@ public final class Format {
    */
   @UnstableApi public final boolean hasPrerollSamples;
 
+  /**
+   * Indicates whether per-sample presentation timestamps for this track are known to be
+   * accurate.
+   *
+   * <p>When {@code false}, the source (typically a container demuxer) could not determine each
+   * sample's true composition/presentation time — for example an MP4 track with B-frame
+   * reordering but no {@code ctts} box, where every sample's timestamp collapses to its decode
+   * time. Renderers should not treat such timestamps as ground truth for scheduling decisions
+   * (e.g. dropping "late" output buffers); doing so can misread decoder-level reordering as
+   * lateness. Defaults to {@code true}.
+   */
+  @UnstableApi public final boolean hasReliablePresentationTimestamps;
+
   // Video specific.
 
   /** The width of the video in pixels, or {@link #NO_VALUE} if unknown or not applicable. */
@@ -1218,6 +1251,7 @@ public final class Format {
     drmInitData = builder.drmInitData;
     subsampleOffsetUs = builder.subsampleOffsetUs;
     hasPrerollSamples = builder.hasPrerollSamples;
+    hasReliablePresentationTimestamps = builder.hasReliablePresentationTimestamps;
     // Video specific.
     width = builder.width;
     height = builder.height;
--- a/libraries/common/src/main/java/androidx/media3/common/Format.java
+++ b/libraries/common/src/main/java/androidx/media3/common/Format.java
@@ -219,10 +219,7 @@ public final class Format {
       maxInputSize = NO_VALUE;
       maxNumReorderSamples = NO_VALUE;
       subsampleOffsetUs = OFFSET_SAMPLE_RELATIVE;
-      // Default to true: nearly every extractor/format provides trustworthy per-sample
-      // presentation timestamps. Only set to false where the source is known not to (see
-      // Format#hasReliablePresentationTimestamps).
-      hasReliablePresentationTimestamps = true;
+      hasReliablePresentationTimestamps = true; // most formats have accurate per-sample timestamps
       // Video specific.
       width = NO_VALUE;
       height = NO_VALUE;
@@ -1072,15 +1069,11 @@ public final class Format {
   @UnstableApi public final boolean hasPrerollSamples;
 
   /**
-   * Indicates whether per-sample presentation timestamps for this track are known to be
-   * accurate.
+   * Indicates whether per-sample presentation timestamps for this track are known to be accurate.
    *
-   * <p>When {@code false}, the source (typically a container demuxer) could not determine each
-   * sample's true composition/presentation time — for example an MP4 track with B-frame
-   * reordering but no {@code ctts} box, where every sample's timestamp collapses to its decode
-   * time. Renderers should not treat such timestamps as ground truth for scheduling decisions
-   * (e.g. dropping "late" output buffers); doing so can misread decoder-level reordering as
-   * lateness. Defaults to {@code true}.
+   * <p>When {@code false}, the source couldn't determine each sample's true presentation time
+   * (for example an MP4 track with no {@code ctts} box). Consumers shouldn't treat the
+   * timestamps as ground truth for scheduling decisions. Defaults to {@code true}.
    */
   @UnstableApi public final boolean hasReliablePresentationTimestamps;
 
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecRenderer.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/mediacodec/MediaCodecRenderer.java
@@ -422,18 +422,18 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
   // B-frame reordering but no ctts box, where every sample's presentationTimeUs is just its
   // decode time. MediaCodec faithfully echoes back whatever (unreliable) timestamp it was
   // queued with, so the codec's raw output can look non-monotonic even though its buffer
-  // *release order* is correct. Re-deriving each buffer's timestamp from its release order and
-  // the stream's known frame duration corrects this without reordering or holding buffers — this
-  // is a best-effort reconstruction for a source that has already told us it doesn't know the
-  // real answer, not a workaround for a specific decoder defect (see
-  // https://github.com/androidx/media/issues/3347).
-  // Kept unrounded and only rounded per-frame at the point of use (see below) — rounding this
-  // once and multiplying by a growing frame index would accumulate error linearly over the
-  // life of the stream (e.g. ~0.33us/frame for 24000/1001fps content is only 0.7ms over a 90s
-  // clip, but ~58ms over a 2-hour movie).
-  private double arrivalOrderPtsFrameDurationUs = -1;
-  private long arrivalOrderPtsBaseUs = C.TIME_UNSET;
-  private long arrivalOrderPtsFrameIndex;
+  // *release order* is correct.
+  //
+  // Re-associating each dequeued output buffer with the presentationTimeUs of the next
+  // input buffer *in queue order* (not the codec's own echoed value) corrects this without
+  // reordering or holding buffers. This mirrors what a completely independent player (VLC, via
+  // its MP4 demuxer + a FIFO of queued-but-not-yet-consumed timestamps) already does for exactly
+  // this situation. Queue order is always monotonic by construction (it's a direct echo of the
+  // container's decode-time-to-sample table), and — importantly — this preserves each sample's
+  // *real* duration exactly as declared by the container: unlike re-deriving from a single
+  // assumed constant frame rate, this does not corrupt genuinely variable-frame-rate content
+  // (where Format#frameRate is at best an average, not a true per-sample value).
+  private final ArrayDeque<Long> arrivalOrderPtsQueue = new ArrayDeque<>();
 
   /**
    * @param context A context.
@@ -1129,8 +1129,7 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
         codecReconfigured ? RECONFIGURATION_STATE_WRITE_PENDING : RECONFIGURATION_STATE_NONE;
     hasSkippedFlushAndWaitingForQueueInputBuffer = false;
     skippedFlushOffsetUs = 0;
-    arrivalOrderPtsBaseUs = C.TIME_UNSET;
-    arrivalOrderPtsFrameIndex = 0;
+    arrivalOrderPtsQueue.clear();
   }
 
   /**
@@ -1668,6 +1667,14 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
       hasSkippedFlushAndWaitingForQueueInputBuffer = false;
     }
 
+    if (getTrackType() == C.TRACK_TYPE_VIDEO
+        && inputFormat != null
+        && !inputFormat.hasReliablePresentationTimestamps) {
+      // Recorded in the same (pre skippedFlushOffsetUs) scale that drainOutputBuffer converts
+      // dequeued output timestamps back to, so it can be substituted directly for the codec's
+      // own echoed value there. See arrivalOrderPtsQueue's declaration for why.
+      arrivalOrderPtsQueue.addLast(presentationTimeUs);
+    }
     onQueueInputBuffer(buffer);
     int flags = getCodecBufferFlags(buffer);
     presentationTimeUs += skippedFlushOffsetUs;
@@ -2277,24 +2284,13 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
       if (getTrackType() == C.TRACK_TYPE_VIDEO
           && inputFormat != null
           && !inputFormat.hasReliablePresentationTimestamps
-          && inputFormat.frameRate != Format.NO_VALUE
-          && inputFormat.frameRate > 0) {
-        if (arrivalOrderPtsFrameDurationUs < 0) {
-          arrivalOrderPtsFrameDurationUs = 1_000_000.0 / inputFormat.frameRate;
-        }
-        if (arrivalOrderPtsBaseUs == C.TIME_UNSET) {
-          // Anchor to the first real output buffer's own reported timestamp; only the spacing
-          // of subsequent buffers is re-derived, not the absolute position in the stream.
-          arrivalOrderPtsBaseUs = outputBufferInfo.presentationTimeUs;
-          arrivalOrderPtsFrameIndex = 0;
-        } else {
-          arrivalOrderPtsFrameIndex++;
-          // Round only the final offset, not the per-frame duration, so rounding error can't
-          // accumulate across the length of the stream.
-          outputBufferInfo.presentationTimeUs =
-              arrivalOrderPtsBaseUs
-                  + Math.round(arrivalOrderPtsFrameIndex * arrivalOrderPtsFrameDurationUs);
-        }
+          && !arrivalOrderPtsQueue.isEmpty()) {
+        // Replace the codec's own (unreliable, possibly-reordered) echoed timestamp with the
+        // presentationTimeUs of the next sample in queue order. Queue order is a direct echo of
+        // the container's decode-time-to-sample table, so it's always monotonic and always
+        // reflects each sample's true declared duration — including on genuinely
+        // variable-frame-rate content, where a single Format#frameRate value would not.
+        outputBufferInfo.presentationTimeUs = arrivalOrderPtsQueue.removeFirst();
       }
 
       this.outputIndex = outputIndex;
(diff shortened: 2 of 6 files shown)


2bitoperations:
Pushed a fix for a second bug: audio/video would slowly drift out of
sync over long playback. Cause was an accumulating rounding error in
the per-frame timestamp math — invisible on a 90s test clip, ~58ms by
the end of a 2-hour movie. Fixed by rounding only the final offset,
not the per-frame duration.

Tested in two real players on the affected hardware (Jellyfin Android
TV, and Plezy). 40 minutes of real playback, no drift, no stutter.

FongMi:
Thanks for the detailed investigation and for providing a legal,
reproducible sample. I agree that changing the late-frame threshold
does not address the underlying problem.

However, I don't think the current evidence isolates this as a
MediaTek decoder/HAL defect yet. The repro MP4 itself appears to have
a questionable presentation timeline for reordered frames.

I downloaded `tos_x265_3.3_final.mp4` and inspected it with ffprobe.
All 360 video packets have `PTS == DTS`, even though the HEVC stream
uses B-frames and B-pyramid. For example:

```text
packet PTS/DTS:
-0.080000 / -0.080000
-0.041667 / -0.041667
 0.000000 /  0.000000
 0.041667 /  0.041667
...
```

After decoding, the propagated frame PTS becomes non-monotonic:

```text
frame PTS:
0.041667
0.083333
0.000000
0.125000
0.250000
0.291667
0.208333
...
```

while FFmpeg's `best_effort_timestamp` reconstructs approximately:

```text
0.041667
0.083333
0.125000
0.166667
0.250000
0.291667
0.333333
...
```

This looks consistent with missing or incorrect composition offsets
in the MP4, possibly introduced by the raw-HEVC stream-copy muxing
pipeline in the repro README.

Android documents `BufferInfo.presentationTimeUs` as being derived
from the timestamp queued with the corresponding input buffer:

https://developer.android.com/reference/android/media/MediaCodec.BufferInfo#presentationTimeUs

Therefore, reproducing the same output timestamp sequence with
AMediaCodec/AMediaExtractor proves that ExoPlayer did not introduce
the sequence, but it does not by itself prove that the decoder/HAL is
behaving incorrectly. It may simply be propagating input timestamps
that do not describe the HEVC presentation order correctly.

Could we rule this out before changing `MediaCodecRenderer`?

1. Generate the same encoded content in a container with correct
   composition timestamps and verify that some packet PTS values
   differ from DTS where reordering requires it.
2. Log each timestamp queued into MediaCodec and its corresponding
   output `BufferInfo.presentationTimeUs`.
3. Compare packet PTS/DTS for the affected x265 3.3 file and the
   unaffected x265 4.1 file.

There is also a separate scope concern with the current patch: despite
being motivated by one MTK decoder, it rewrites timestamps for every
video decoder whenever `Format.frameRate` is known. This would replace
valid VFR timing, intentional gaps, duplicate frames, and format
transitions with a synthetic CFR timeline. The frame duration is also
not reset when the input frame rate changes.

The patch explains why this particular CFR sample becomes smooth, but
I think the source timeline needs to be ruled out before treating it
as a general Media3 renderer fix. If a player-side workaround is still
needed afterward, it should probably be gated to the affected codec
and device, activated only after detecting the known timestamp pattern,
and kept out of the generic `MediaCodecRenderer` path.

2bitoperations:
## Summary

MANY thanks for your helpful reply. We looked into all three of your suggestions. Short version: I no longer think this is a
MediaTek decoder defect. It's a bug in the reproduction repo's own muxing pipeline (which reflects bugs in the muxing pipelines of a bunch of files running around,) and as far as
we can tell it's independent of x265 version, encoder options (beyond whether B-frames are used
at all), and source content.

## 1. Composition timestamps: correctly-muxed vs. naively-muxed content

You were right about the repro file. Reproduced your `ffprobe` findings exactly on
`tos_x265_3.3_final.mp4`: every packet has `PTS == DTS` (0/360), and the propagated frame PTS is
non-monotonic in the pattern you posted, while `best_effort_timestamp` reconstructs a clean
sequence.

The repro README's mux step (`ffmpeg -fflags +genpts -r 24 -i raw.hevc ... -c:v copy`)
stream-copies a *detached* raw HEVC stream, which carries no composition-time
information of its own. `+genpts` fills the gap by assigning `PTS = DTS = packet_index / fps` —
decode order — which is only correct when there's no B-frame reordering to begin with. Minimal
confirmation, x265 4.1, defaults, nothing else changed:

```bash
ffmpeg -f lavfi -i "testsrc2=size=640x360:rate=24:duration=15" -f yuv4mpegpipe src.y4m

# encode straight to MP4 — libavformat assigns real composition offsets in-process
ffmpeg -i src.y4m -c:v libx265 -tag:v hvc1 good.mp4

# encode to a detached raw stream, then remux exactly like the repro README does
ffmpeg -i src.y4m -c:v libx265 -f hevc raw.hevc
ffmpeg -fflags +genpts -r 24 -i raw.hevc -c:v copy -tag:v hvc1 bad.mp4
```
```
good.mp4:  PTS != DTS on 270/360 packets  (correctly timed)
bad.mp4:   PTS != DTS on   0/360 packets  (every packet's PTS is just its DTS)
```
Same encoder, same options, same source — only the mux step differs.

Confirmed on-device too, on a build with none of our patches applied: correctly-muxed B-frame
content plays clean; content muxed the repro's way drops ~30% of frames, consistently, across
every file we've tried.

## 2. x265 3.3 vs. current x265 4.1

Re-encoded the same Tears of Steel source with current x265 (4.1+1-1d117be), options matched as
closely as possible, muxed with the *unmodified* repro pipeline. Result: indistinguishable from
the 3.3 encode — 0/360 `PTS != DTS`, same broken pattern.

This briefly looked like it contradicted our own earlier finding that x265 4.1 content didn't
reproduce the bug, so we ran it down rather than gloss over it. The earlier "unaffected" 4.1 file
has real composition offsets (`PTS != DTS` on 243/360 packets) — it was muxed correctly, almost
certainly via a single-step `ffmpeg -c:v libx265 ... output.mp4` encode, not the raw-stream-then-
remux process the repro pipeline uses. That two-step process isn't optional for x265 3.3: its CLI
has no muxer at all (`-o` is raw-bitstream-only). Confirmed both ends on-device: the original file
still plays with zero dropped frames; a freshly built, zero-options x265 4.1 file put through the
naive remux drops 682/2280 frames (29.9%). Same decoder, same device, same encoder version — only
the mux method differs.

So: not a contradiction, and not an x265-version-specific bug. Encoder version is irrelevant to
the trigger; whether the file went through a proper muxer is the entire story.

## 3. Logging queued input vs. dequeued output timestamps

Patch (diagnostic-only, no behavior change):
[`6fc70188e49667a894b44ada761193e28ebda8b9`](https://github.com/2bitoperations/media/commit/6fc70188e49667a894b44ada761193e28ebda8b9)
on our fork, based on the clean `release` branch (i.e. without our reorder/relabel patch from this
PR applied — this measures the raw, unpatched decoder). Logs the timestamp on every
`queueInputBuffer`/`queueSecureInputBuffer` call and every buffer dequeued from
`drainOutputBuffer`. Built, installed on the real device, captured against
`tos_x265_3.3_final.mp4` and the 4.1 re-encode:

| File | queued (IN) | IN non-monotonic | dequeued (OUT) | OUT non-monotonic |
|---|---|---|---|---|
| 3.x repro | 291 | 0 | 276 | 117 |
| 4.x re-encode | 284 | 0 | 270 | 114 |

Queued-input order is perfectly monotonic in both — a direct echo of the container's decode-order
timestamps. Dequeued-output order is non-monotonic in both, at essentially the rate the container
alone predicted, with matching non-monotonic transition offsets between the two files.

The key check: **every single dequeued output timestamp, in both files, is one of the exact
values that was queued as input — zero exceptions.** `c2.mtk.hevc.decoder` never emits a
timestamp it wasn't given. Direct on-device confirmation of the `presentationTimeUs` contract you
cited — the decoder echoes back exactly what it was handed; the non-monotonicity is already
present in what it was handed.

## Where this leaves things

I don't think this is evidence of a MediaTek decoder defect. However, what it *does* look like is
that files encoded with the x265 3.x series are likely to actually hit this in practice: its CLI
has no muxer of its own, so any pipeline built around it needs a separate remux step, and if that
step doesn't compute real composition offsets — ours doesn't, and we don't know what tooling
popularly shared media pipelines typically used either — this is exactly the result.

VLC plays these incorrectly-timestamped files just fine, so I want to dig into why. My guess is
it comes down to VLC's "if it's only a little late, just display it right now" approach versus
`shouldDropOutputBuffer`, which is closer to "if the presentation timestamp says it's late, chuck
that frame right in the garbage." I'll look into how VLC actually handles this (and crucially, how it is not overly permissive, as you have cautioned against) and follow up.

2bitoperations:
## Why VLC doesn't stutter on this content

Instrumented VLC itself (not just read its source) and ran it on-device against the same repro file. It receives *equally or more* disordered decoder output than ExoPlayer from `c2.mtk.hevc.decoder` — 52.2% non-monotonic steps in the raw echoed `presentationTimeUs`, vs. ~42% measured earlier from ExoPlayer, same 1×–4× frame-period jump signature. So it's not decoder behavior, and it's not a threading/queuing difference between the two clients.

VLC just never lets that value reach anything that schedules a frame. Its MP4 demuxer refuses to invent a PTS for a video track when the file has no `ctts` box — true of any file built via the raw-ES + `-fflags +genpts` pipeline this whole thread has been using:

```c
// mp4.c, https://code.videolan.org/videolan/vlc/-/blob/ac6c2a405d652b5576128ceb9fec2c342f0e83ec/modules/demux/mp4/mp4.c#L1391-L1396
if( MP4_TrackGetPTSDelta( p_demux, tk, &i_delta ) )
    p_block->i_pts = p_block->i_dts + i_delta;
else if( tk->fmt.i_cat != VIDEO_ES )
    p_block->i_pts = p_block->i_dts;
else
    p_block->i_pts = VLC_TICK_INVALID;   // video track, no ctts: don't guess
```

That `VLC_TICK_INVALID` trips a fallback in the MediaCodec wrapper that relabels every output picture by decode-order index instead of trusting the (disordered) decoder echo — confirmed on-device: 0/2280 non-monotonic after the fallback, vs. the 52.2% it replaced.

`androidx/media3`'s MP4 extractor has no "PTS unknown" state. When `ctts` is absent it silently produces `PTS = DTS`, because the offset term it would add just stays at its zero default:

```java
// BoxParser.java, https://github.com/androidx/media/blob/5fb306449733dd71595700c1227ad6087578c559/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java#L516
int timestampOffset = 0;
...
// https://github.com/androidx/media/blob/5fb306449733dd71595700c1227ad6087578c559/libraries/extractor/src/main/java/androidx/media3/extractor/mp4/BoxParser.java#L627
timestamps[i] = timestampTimeUnits + timestampOffset;  // ctts == null → always 0 → PTS = DTS
```

That definite-but-wrong PTS is exactly what `shouldDropOutputBuffer` (`MediaCodecVideoRenderer.java:2166`) later measures against the audio clock and drops.

So: not a MediaTek defect, not an ExoPlayer bug in the usual sense — one demuxer treats a missing `ctts` as "unknown," the other treats it as "zero," and that's the entire difference in outcome for this class of file.

Next: I'm going to look at overhauling this to give media3 a first-class notion of "missing PTS," and only fall back to today's spray-and-pray frame-dropping behavior in that specific case.

2bitoperations:
ok - i think this approach is way more contained and correct. @FongMi please let me know what you think!

FongMi:
This is much better scoped than the previous version, but I found several correctness issues that should be addressed before this is ready:

1. `hasReliablePresentationTimestamps` is only set near the final return path of `BoxParser.parseStbl()`. The no-edit-list, `omitTrackSampleTable`, zero-duration-edit, and other early-return paths bypass it. As a result, many video tracks without `ctts` will never activate the workaround.

2. Missing `ctts` alone is too broad a trigger. A correctly muxed video without frame reordering legitimately does not need `ctts`. Media3 already parses reordering information such as `Format.maxNumReorderSamples`; the fallback should also require evidence that the bitstream actually reorders frames.

3. Returning `false` unconditionally from both `shouldDropOutputBuffer()` and `shouldDropBuffersToKeyframe()` disables recovery from genuine decoder slowness. On an overloaded device, video may remain behind audio indefinitely.

4. The FIFO is gated using the current `inputFormat` when queuing and dequeuing, while the drop logic checks `codecInputFormat`. During seamless format changes these may describe different buffers, leaving stale FIFO entries or applying the wrong policy.

5. Tunneling needs explicit handling. In tunneling mode, decoded buffers are not returned to the renderer, so `arrivalOrderPtsQueue` will be filled but never drained, and the relabeling cannot take effect.

6. The VFR test without B-frames does not validate the difficult case. With no reordering, input and output order already match. A VFR stream with B-frame reordering and missing composition timestamps cannot have its original presentation timing reconstructed exactly.

Please also add focused tests covering:

- no `ctts` and no edit list;
- legitimate no-`ctts`, non-reordered video;
- reordered video without `ctts`;
- flush, seek, EOS, and seamless format changes;
- decoder input/output count mismatch;
- tunneling behavior.

The investigation and the move away from a synthetic CFR timeline are solid improvements, but the current implementation still has correctness and lifecycle gaps.

2bitoperations:
Addressed point 2 in `ce9dde0`: the trigger now also requires `track.format.maxNumReorderSamples != 0` (from the bitstream's own SPS, already parsed before this point), so a correctly-muxed non-reordering track no longer gets flagged just for lacking `ctts` — verified on-device that `good_95s.mp4` (`maxNumReorderSamples=0`) now correctly skips the workaround while `broken_95s.mp4` (`maxNumReorderSamples=2`) still triggers it. Added a synthetic MPEG-4 Part 2 test file exercising the conservative `NO_VALUE` case (a codec media3 never derives `maxNumReorderSamples` for at all) to the repro repo, with the exact generation steps: https://github.com/2bitoperations/google-tv-streamer-hevc-decoder-repro#second-file-mp4v_no_ctts_no_maxreordermp4

PR #3354, opened 2026-07-30
Preserve HDR bitmap in image playback/export pipeline

SathyaTadepalli:
Bug is https://partnerissuetracker.corp.google.com/issues/531143289

HardwareBufferFrameReader derives ColorInfo from the bitmap's ColorSpace (BT2020 HLG/PQ) instead of always emitting SRGB_BT709_FULL.
BitmapToHardwareBufferProcessor derives the HardwareBuffer pixel format from Bitmap.Config (RGBA_1010102 → RGBA_1010102, RGBA_F16 → RGBA_FP16, else RGBA_8888) so bit depth is preserved independently of
color space.
Prefers the Bitmap.copy(Config.HARDWARE) fast path for all configs. Falls back to CPU/JNI copy only when the HARDWARE path is unavailable or silently downgrades an HDR bitmap's bit depth.


shahdDaghash:
Thanks @SathyaTadepalli for making the new PR!

PR #3368, opened 2026-08-12
Support injectable charset detection for standalone SubRip subtitles

### LINKED ISSUE #2247: Enable heuristic character set detection for SRT subtitles
verion: 1.6.0-rc01

Two ways:

1. Using a local cheat, it appears as garbled code.
After transcoding the file stream to UTF-8, it can be displayed normally.
`Path.utf8(file)`

2. When adding subtitles online, it appears as garbled text.
`new MediaItem.SubtitleConfiguration.Builder(Uri.parse(getUrl())).setLabel(getName()).setMimeType(getFormat()).setSelectionFlags(getFlag()).setLanguage(getLang()).build();`

file : 

[【ATV】【大俠霍元甲】01.zip](https://github.com/user-attachments/files/19338318/ATV.01.zip)

hoyahozz:
## Summary

External standalone and raw-text SubRip inputs without a supported UTF byte order mark currently fall back to UTF-8. As a result, subtitles encoded with legacy character encodings such as GB18030, EUC-KR, or Shift_JIS may be decoded incorrectly.

This change introduces an injectable `CharsetDetector` contract in the extractor text package and passes it through `DefaultSubtitleParserFactory` to applicable `SubripParser` instances. Applications can provide their own charset detection implementation without requiring Media3 to depend on a specific detection library.

This follows the extension-point approach proposed in [google/ExoPlayer#3644](https://github.com/google/ExoPlayer/issues/3644#issuecomment-565503504).

## Behavior
- Existing behavior is unchanged when no detector is configured or when the detector returns `null`: a UTF-8 or UTF-16 byte order mark still wins, and everything else falls back to UTF-8.
- A byte order mark takes precedence over the detector, which is not called at all in that case.
- The detector is only used for SubRip `Format`s whose `containerMimeType` is null or a text MIME type, i.e. standalone `.srt` input. 
  - SubRip muxed into a container (Matroska/WebM) keeps the current UTF-8 behavior, because a single sample may hold only a small part of the subtitle stream.
- `CharsetDetector.detect()` receives exactly the `offset` and `length` passed to `SubtitleParser.parse()`. 
  - That range is not guaranteed to be a complete file, so implementations have to tolerate partial input.
- Scoping is done by `DefaultSubtitleParserFactory`. `SubripParser` itself always uses whatever detector it is constructed with.

This PR intentionally does not provide a charset detection implementation. Applications remain responsible for choosing and configuring one.

### Example - injecting a detector for an external SRT

The following example assumes that `appCharsetDetector` wraps a detection library selected by the application.

```java
CharsetDetector charsetDetector =
    (data, offset, length) -> appCharsetDetector.detect(data, offset, length);

DefaultMediaSourceFactory mediaSourceFactory =
    new DefaultMediaSourceFactory(context)
        .setSubtitleParserFactory(new DefaultSubtitleParserFactory(charsetDetector));

ExoPlayer player =
    new ExoPlayer.Builder(context)
        .setMediaSourceFactory(mediaSourceFactory)
        .build();

MediaItem.SubtitleConfiguration subtitleConfiguration =
    new MediaItem.SubtitleConfiguration.Builder(subtitleUri)
        .setMimeType(MimeTypes.APPLICATION_SUBRIP)
        .setLanguage("ko")
        .build();

MediaItem mediaItem =
    new MediaItem.Builder()
        .setUri(videoUri)
        .setSubtitleConfigurations(Collections.singletonList(subtitleConfiguration))
        .build();

player.setMediaItem(mediaItem);
player.prepare();
player.play();
```

When the external SRT is opened, `DefaultMediaSourceFactory` creates a standalone subtitle `Format`. `DefaultSubtitleParserFactory` then creates a `SubripParser` with the configured detector. If the input has no supported UTF BOM, the parser asks the detector to inspect the exact requested byte range and decodes that range using the returned charset.

The same detector is also used for raw-text SubRip inputs. It is not passed to SubRip samples embedded in Matroska or WebM containers.

## Before / After

The same BOM-free subtitle files were played on the same physical device. Without a detector, the existing UTF-8 fallback produces replacement characters. Injecting a detector that returns the matching charset restores the original Korean, Chinese, and Japanese text.

| Language and encoding | Before : no detector | After : matching detector |
| --- | --- | --- |
| Korean (`EUC-KR`) | 02-euc-kr-without-detector | 03-euc-kr-with-detector |
| Chinese (`GB18030`) | 05-zh-gb18030-before | 06-zh-gb18030-after |
| Japanese (`Shift_JIS`) | 07-ja-shiftjis-before | 08-ja-shiftjis-after  |

Fixes #2247

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -57,6 +57,10 @@
     *   Fix reporting of late video frames with identical release timestamps so
         that they are reported as dropped instead of skipped.
 *   Text:
+    *   SubRip: Add support for injecting a `CharsetDetector` into
+        `DefaultSubtitleParserFactory` to detect the character encoding of
+        standalone SubRip subtitles without a byte order mark
+        ([#2247](https://github.com/androidx/media/issues/2247)).
 *   Metadata:
 *   Image:
 *   DataSource:
--- a/libraries/extractor/src/main/java/androidx/media3/extractor/text/CharsetDetector.java
+++ b/libraries/extractor/src/main/java/androidx/media3/extractor/text/CharsetDetector.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package androidx.media3.extractor.text;
+
+import androidx.annotation.Nullable;
+import androidx.media3.common.util.UnstableApi;
+import java.nio.charset.Charset;
+
+/** Detects the character encoding of subtitle byte data. */
+@UnstableApi
+public interface CharsetDetector {
+
+  /**
+   * Detects the character encoding of the requested range of {@code data}.
+   *
+   * <p>The requested range is not guaranteed to contain a complete subtitle file.
+   *
+   * @param data The subtitle byte data.
+   * @param offset The start offset in {@code data}.
+   * @param length The number of bytes to inspect.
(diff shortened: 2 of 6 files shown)


tonihei:
Thanks for the contribution. @icbaker  I think that matches the idea in https://github.com/google/ExoPlayer/issues/3644#issuecomment-565503504 and could probably be merged.

icbaker:

PR #3369, opened 2026-08-13
Add dby1 compatible brand to the ftyp box if muxing Dolby Content

ybai001:
The "compatible brands" is part of 'ftyp' box, which is defined in ISO 14496-12. The usage of 'dby1' (Dolby compatible brand) is in https://professionalsupport.dolby.com/s/article/How-to-signal-Dolby-Vision-in-ISOBMFF-format-AKA-mp4-container?language=en_US

This PR updates both regular MP4 and fragmented MP4 writer paths so Dolby outputs advertise the expected compatible brand in the container header.


ybai001:
Thanks, @microkatz!

PR #3390, opened 2026-08-25
Fix stale decryptOnlyCodecFormat used for non-DRM periods

kmate-dolby:
When an audio stream is played back directly (passthrough or offload) but is DRM protected, MediaCodecAudioRenderer still instantiates a MediaCodec that is used only to decrypt the samples. In that case getMediaCodecConfiguration() stores the input format in decryptOnlyCodecFormat, and onOutputFormatChanged() uses it to configure the AudioSink with the original (encoded) format instead of the codec's raw output.

decryptOnlyCodecFormat is only ever assigned in
getMediaCodecConfiguration(), which runs from initCodec() and is therefore only reached for periods that actually create a codec. It was never cleared when the codec was released, so it lived for as long as the renderer did.

As a result, for a DRM, non-DRM, DRM period sequence the non-DRM period is played through codec bypass and does not initialize a codec, so getMediaCodecConfiguration() is not called. onOutputFormatChanged() then still sees the decryptOnlyCodecFormat left over from the previous DRM period and configures the AudioSink with that stale format instead of the current one. Because the two formats can differ (for example in channel count), the sink is configured with the wrong input format.

Fix this by clearing decryptOnlyCodecFormat when the codec is released, by overriding resetCodecStateForRelease() in MediaCodecAudioRenderer.

### DIFF
--- a/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/MediaCodecAudioRenderer.java
+++ b/libraries/exoplayer/src/main/java/androidx/media3/exoplayer/audio/MediaCodecAudioRenderer.java
@@ -817,6 +817,12 @@ public class MediaCodecAudioRenderer extends MediaCodecRenderer implements Media
   }
 
   @Override
+  protected void resetCodecStateForRelease() {
+    super.resetCodecStateForRelease();
+    decryptOnlyCodecFormat = null;
+  }
+
+  @Override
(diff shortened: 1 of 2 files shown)


kmate-dolby:
Hi @tianyif, thank you for self assigning. Anything I can do to help you review my proposed fix? Thank you in advance.

PR #3395, opened 2026-08-29
Supports any PCM encoding in SilenceSkippingAudioProcessor

nift4:


FongMi:
One question about the `Format` passed to `audioProcessorsFactory`: Should it describe the actual input format received by the processors created by the factory?

For example, with 6-channel, 24-bit PCM input and an output channel mapping of `[0, 1]`, the processing pipeline is effectively:

- Input: 6-channel / 24-bit PCM
- After `ChannelMappingAudioProcessor`: 2-channel / 24-bit PCM
- After `ToInt16PcmAudioProcessor`: 2-channel / 16-bit PCM
- Factory-created processor input: 2-channel / 16-bit PCM

However, the factory currently receives a `Format` describing 6-channel / 16-bit PCM.

I verified this by comparing the `Format` passed to the factory with the `AudioFormat` subsequently received by the created processor during `configure()`. The PCM encoding matched, but the channel counts were 6 and 2 respectively.

This may be intentional if the supplied `Format` is only meant to select processors based on `pcmEncoding`. If so, could that contract be documented? Otherwise, would it make sense to pass the effective post-mapping format to the factory?

nift4:
Fixed, thanks

Issue #3397, opened 2026-08-30
ProgressiveMediaPeriod.getFlags() reports container preroll only, never seek-induced preroll


oguzhaneksi:
### Version

Media3 main branch

### More version details

`a6db173f47f9b79a49d473293c6ebb65578a0678`

### Devices that reproduce the issue

Device-independent — this is a reporting defect in library code, not device behaviour.

### Reproducible in the demo app?

No

### Reproduction steps

Not reachable from the demo app: the only consumer of these flags sits behind
`Flags.FLAG_PER_STREAM_MEDIA_PROGRESSION`, which is off by default. It reproduces
directly at the `MediaPeriod` level, measured against a repo test asset:

1. Prepare a `ProgressiveMediaPeriod` over
   `libraries/test_data/src/test/assets/media/mp4/h264_1080p_30fps_10sec.mp4`.
   Its video track has no edit list, so `Format.hasPrerollSamples` is `false`,
   and it has 11 sync samples roughly 1 s apart.
2. Select the video track and enable the period at **1 500 000 us**, which falls
   mid-GOP between the sync samples at **999 411 us** and **1 998 489 us**.
3. Read buffers from the returned `SampleStream` and record their timestamps, and
   read `SampleStream.getFlags()`.

### Expected result

`getFlags()` reports the preroll that the stream is about to deliver. The stream
resumes at the preceding sync sample, so the samples between that sync sample and
the enable position are decode-only preroll, and the flag channel should say so —
either `FLAG_HAS_PREROLL`, or `FLAG_MAYBE_HAS_PREROLL` while the answer is still
undecided.

This is what `ChunkSampleStream` does. It returns `FLAG_MAYBE_HAS_PREROLL` while
`needToEvaluateInitialDiscontinuity` is set, then resolves to a definite answer
from `chunk.startTimeUs < lastSeekPositionUs` — i.e. from the position actually
resumed at.

### Actual result

The stream resumes from the sync sample at 999 411 us and delivers **16 preroll
samples**, timestamps spanning **[999 411, 1 495 233] us** — every one of them
before the 1 500 000 us enable position — while

```
getFlags() == 0        // no flags set; FLAG_HAS_PREROLL absent
```

Control, same asset, period enabled at position 0: 0 preroll samples,
`getFlags() == 0`. So the reported value is identical whether preroll is present
or not — it carries no information about position preroll at all.

`ProgressiveMediaPeriod.SampleStreamImpl.getFlags()` returns

```java
[...]

PR #3398, opened 2026-08-30
Do not report clipping preroll for clips that only set an end position

### LINKED ISSUE #3371: Returning from an interstitial playlist causes renderer disable and then reenable
### Version

Media3 1.11.0

### Description

Let's assume we are playing a playlist with several interstitial playlists, so the resulting playback looks like AaBbC, where (A), (B) and (C) are from the main playlist and (a) and (b) are interstitial playlists. The result is that between (a) and (B) the player is disabling and then reenabling the renderers. This happens because the player creates in MediaPeriodHolder.createMediaPeriod() a ClippingMediaPeriod for (A) and (B) (but not for (C) which has endPositionUs = C.TIME_UNSET).
Both new ClippingMediaPeriod() calls have enableInitialDiscontinuity=false (as isPrecededByTransitionFromSameStream=false) and such initial discontinuity causes renderers restart at the moment (B) starts playback. Modifying code to have enableInitialDiscontinuity=true for the periods following interstitials solves this problem.

Was reproduced on a proprietary Android device, but I do not believe the device is important.

Yes

Note: The demo app was modified to allow playback of the content with interstitials.

### Reproduction steps

Play a VoD content with several interstitials with SNAP=IN,
The main VoD content and all the interstitials should use the same set of video variants.

### Expected result

Smooth playback with all the transitions from the main content to interstitials and back happening without the renderers being disabled (and then reenabled).

### Actual result

Returning from all but the last interstitial playlist causes renderer disable/re-enable (and as a result quite a noticeable "hesitation" in the playback).

oguzhaneksi:
Fixes #3371.

## Problem

Playing a VoD playlist with HLS interstitials produces the period structure
`A a B b C c D`. Every return from an interstitial into content disables and
re-enables the renderers — a visible hesitation at each ad break, even when the
content resumes exactly on a segment boundary and there is nothing to skip.

## Root cause

`AdsMediaSource.createPeriod()` wraps a content period that has a following ad
group in a `ClippingMediaPeriod` with `enableInitialDiscontinuity = true`,
`startUs = 0` and an end clip. `shouldKeepInitialDiscontinuity` then applies a
heuristic: *resumed at a non-zero position, and not all-sync-samples ⇒ assume
preroll*. For H.264 that is always true, so every clipped content period claims
preroll.

The heuristic itself is a reasonable fallback. The problem is where its answer
went. `ClippingSampleStream.getFlags()` wrote it onto the
`SampleStream.getFlags()` protocol:

```java
if (hasPreroll) {
  flags &= ~FLAG_MAYBE_HAS_PREROLL;   // discards the child's "I don't know"
  flags |= FLAG_HAS_PREROLL;          // publishes a guess as a fact
}
```

So a clip overwrote whatever its child honestly reported and presented a guess
to every consumer of the protocol — including a child that had measured the
answer and reported no preroll.

## The change

`getFlags()` adds `FLAG_HAS_PREROLL` only when the clipping itself introduces
preroll:

```java
if (startUs != 0 && hasPreroll) {
```

With a non-zero clipping start the wrapped period genuinely cannot know about
the preroll, because sample streams provide buffers from a key-frame that may
precede the clip start. With a zero clipping start the samples delivered are
exactly the samples the wrapped period delivers, so its own report is exact and
is now passed through.

`shouldKeepInitialDiscontinuity`, `hasPreroll`, `anyTrackHasPreroll` and
`pendingInitialDiscontinuityPositionUs` are untouched, so `readDiscontinuity()`
behaves exactly as before.

The `startUs != 0` restriction is deliberate. Removing the discontinuity for
non-zero clip starts was tried and broke 88 `ClippingPlaylistPlaybackTest` /
`MergingPlaylistPlaybackTest` cases by feeding pre-clip-start buffers into the
codec.

## The default configuration is unaffected

The two channels are never both consulted:

| Configuration | Initial discontinuity read from |
|---|---|
| gate closed (default) | `MediaPeriod.readDiscontinuity()` |
| gate open (`FLAG_PER_STREAM_MEDIA_PROGRESSION`) | `SampleStream.getFlags()` |

The preroll bits have exactly one consumer,
`ExoPlayerImplInternal.updateRendererForTransition`, reachable only via
`maybeUpdateReadingRenderersPerStream()`, which `updatePeriods()` calls only when
`perStreamMediaProgressionEnabled`. Every other `getFlags()` call site
(`MediaCodecRenderer`, `DecoderVideoRenderer`, `ServerSideAdInsertionMediaSource`)
reads only `FLAG_STRICT_DURATION`.

`DefaultAnalyticsCollectorTest.adPlayback` passes unedited, which is the
regression proof for the default path.

## Tests

`ExoPlayerAdTest` gains two end-to-end reproductions and is parameterised on the
flag, following the existing pattern in `ExoPlayerTest`:

* `playAds_returningFromSegmentAlignedMidRoll_keepsRenderersEnabled` — mid-rolls
  on segment boundaries, content resumes at a sync sample. Goes from
  `[2 enables, 0 preroll rendered]` to `[1, 0]` with the gate open. This is #3371.
* `playAds_returningFromMidSegmentMidRoll_skipsPrerollOfClippedPeriodsOnly` —
  mid-rolls mid-segment, content really does carry preroll. Unchanged by this
  patch, and pinned so that any future change to it is deliberate.

Their `FakeSampleStream` reports preroll honestly
(`hasPreroll = segmentStartUs < preparePositionUs`), so the segment-aligned case
reports `0` and the mid-segment case reports `FLAG_HAS_PREROLL`. That isolates
exactly the fabrication this patch removes: the second test's numbers do not move.

`ClippingMediaPeriodTest`: three `startUs == 0` assertions lose the fabricated
flag, and one test is added for the newly reachable behaviour — a child reporting
`FLAG_MAYBE_HAS_PREROLL` under an end-only clip now has it passed through rather
than replaced by `FLAG_HAS_PREROLL`. That distinction matters: at
`ExoPlayerImplInternal` `MAYBE` defers the transition until the child resolves,
where `HAS_PREROLL` forces a renderer reset.

Full run over `:lib-exoplayer`, `:lib-exoplayer-dash`, `:lib-exoplayer-hls`,
`:lib-exoplayer-ima` and `:lib-exoplayer-rtsp`: 11 753 cases, 0 failures. Against
a baseline recorded on unmodified `main`, the only differences are in
`ExoPlayerAdTest` itself. No golden dump file changes.

## Device verification

Demo app, emulator, `bipbop-with-interstitials-boundary.m3u8` (3 midrolls with
`X-SNAP="IN"`, resuming on a segment boundary). Same build, the flag is the only
variable, counted from `EventLogger` over a full ~170 s playback:

| Configuration | videoEnabled | videoDisabled |
|---|---:|---:|
| gate closed | 3 | 2 |
| gate open | **1** | **0** |

Negative control, `bipbop-with-interstitials-unsnapped.m3u8` (midrolls
mid-segment, content genuinely carries preroll), gate open: 4 enables / 3
disables — all three ad→content transitions still reset, as they must. This patch
does not suppress a real discontinuity.

## Two things noticed nearby, not changed here

Both are pre-existing and out of scope for this patch; happy to split either into
its own change if useful.

1. `shouldKeepInitialDiscontinuity` calls
   `MimeTypes.allSamplesAreSyncSamples(mimeType, codecs)` for any track type,
   while the equivalent reasoning in `SampleQueue.canDiscardAllSamplesToStartTime`
   pairs it with `trackType == C.TRACK_TYPE_AUDIO`, on the grounds that the
   optimization is only safe where a sample's inherent duration is negligible.
2. `ClippingMediaPeriod.readDiscontinuity()` calls itself where the comment says
   "read an initial discontinuity from the child". It terminates only because the
   lines above it have already cleared `pendingInitialDiscontinuityPositionUs`,
   so the re-entry takes the other branch. It works, but it reads as a bug.


tonihei:
Thanks for the fix! I'll work on merging this internally, but I'll fix #3397 first.

PR #3404, opened 2026-09-04
fix: be less strict about range request responses

costela:
The code currently only does manual byte skipping for HTTP responses with a 200 status. There are other valid status codes that may be returned for a non-range response (e.g. 203), so we should be less strict about it.


rohitjoins:
@costela 

Thank you for the PR.

Additionally, please ensure your pull request targets the `main` branch instead of `release`. All of our ongoing development and bug fixes need to be integrated into `main` first.

costela:
@rohitjoins ah, my bad! Just recreated as https://github.com/androidx/media/pull/3410 :heavy_check_mark: 

Closing this one in favor of the new...

PR #3410, opened 2026-09-10
fix: be less strict about range request responses

costela:
The code currently only does manual byte skipping for HTTP responses with a 200 status. There are other valid status codes that may be returned for a non-range response (e.g. 203), so we should be less strict about it.

This PR is a re-creation of #3404 to ease merging.


rohitjoins:

PR #3411, opened 2026-09-13
Rechunk TrueHD samples in FragmentedMp4Extractor

### LINKED ISSUE #1519: Playback issues with Dolby TrueHD Fragmented MP4 files in ExoPlayer: Slow Time Progression and Playback Stops
### Version

Media3 1.3.1

### More version details

Tested with debug build. Also tested with version 1.3.0.

Fire TV Stick 4K - 2nd Generation (2023) - 8GB (Build Model: AFTKM)

_No response_

Yes

### Reproduction steps

1. Use the ExoPlayer demo in this repository.
2. Play a Dolby TrueHD Fragmented MP4 file by adding it to `media.exolist.json`.
3. Observe the playback behavior.

### Expected result

- The video player's time should progress normally.
- Playback should not stop.
- No "Audio sink error" should appear in the console.

### Actual result

- The video player's time progresses very slowly and sometimes even decreases.
- For video files (not audio-only), the video also progresses very slowly.
- Playback stops after 55 seconds of real-time (corresponding to 4 seconds of displayed time).
- A large number of `Audio sink error`s appear in the console.
```
Audio sink error
androidx.media3.exoplayer.audio.AudioSink$UnexpectedDiscontinuityException: Unexpected audio track timestamp discontinuity: expected 1000000226666, got 1000000014166
at androidx.media3.exoplayer.audio.DefaultAudioSink.handleBuffer(DefaultAudioSink.java:994)
at androidx.media3.exoplayer.audio.MediaCodecAudioRenderer.processOutputBuffer(MediaCodecAudioRenderer.java:744)
at androidx.media3.exoplayer.mediacodec.MediaCodecRenderer.bypassRender(MediaCodecRenderer.java:2310)
at androidx.media3.exoplayer.mediacodec.MediaCodecRenderer.render(MediaCodecRenderer.java:822)
at androidx.media3.exoplayer.ExoPlayerImplInternal.doSomeWork(ExoPlayerImplInternal.java:1112)
at androidx.media3.exoplayer.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:544)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:223)
at android.os.HandlerThread.run(HandlerThread.java:67)
```
- The non-fragmented version of the same content can be played without any problem.

### Media

Dolby TrueHD 48kHz audio-only fragmented mp4 (10sec fragment):
https://d2pbyw961c6y8f.cloudfront.net/dolby_truehd/sunahara_5.1.4_48kHz_dthd_part1_fragmented.mp4
This is a Dolby TrueHD mp4 file created by dolby's official encoder (dolby enocoding engine) and fragmented using MP4Box tool.

Dolby TrueHD 48kHz audio-only mp4 :
https://d2pbyw961c6y8f.cloudfront.net/dolby_truehd/sunahara_5.1.4_48kHz_dthd_part1.mp4
This is the non-fragmented version of the same content. This can be played without any problem.

### Bug Report

mvletter:
Fixes #1519.

I hit this while fixing something else in Jellyfin, and I need it fixed before lossless TrueHD passthrough actually works.

Right now you get something beautiful: the receiver proudly says "Dolby Atmos" and zero sound comes out. The bytes are fine, so something further down the line can't agree on how long they last.

## What is going on

`Mp4Extractor` groups TrueHD samples into chunks of `Ac3Util.TRUEHD_RECHUNK_SAMPLE_COUNT` (16) before it outputs their metadata, which I first took for a packing optimisation. It isn't, because `ExtractorUtil.getFramesPerEncodedSample` works out the duration of a TrueHD buffer by reading the audio frame count from its first syncframe and multiplying that by the same 16. The audio path runs on a standing agreement that every TrueHD buffer holds sixteen container samples, and the rechunker is the half of that agreement the extractor is supposed to keep.

`FragmentedMp4Extractor` never signed it, so it handed over one container sample per buffer. An MLP access unit at 48 kHz is 40 audio frames while the sink kept counting 640, which meant the sink's clock ran sixteen times faster than the media until it gave up with `AudioSink$UnexpectedDiscontinuityException`.

That ratio is visible in the original report, where 226666 / 14166 comes out at exactly 16.0.

@Hiroki-Tamaru found the missing rechunker and @icbaker confirmed the approach, so this implements it.

## What this does

`TrackBundle` now owns a `TrueHdSampleRechunker` whenever its format is TrueHD, and sample metadata runs through it. Three things needed care, and all three follow from reading a stream fragment by fragment instead of in one pass.

A pending chunk has to survive fragment boundaries, because `resetFragmentInfo()` runs for every `moof` and resetting the rechunker there would quietly drop up to fifteen samples per fragment. The chunk is dropped only in `seek()`, where its samples genuinely have left the sample queue.

A partial chunk is committed when the input ends, so the tail of a stream survives instead of disappearing. That still holds when `BundledChunkExtractor` reuses one extractor across segments, since `read()` returns `RESULT_END_OF_INPUT` before the next `init()` calls `seek()`.

Encrypted samples keep their own `CryptoData` and are still output one at a time. The pending chunk is committed in `STATE_READING_SAMPLE_START`, before any of that sample's bytes reach the output, because `TrackOutput.sampleMetadata` measures size and offset against everything written so far and would otherwise point the chunk at the wrong bytes. The syncframe peek moved to that same spot for a related reason: `startSample` can throw inside `peekFully` while the parser stays in `STATE_READING_SAMPLE_START`, so peeking after `outputSampleEncryptionData` would write the encryption header twice on the retry.

## Tests

There are two levels here, because a matching dump only proves the extractor behaves and says nothing about whether the sink is happy.

`FragmentedMp4ExtractorParameterizedTest` gets `sample_dthd_fragmented.mp4`, the existing `sample_dthd.mp4` remuxed into five fragments and weighing 97 KB. Its fragments hold 120, 120, 120, 120 and 22 samples, so none of them contains a whole number of chunks and the rechunker is forced to carry chunks across fragment boundaries, while the final chunk of 6 samples exercises the flush at end of input. The dump comes out identical to `sample_dthd.mp4`'s on every single sample: 32 chunks, 94656 bytes, same timestamps, same hashes.

`TrueHdFragmentedMp4PlaybackTest` plays both files through a real `DefaultAudioSink` in passthrough and asserts that the sink reports no discontinuity. Two things there are worth writing down, since I got them wrong the first time round: passthrough capabilities are only queried when `Util.isTv()`, so the test has to pretend to be a television, and `ShadowAudioTrack` needs both `addDirectPlaybackSupport` and `addAllowedNonPcmEncoding`. Without those the audio track is never selected at all and the test passes while proving nothing, which is why it also asserts on `onAudioInputFormatChanged`.

Run against `main` without the extractor change, that same test reproduces the reported failure down to the timestamps:

```
expected 1000000226666, got 1000000014166
```

Those are the numbers from the original bug report, and they match what I saw on a Chromecast with Google TV feeding a Denon AVR.

Locally I ran the unit tests for `:lib-extractor`, `:lib-exoplayer`, `:lib-exoplayer-dash`, `:lib-exoplayer-hls`, `:lib-exoplayer-smoothstreaming` and `:lib-container`, which comes to 17268 tests with no failures.

## How this was made

I used an AI agent for the research and for a large part of the code, and I would rather say that up front than leave you guessing. It wasn't taken on trust, which is why this description leans on evidence you can re-run.

Every test was written to fail first and verified on a clean worktree of `main`, where the extractor test fails all eighteen of its parameterised cases and the playback test reproduces the original timestamps exactly. Review turned up two real defects that were fixed before this went out: the syncframe peek could write the encryption header twice on a retried read, and a clear-to-encrypted transition could commit a pending chunk against the wrong bytes in the sample queue. Both are why the code is shaped the way it is.

If anything here still looks wrong, say so plainly and I'll dig into it rather than defend it.

## What this does not fix

A chunk holding fewer than sixteen samples is still counted as a full sixteen by the sink, because `getFramesPerEncodedSample` multiplies by the constant instead of summing the syncframes actually present in the buffer. That touches the last chunk of a stream, and in DASH or HLS a chunk at a segment boundary whenever the segment doesn't hold a whole number of chunks, though segments of a whole number of seconds do at n * 1200 samples. Summing the real syncframes would remove the assumption altogether and I'm happy to do that separately, but it changes the audio path and felt like a different review. `Mp4Extractor` behaves the same way today.

Encrypted TrueHD keeps its old behaviour, since merging samples would throw away the per-sample `CryptoData`. If anyone has a TrueHD stream under CENC, I would genuinely like to see it.


mvletter:
I found one more edge in the segment lifecycle after submitting this PR.

`BundledChunkExtractor` reuses the same `FragmentedMp4Extractor` for the next media segment and calls `seek(0, 0)` before reading it. The full TrueHD reset forgot that a major syncframe had already been found. The bytes were still written, but samples before the next major syncframe had no metadata. With this fixture that meant 94,656 bytes written and only 81,782 covered.

The follow-up separates those two bits of state. A segment transition drops the pending sample count while keeping syncframe detection alive. A non-zero seek still resets both.

The regression feeds the same extractor five bounded inputs, matching how `BundledChunkExtractor` sees media segments. It produces 34 metadata chunks, and the combined sample bytes match a single-pass extraction exactly.

I also cut the test from 165 to 88 lines after spotting that I had rebuilt infrastructure the repository already provides. It now uses `FakeExtractorInput` and `FakeTrackOutput`, without a custom datasource, output implementation or MP4 parser.

Checks completed:

- `:lib-extractor:testDebugUnitTest`
- `:lib-exoplayer:testDebugUnitTest`
- lint for both modules
- `google-java-format`

PR #3414, opened 2026-09-15
Fix ArrayIndexOutOfBoundsException in ScriptTagPayloadReader on malformed FLV metadata

Jobians:
`readAmfObject`, `readAmfEcmaArray`, and `readAmfStrictArray` in
ScriptTagPayloadReader trusted the encoded AMF stream unconditionally:
`readAmfObject`'s `while (true)` loop had no check against
`data.bytesLeft()` and just kept reading until it found an
AMF_TYPE_END_MARKER byte, while the array readers looped exactly
`count` times based on a declared length field with no check that the
buffer actually contained that much data.

Streams with malformed `onMetaData` (e.g. a missing end marker, or a
declared array count exceeding the actual encoded data) cause these
loops to read past the end of the tag's ParsableByteArray buffer,
throwing ArrayIndexOutOfBoundsException from readAmfType() ->
data.readUnsignedByte() once the read position reaches the buffer's
length.

This adds bytesLeft() bounds checks throughout, so malformed metadata
is gracefully truncated/ignored (falling back to no duration/seek-map,
consistent with existing behavior for unsupported metadata) instead of
crashing playback. Well-formed FLV metadata parses identically to
before.

Reproduced and verified against a live FLV stream from a third-party
RTMP publisher whose onMetaData reliably triggered this crash;
confirmed the fix resolves playback while preserving seek-map/duration
support, unlike a workaround that skips script tags entirely.


Jobians:
Please review @rohitjoins @tonihei @microkatz @tianyif @microkatz

PR #3422, opened 2026-09-20
Continue playback on another decoder when one fails mid-stream

### LINKED ISSUE #3421: Recover from runtime decoder failures, not just initialization failures
`setEnableDecoderFallback` on `DefaultRenderersFactory`/`MediaCodecRenderer` only retries with another decoder when the first fails to *initialize*. When a decoder initializes successfully and then throws (e.g. a `MediaCodecVideoDecoderException`/`MediaCodecAudioDecoderException`) partway through playback, there's no equivalent recovery. The player just fails, even though a different decoder for the same format may well have kept working.

This is a recurring pattern, a few examples already on this tracker:
- #2616: an EAC-3 decoder dies mid-stream on several tablets/TVs, no fallback.
- #2711 (closed by the stale bot, not resolved): a MediaTek HEVC decoder produces no frames after an OS update on several OEM devices.

I hit the same shape of bug in Jellyfin's Android TV client: a MediaTek HEVC decoder (`OMX.MTK.VIDEO.DECODER.HEVC`) throws a vendor-internal error after successful initialization, specifically when resuming/seeking into a particular file. In each of these cases, the underlying decoder bug is out of the library's control (and often out of the app's control too, since it's proprietary vendor firmware), but the *lack of any retry path* is something the library could address in the same spirit as `setEnableDecoderFallback` already does for init failures.

I built and verified an app-level workaround using only existing public APIs: a custom `MediaCodecSelector` that excludes decoder names which have thrown at runtime, combined with catching the resulting `PlaybackException` and retrying locally (re-`prepare()` + seek back). It worked reliably on the affected device. I proposed shipping it in Jellyfin's Android TV client, but a maintainer there closed the PR (https://github.com/jellyfin/jellyfin-androidtv/pull/5827), citing both the risk of adding this to their legacy playback code and the view that this class of resilience belongs in the library rather than in each app individually: "Dealing with bad decoders should also not be a responsibility of our app, this should be done upstream in media3. As long as the library tells us it can play something, it should do so." That's exactly the gap this issue is asking about.

Given how many apps are independently re-implementing some version of this, it seems like a reasonable candidate for the library itself, for example by extending `setEnableDecoderFallback` (or a new sibling option) to also exclude-and-retry on a `MediaCodecDecoderException` thrown after initialization, not just on `DecoderInitializationException`.

HaraldNordgren:
I ran into a specific HEVC file that would reliably fail on a Philips Android TV: it played fine for the first 48 minutes, then hard failed every single time I tried to resume from that point, at the exact same timestamp. Device logs showed the TV's MediaTek hardware HEVC decoder itself crashing partway through decoding, well after it had already initialized and decoded almost an hour of video successfully. Since decoder fallback (setEnableDecoderFallback) only ever kicks in when a decoder fails to initialize, a crash like this had no recovery path and surfaced as a fatal playback error every time, on that device, for that file.

This extends decoder fallback to also cover a decoder that crashes partway through decoding, not just one that fails to start. When that happens and another decoder is available for the format, it's excluded and a different decoder is selected, the same way initialization failures are already handled. If no other decoder is available, the failure still surfaces as before.

I verified this against the real failure: with this change, the same file that previously failed every time on this device now falls back to a software decoder mid-stream and plays normally, both from the start and when resuming partway through.

Closes #3421


HaraldNordgren:
Ping @rohitjoins @tonihei

microkatz:
@HaraldNordgren

As per repo policy, before we review your CL, would you mind rebasing your CL to the androidx:main branch?

HaraldNordgren:
@microkatz Done! Maybe it would be good to change the default branch of this repo to `main` to avoid this in the future?

